PR #30, opened 2019-09-11
ent/gen: add support for external tempaltes execution

a8m:
Reviewed By: dlvhdr

Differential Revision: D17318127


chris-rock:
@a8m Just discovered the external template mechanism (not sure why I missed it sofar). It saved my day today. Thank you

PR #55, opened 2019-10-03
fix import in `first edge` docs code sample

tmus:


a8m:
Thanks for your contribution @tmus

Issue #70, opened 2019-10-06
entc generate failed


freezeChen:
user.go
```
package schema

import (
	"github.com/facebookincubator/ent"
	"github.com/facebookincubator/ent/schema/edge"
	"github.com/facebookincubator/ent/schema/field"
)

// User holds the schema definition for the User entity.
type User struct {
	ent.Schema
}

// Fields of the User.
func (User) Fields() []ent.Field {
	return []ent.Field{
		field.Int("age"),
		field.String("name"),
	}
}

// Edges of the User.
func (User) Edges() []ent.Edge {
	return []ent.Edge{
		edge.To("friends", User.Type),
	}
}

```

run entc generate ./ent/schema   is failed,only generate an empty folder. 
```
$ entc generate ./ent/schema
entc/gen: write file C:\project\gowork\src\awesomeProject\ent\user_create.go: formatting source: C:\project\gowork\src\awesomeProject\e
nt\user_create.go:21:19: unknown escape sequence (and 1 more errors)
```

alexsn:
Don't think we ever tested `entc` on windows. I'll take a look.

freezeChen:
in /entc/cmd/entc/entc.go  function  loadGraph   
```
cfg.Package = filepath.Dir(spec.PkgPath)
```
when os is windows,  package is failed

a8m:
Closed with 7597f079129814c3ac37f823fae9d1b6f862e40b.

Thanks for reporting this issue.

PR #71, opened 2019-10-06
fix generate is failed in windows

freezeChen:
#70

### DIFF
--- a/entc/cmd/entc/entc.go
+++ b/entc/cmd/entc/entc.go
@@ -133,6 +133,7 @@ func loadGraph(path string, cfg gen.Config) (*gen.Graph, error) {
 	}
 	cfg.Schema = spec.PkgPath
 	cfg.Package = filepath.Dir(spec.PkgPath)
+	cfg.Package = strings.Replace(cfg.Package, "\\", "/", -1)
 	return gen.NewGraph(cfg, spec.Schemas...)
 }
 


entc/cmd/entc/entc.go
```diff
@@ -133,6 +133,7 @@ func loadGraph(path string, cfg gen.Config) (*gen.Graph, error) {
 	}
 	cfg.Schema = spec.PkgPath
 	cfg.Package = filepath.Dir(spec.PkgPath)
+	cfg.Package = strings.Replace(cfg.Package, "\\", "/", -1)
```
a8m: Thanks.
Can you please add a comment here? Thanks

alexsn:
Please use `filepath.ToSlash` instead of `strings.Replace`, in addition collapse both assignments into a single line (e.g. `cfg.Package = filepath.ToSlash(filepath.Dir(spec.PkgPath))`)

entc/cmd/entc/entc.go
a8m: Wait, I think the problem is with filepath -> path.

entc/cmd/entc/entc.go
```diff
@@ -133,6 +133,7 @@ func loadGraph(path string, cfg gen.Config) (*gen.Graph, error) {
 	}
 	cfg.Schema = spec.PkgPath
 	cfg.Package = filepath.Dir(spec.PkgPath)
```
a8m: From the `filepath godoc:
filepath.Dir:
> Dir calls Clean on the path and trailing slashes are removed.
filepath.Clean
> Finally, any occurrences of slash are replaced by Separator.

And `Separator` is set to `os.PathSeparator`. Can you try replace `filepath` with `path`? 
```suggestion
	cfg.Package = path.Dir(spec.PkgPath)
```

a8m: OK, checked it now on Windows machine.

```go
println(path.Dir("a/b/c"))     // prints "a/b"
println(filepath.Dir("a/b/c")) // prints "a\b"
```

a8m:
Closed with 7597f079129814c3ac37f823fae9d1b6f862e40b

Thanks for your contribution.

PR #78, opened 2019-10-08
wip: postgresql

cliedeman:
TODO:
- [ ] Handle enum creation
- [ ] Auto increment mapping (Postgres only has identity and serial)

Design
- [ ] Postgres Range type?
- [ ] Should more specialised date/time types be created?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,6 +71,10 @@ jobs:
         environment:
           <<: *mysql_env
           MYSQL_TCP_PORT: 3308
+      - image: postgres:11
+        environment: &postgresql_env
+          POSTGRES_DB: test
+          POSTGRES_PASSWORD: pass
       - image: entgo/gremlin-server
         entrypoint: /opt/gremlin-server/bin/gremlin-server.sh
         command: conf/gremlin-server.yaml
--- a/dialect/sql/driver.go
+++ b/dialect/sql/driver.go
@@ -43,7 +43,7 @@ func (d Driver) DB() *sql.DB {
 // Dialect implements the dialect.Dialect method.
 func (d Driver) Dialect() string {
 	// if the underlying driver is wrapped with opencensus driver.
-	for _, name := range []string{dialect.MySQL, dialect.SQLite} {
+	for _, name := range []string{dialect.MySQL, dialect.Postgres, dialect.SQLite} {
 		if strings.HasPrefix(d.dialect, name) {
 			return name
 		}
--- a/dialect/sql/schema/migrate.go
+++ b/dialect/sql/schema/migrate.go
@@ -64,6 +64,8 @@ type Migrate struct {
 func NewMigrate(d dialect.Driver, opts ...MigrateOption) (*Migrate, error) {
 	m := &Migrate{}
 	switch d.Dialect() {
+	case dialect.Postgres:
+		m.sqlDialect = &PostgreSQL{Driver: d}
 	case dialect.MySQL:
 		m.sqlDialect = &MySQL{Driver: d}
 	case dialect.SQLite:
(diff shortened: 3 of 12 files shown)


erka:
You could use `pg_isready` for healthcheck of postgres

alexsn:
Thanks for the PR @cliedeman, please note that integration tests are failing and needs fixing.

a8m:
Thanks for the PR @cliedeman, I'll review this after finishing and landing #84.

cliedeman:
@alexsn This is as far as I got before getting stuck.

builder.go will need to be refactored to allow changing the quote character and how arguments are generated.

Some example code that needs to be changed

```
func (s *Selector) C(column string) string {
	if s.as != "" {
		return fmt.Sprintf("`%s`.`%s`", s.as, column)
	}
	return s.Table().C(column)
}

// Columns returns a list of formatted strings for a selected columns from this statement.
func (s *Selector) Columns(columns ...string) []string {
	var names []string
	for _, c := range columns {
		names = append(names, s.C(c))
	}
	return names
}
```

The identifier quoting needs to be deferred to later execution process. Ideally when Query is invoked probably. 

Arg is even more problematic because postgres requires numbered arguments. This means we need to track query variables and handle the nested query case

```
func (b *Builder) Arg(a interface{}) *Builder {
	switch a := a.(type) {
	case *raw:
		b.WriteString(a.s)
	default:
		b.WriteString("?")
		b.args = append(b.args, a)
	}
	return b
}
```

I have tried a few attempts to refactor it but nothing promising yet. My latest idea if to remove the buffer embed and replace all the WriteStrings with tokens or objects that can be converted to dialect specific sql later

Ciaran

cliedeman:
Rebased on master. next hurdle is handling postgres enums. They need a separate sql statement to be emitted to create a user type.

Postgres also does not have a real auto increment type and the serial type [should be avoided](https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_serial)

a8m:
Hey @cliedeman, can you please split it into separated PRs? For example, `Column.Postgres` and `Column.ScanPostgres` should be in one PR, and so on.

BTW, we start to work on this as well, so it'll help us to push this feature faster.

cliedeman:
@a8m I will probably only get to it this weekend.I don't mind if you roll the changes into another PR if you get to it before I do

ajwerner:
Do you have any interest in not using postgres’s Reum type? SQLite doesn’t have them. 

I’m asking because I suspect that will be the only compatibility problem that would prevent this dialect from also working with CockroachDB.

cliedeman:
It’s not on the roadmap to support cockroach but it should work if you just don’t use enums?

ajwerner:
Cool, I’ll try it out after the Postgres support merges. Cockroach speaks Postgres wire protocol and the Postgres sql dialect but it doesn’t have 100% compatibility. Enumerated are a big one. Triggers are also not supported but there is a changefeed mechanism which can generally be used instead. I understand cockroach isn’t on the roadmap currently but I’m hopeful getting it to work after Postgres support merges should be a small amount of work which I’d be happy to try to tackle.

a8m:
PostgreSQL support was added. Closing.

Thank you very much @cliedeman for your contribution! I really appreciate the time you invested in this PR, but I tried to use it and some of the major functionally didn't work for me.

@ajwerner, you are welcome to give it a try and let me know if that works for you. Thanks!

cliedeman:
@a8m Not a problem Thanks, I am sure I will find something else to contribute.

T-M-A:
> I understand cockroach isn’t on the roadmap currently but I’m hopeful getting it to work after Postgres support merges should be a small amount of work which I’d be happy to try to tackle.
Hello, 
have news with cockroach?

PR #95, opened 2019-10-15
Restore ids declaration in query builder IDs

htdvisser:
This pull request restores a line in the query builder template that I believe was [removed by accident](https://github.com/facebookincubator/ent/commit/0c46ce4e5e6904dfb8fbc0610bcba7a8de268e43#diff-855ebdbcdbdd5bf91fcb98886276275eL203).


a8m:
Thanks for taking this @htdvisser. I've landed it.

PR #109, opened 2019-10-21
ent/entc: rename time mixin fields

alexsn:
Summary:
- created_at -> creation_time
- updated_at -> update_time

Differential Revision: D18035896


marwan-at-work:
Out of curiosity, any reason why created_at/updated_at should be avoided?

alexsn:
> Out of curiosity, any reason why created_at/updated_at should be avoided?

It's personal preference and there's no reason why created_at/updated_at should be avoided.

Note that the reason for the rename being that I plan to also rename time fields predicates to `Before` / `After` instead of `LT` / `GT` and `CreatedAtBefore` / `UpdatedAtAfter` just sound incorrect. I addition I though of calling those fields `Created` / `Updated` but those names imply boolean types.

PR #120, opened 2019-10-24
schema-edges.md: fix conflict

gorda:


a8m:
Hey @gorda, thanks for the contribution. 

It’s not a git conflict. It uses the `diff ` option intentionally in order to explain which lines to delete/add. .

PR #136, opened 2019-11-01
schema: enum builder default value setter

darioielardi:
Enum field builder should have a Default() setter, as stated in the inline docs example:
```
field.Enum("state").
	Values(
		"on",
		"off",
	).
	Default("on")
```
but it's actually missing.


a8m:
Hey @darioielardi, thanks for your contribution. 

Can you please add UT and integration tests for it? (codegen should validate that the given default is one of the values).

Thanks

darioielardi:
Right I'll add tests for everything, should I open another pull request for json default value setter?

a8m:
We intentionally disabled the default value option for JSON. Mainly, because the lack of type safety.

darioielardi:
I removed the JSON part, and wrote UT for the enum default value. Writing integration tests I found out that the code generated for the default value ( the constant ) has two problems with enums:

1. type assertion fails because the field type is a custom generated one and the value passed to the Default() setter is a string.
2. Injecting $f.Type in the template results in an import of the current package, because the type is stored in TypeInfo as packageName.Type.

example const generated in `mymodule/ent/user/user.go`: 
```go
import (
  ...
  "mymodule/ent/user"
  ...
)

const (
  ..
  DefaultRole = descRole.Default.(user.Role)
  ...
)
```

First problem is solved with a `string` type assertion and then a conversion to the real generated enum type:

```go
{{ $default }} = {{ $f.Type }}({{ $desc }}.Default.(string))
```
but it still injects the type with the package name:

```go
DefaultRole = user.Role(descRole.Default.(string))
```

I can't figure out a good way to solve that. Any ideas?

darioielardi:
Didn't realize the solution was under my eyes, everything is now solved and I'm just writing integration tests.

darioielardi:
Everything should be done, you can now review.

.gitignore
```diff
@@ -0,0 +1 @@
+.idea/
```
a8m: Please remove this file. 

You can use [global gitignore](https://help.github.com/en/github/using-git/ignoring-files) for cases like these.

entc/gen/template/meta.tmpl
```diff
@@ -58,8 +58,14 @@ var (
 		{{ end -}}
 		{{ if $f.Default }}
 			{{- $default := $f.DefaultName -}}
-			// {{ $default }} holds the default value on creation for the {{ $f.Name }} field.
-			{{ $default }} = {{ $desc }}.Default.({{ if $f.IsTime }}func() {{ end }}{{ $f.Type }})
+			{{ if $f.IsEnum }}
+			    {{ $enum := trimPackage $f.Type.String $.Package }}
+			    // {{ $default }} holds the default value on creation for the {{ $f.Name }} field.
+			    {{ $default }} = {{ $enum }}({{ $desc }}.Default.(string))
```
a8m: I'll prefer to use one of the constants that are defined in this file (yeah, enum is a pretty special here).

entc/load/schema.go
```diff
@@ -37,6 +37,7 @@ type Field struct {
 	Tag           string          `json:"tag,omitempty"`
 	Size          *int64          `json:"size,omitempty"`
 	Enums         []string        `json:"enums,omitempty"`
+	EnumDefault   string          `json:"enum_default,omitempty"`
```
a8m: I don't think a new field is required here. Maybe we can just change the `Default` field to be type `interface{}`.

Let me know if you are fine with this, and I can help with the implementation part.

.gitignore
darioielardi: yeah I'm sorry, that was unintentional

entc/load/schema.go
darioielardi: As far as I understand right now the `load.Field.Default` is a bool and is used to know if that field has a default value set, it does not hold the actual value, right? That's why I came up with that `EnumDefault`, which is not the best solution, I agree, but I didn't want to mess up with the whole thing.

entc/gen/template/meta.tmpl
darioielardi: I'm afraid I don't get what you mean here.

a8m:
Hey, I didn't forget this diff.
I want to push it, however, there are some changes that need to be done first.

I'll get to it in the upcoming days.

darioielardi:
No worries, feel free to answer my questions when you got the time.

a8m:
Hey, I've added some changes to the `load` package (with default values) that can help you.
Let me know if you want to rebase and take it from there, or want me to pull and work on top of your changes. 

Thanks!

darioielardi:
I'd probably have to wait some weeks to find the time to work on this, so it's probably better if you pull and work yourself on top of my changes, thank you.

a8m:
I pulled this PR and rebased this on master (created a separated PR with your commit #271) . I'll merge this soon. 

Closing this one. 

Thanks for your contribution!

PR #137, opened 2019-11-02
Remove update validators for immutable fields

htdvisser:
This pull request modifies the update builder template so that it no longer generates code to validate immutable fields.

The first commit breaks the integration code by making the `number` field of `Card` immutable (the generated validation code references non-existent fields of `CardUpdate`/`CardUpdateOne`). 

The second commit fixes the issue by making the `update/save` template not render validation code for immutable fields.


a8m:
Thanks for your contribution @htdvisser. 

Code looks good, but build is failing.
Maybe you can test this change on a new field with a default value.

htdvisser:
Looks like CI is happy now.

Since `number` didn't have a default value, I'm guessing that the default values that were being tested here were in fact `CreateTime` and `UpdateTime`. Let me know if that assumption was wrong, or if you want any other changes.

a8m:
The tests were failing because `number` didn't have setters in the updaters, so I suggested that instead of fixing/changing the tests, you could just add a new immutable field with default value (the `Default`/`Optional` options make it easy to add new fields without changing all test cases).

PR #219, opened 2019-12-05
Resolve panic where nothing would be migrated with SQLite3 driver

erikh:
This was caused by the current state being represented as nil; this was
resolved by returning an empty change set when the current value is nil.

Hoping to turn this into a series of commits where I contribute small additions towards the sqlite3 work. LMK if I'm heading in the right direction; specifically, I don't know where the best place for tests are.


erikh:
Signed the CLA, fwiw. I imagine it takes some time to sync.

erikh:
I think I got this wrong.. it looks like table is returning nil instead of handling the schema.

OK. I will revisit this patch and submit something new when i have it. sorry for the noise.

PR #221, opened 2019-12-05
WIP: SQLite3 support for migrations

erikh:
This is a basic start, a lot of cut and paste but it does pass some very basic tests now. I still have more work to do but I was hoping we could discuss some of it as it arrives, so here's the PR.

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -254,7 +254,11 @@ func (t *TableAlter) DropColumn(c *ColumnBuilder) *TableAlter {
 
 // AddForeignKey adds a foreign key constraint to the `ALTER TABLE` statement.
 func (t *TableAlter) AddForeignKey(fk *ForeignKeyBuilder) *TableAlter {
-	t.Queries = append(t.Queries, &Wrapper{"ADD CONSTRAINT %s", fk})
+	if t.Dialect() == dialect.SQLite {
+		t.Queries = append(t.Queries, &Wrapper{"ADD COLUMN %s", fk})
+	} else {
+		t.Queries = append(t.Queries, &Wrapper{"ADD CONSTRAINT %s", fk})
+	}
 	return t
 }
 
@@ -339,17 +343,24 @@ func (fk *ForeignKeyBuilder) OnUpdate(action string) *ForeignKeyBuilder {
 
 // Query returns query representation of a foreign key constraint.
 func (fk *ForeignKeyBuilder) Query() (string, []interface{}) {
-	if fk.symbol != "" {
+	if fk.Dialect() != dialect.SQLite && fk.symbol != "" {
 		fk.Ident(fk.symbol).Pad()
 	}
-	fk.WriteString("FOREIGN KEY")
-	fk.Nested(func(b *Builder) {
-		b.IdentComma(fk.columns...)
-	})
+
+	if fk.Dialect() == dialect.SQLite {
+		fk.Pad().Ident(fk.columns[0])
+	} else {
+		fk.WriteString("FOREIGN KEY")
+		fk.Nested(func(b *Builder) {
+			b.IdentComma(fk.columns...)
+		})
+	}
+
 	fk.Pad().Join(fk.ref)
 	for _, action := range fk.actions {
 		fk.Pad().WriteString(action)
 	}
+
 	return fk.String(), fk.args
 }
 
--- a/dialect/sql/schema/migrate.go
+++ b/dialect/sql/schema/migrate.go
@@ -7,6 +7,7 @@ package schema
 import (
 	"context"
 	"crypto/md5"
+	"errors"
 	"fmt"
 	"math"
 	"sort"
@@ -154,7 +155,7 @@ func (m *Migrate) create(ctx context.Context, tx dialect.Tx, tables ...*Table) e
 		}
 		fks := make([]*ForeignKey, 0, len(t.ForeignKeys))
 		for _, fk := range t.ForeignKeys {
-			exist, err := m.fkExist(ctx, tx, fk.Symbol)
+			exist, err := m.fkExist(ctx, tx, t, fk)
 			if err != nil {
 				return err
 			}
@@ -250,6 +251,14 @@ func (c *changes) dropColumn(name string) (*Column, bool) {
 // changeSet returns a changes object to be applied on existing table.
 // It fails if one of the changes is invalid.
 func (m *Migrate) changeSet(curr, new *Table) (*changes, error) {
+	if curr == nil {
+		return nil, errors.New("current state could not be determined during change set generation")
+	}
+
+	if new == nil {
+		return nil, errors.New("determined state could not be determined during change set generation")
+	}
+
 	change := &changes{}
 	// pks.
 	if len(curr.PrimaryKey) != len(new.PrimaryKey) {
@@ -449,7 +458,7 @@ type sqlDialect interface {
 	init(context.Context, dialect.Tx) error
 	table(context.Context, dialect.Tx, string) (*Table, error)
 	tableExist(context.Context, dialect.Tx, string) (bool, error)
-	fkExist(context.Context, dialect.Tx, string) (bool, error)
+	fkExist(context.Context, dialect.Tx, *Table, *ForeignKey) (bool, error)
 	setRange(context.Context, dialect.Tx, string, int) error
 	dropIndex(context.Context, dialect.Tx, *Index, string) error
 	// table, column and index builder per dialect.
--- a/dialect/sql/schema/mysql.go
+++ b/dialect/sql/schema/mysql.go
@@ -47,9 +47,9 @@ func (d *MySQL) tableExist(ctx context.Context, tx dialect.Tx, name string) (boo
 	return exist(ctx, tx, query, args...)
 }
 
-func (d *MySQL) fkExist(ctx context.Context, tx dialect.Tx, name string) (bool, error) {
+func (d *MySQL) fkExist(ctx context.Context, tx dialect.Tx, table *Table, fk *ForeignKey) (bool, error) {
 	query, args := sql.Select(sql.Count("*")).From(sql.Table("INFORMATION_SCHEMA.TABLE_CONSTRAINTS").Unquote()).
-		Where(sql.EQ("TABLE_SCHEMA", sql.Raw("(SELECT DATABASE())")).And().EQ("CONSTRAINT_TYPE", "FOREIGN KEY").And().EQ("CONSTRAINT_NAME", name)).Query()
+		Where(sql.EQ("TABLE_SCHEMA", sql.Raw("(SELECT DATABASE())")).And().EQ("CONSTRAINT_TYPE", "FOREIGN KEY").And().EQ("CONSTRAINT_NAME", fk.Symbol)).Query()
 	return exist(ctx, tx, query, args...)
 }
 
@@ -289,9 +289,8 @@ func (d *MySQL) scanColumn(c *Column, rows *sql.Rows) error {
 	if nullable.Valid {
 		c.Nullable = nullable.String == "YES"
 	}
-	switch parts := strings.FieldsFunc(c.typ, func(r rune) bool {
-		return r == '(' || r == ')' || r == ' ' || r == ','
-	}); parts[0] {
+
+	switch parts := typeFields(c.typ); parts[0] {
 	case "int":
 		c.Type = field.TypeInt32
 		if len(parts) == 3 { // int(10) unsigned.
--- a/dialect/sql/schema/sqlite.go
+++ b/dialect/sql/schema/sqlite.go
@@ -7,10 +7,14 @@ package schema
 import (
 	"context"
 	"fmt"
+	"math"
+	"strconv"
+	"strings"
 
 	"github.com/facebookincubator/ent/dialect"
 	"github.com/facebookincubator/ent/dialect/sql"
 	"github.com/facebookincubator/ent/schema/field"
+	"github.com/pkg/errors"
 )
 
 // SQLite is an SQLite migration driver.
@@ -119,6 +123,52 @@ func (*SQLite) cType(c *Column) (t string) {
 	return t
 }
 
+func (d *SQLite) typeField(c *Column, str string) error {
+	switch parts := typeFields(str); parts[0] {
+	case "int", "integer":
+		c.Type = field.TypeInt32
+	case "smallint":
+		c.Type = field.TypeInt16
+	case "bigint":
+		c.Type = field.TypeInt64
+	case "tinyint":
+		c.Type = field.TypeInt8
+	case "double", "real":
+		c.Type = field.TypeFloat64
+	case "timestamp", "datetime":
+		c.Type = field.TypeTime
+	case "blob":
+		c.Size = math.MaxUint32
+		c.Type = field.TypeBytes
+	case "text":
+		c.Size = math.MaxInt32
+		c.Type = field.TypeString
+	case "varchar":
+		c.Type = field.TypeString
+		if len(parts) > 1 {
+			size, err := strconv.ParseInt(parts[1], 10, 64)
+			if err != nil {
+				return fmt.Errorf("converting varchar size to int: %v", err)
+			}
+			c.Size = size
+		}
+	case "json":
+		c.Type = field.TypeJSON
+	case "uuid":
+		c.Type = field.TypeUUID
+	case "enum":
+		c.Type = field.TypeEnum
+		c.Enums = make([]string, len(parts)-1)
+		for i, e := range parts[1:] {
+			c.Enums[i] = strings.Trim(e, "'")
+		}
+	default:
+		return fmt.Errorf("unknown column type %q", parts[0])
+	}
+
+	return nil
+}
+
 // addColumn returns the DSL query for adding the given column to a table.
 func (d *SQLite) addColumn(c *Column) *sql.ColumnBuilder {
 	b := sql.Column(c.Name).Type(d.cType(c)).Attr(c.Attr)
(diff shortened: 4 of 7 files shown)


erikh:
@a8m at the risk of spamming you, here are some highlights in the existing work you should take note of. I have the FK work to do still and a few tests, but otherwise I think everything is functioning.

Note that due to how sqlite3 lib handles parameters, I don't think using a `?` or binding syntax will work in the pragmas. I think since we're working totally on trusted input that a sprintf is safe, but I would like to hear your input.

https://github.com/facebookincubator/ent/pull/221/files#diff-39e4b302014280cdc713d8c1a32acb2eR140 -- this tinyint(1) special case for mysql is perpetuated here, do we want that?

https://github.com/facebookincubator/ent/pull/221/files#diff-56a4aef6402a64fbe636f02a23d8fa57R22 better place for this test?

also the following refactors were made:

https://github.com/facebookincubator/ent/pull/221/files#diff-932a340b0ff6f22c5b5a59ce1e8a4972R434 - first function is taken from mysql driver and the second is taken from pg, these are just utility functions to perform standard operations I needed. I hope the moves are ok.

erikh:
Additionally, it doesn't seem like constraints are uniquely named throughout sqlite3, leading to an issue i've pushed here:

you can see there is no globally unique identifier. what matches against the name provided to `fkExist`?

```
sqlite> create table foo (bar int primary key);
sqlite> create table quux (baz int references foo (bar));
sqlite> .schema
CREATE TABLE foo (bar int primary key);
CREATE TABLE quux (baz int references foo (bar));
sqlite> pragma foreign_key_list('quux');
0|0|foo|baz|bar|NO ACTION|NO ACTION|NONE
sqlite> .headers on
sqlite> pragma foreign_key_list('quux');
id|seq|table|from|to|on_update|on_delete|match
0|0|foo|baz|bar|NO ACTION|NO ACTION|NONE
```

I'm committing some code now that shows my problem; read the fkexist method. 

Thanks again for your patience with me. :)

a8m:
Hey @erikh, thanks for working on this!
I'll review this on Sunday (weekend in Israel is Friday–Saturday).

erikh:
No rush; I'll keep chipping away at it.

a8m:
> @a8m at the risk of spamming you, here are some highlights in the existing work you should take note of. I have the FK work to do still and a few tests, but otherwise I think everything is functioning.

You're not spamming at all. Thanks for working on this.
If you prefer to work with smaller PRs (like, only table scanning and them reading indexes, etc..) that's fine too. 

> Note that due to how sqlite3 lib handles parameters, I don't think using a ? or binding syntax will work in the pragmas. I think since we're working totally on trusted input that a sprintf is safe, but I would like to hear your input.

I think `Sprintf` is fine here. 

> https://github.com/facebookincubator/ent/pull/221/files#diff-39e4b302014280cdc713d8c1a32acb2eR140 -- this tinyint(1) special case for mysql is perpetuated here, do we want that?

I don't think so. I think it's OK to support only `bool` and `boolean`, because these are the common types in the ORMs outside (feel free to correct me if I have a wrong assumption). 

> https://github.com/facebookincubator/ent/pull/221/files#diff-56a4aef6402a64fbe636f02a23d8fa57R22 better place for this test?

`entc/integration/migrate/migrate_test.go`

> you can see there is no globally unique identifier. what matches against the name provided to fkExist?

Let's change the method then to:
```go
fkExist(ctx context.Context, tx dialect.Tx, table, fk string) (bool, error) 
```

---

Let me know if you plan to break it to smaller PRs, or I'll just review it. 
Thanks again for your contribution.

erikh:
I'll probably put more in on this today. This might be up for a while as I
continue to chip on it; hope that's ok. Thanks for the feedback!

erikh:
I'm in process of porting the postgres table migration tests over to sqlite3 but your other change suggestions have been implemented.

erikh:
ok the migration tests are up as well as some related fixes. I'll fix up any tests next, but what should I be doing with this patch after this? writing more tests?

erikh:
how do I safely run the integrations? is there a doc for it?

erikh:
nvm found it

erikh:
https://www.sqlite.org/lang_altertable.html#otheralter is going to be necessary for the FK moves and other operations on constraints such as changing the on delete or update triggers.

Where would this code best live?

a8m:
> https://www.sqlite.org/lang_altertable.html#otheralter is going to be necessary for the FK moves and other operations on constraints such as changing the on delete or update triggers.

Maybe I miss something, but why do we need to create a temporary table if we can create FKs as mentioned in https://github.com/facebookincubator/ent/issues/200#issuecomment-560369316?

erikh:
it's when FKs need to be migrated or altered that it'll be an issue. if you
are not adding the column -- for example adding a foreign key to an
existing column -- you have to perform this operation.

a8m:
> for example adding a foreign key to an existing column -- you have to perform this operation.


Currently, a FK is added with its column to the schema file, when a new edge is added to the `ent.Schema`. Meaning, we don't have situations that a FK is added to a column that already exists in the table. 

We can add support for this (I don't really mind), but I think it adds much more complexity for a feature/functionality that we don't really need.

Let me know what do you think.

erikh:
Ok yeah, if you don't think we'll need it I'll review and make sure none of
our tests require it... I think that's where I got blocked last time I
checked in on it.

I'll review later tonight and get back to you, on another project atm.

erikh:
Sorry, I haven't had time to get to this over the week. I'll try to polish this off next week.

erikh:
This should pass all tests now; I've squashed the patch and I think it's ready for at least an initial review. I'm sure there are still some things outstanding.

erikh:
some notes:

* we now pass *Table and *ForeignKey objects directly to the fkExist interface method; it's simpler this way. let me know if you object.
* since there is no symbol in sqlite for FK columns, we simply match the parameters. I may be missing a few that are necessary for ent, please advise.

erikh:
one last thing; I just don't know if this is going to behave in the following scenario:

* you define a schema with two tables, each with a Node and no edges
* migrate
* define an edge between the two existing nodes

This is probably not a problem for very simple situations but could be in the future.

erikh:
@a8m what should I do with this patch?

a8m:
Hey @erikh, sorry for the delay. I missed the previous messages. 

I'll review this later today (or tomorrow). 
Thanks!

dialect/sql/builder.go
```diff
@@ -254,7 +254,11 @@ func (t *TableAlter) DropColumn(c *ColumnBuilder) *TableAlter {
 
 // AddForeignKey adds a foreign key constraint to the `ALTER TABLE` statement.
 func (t *TableAlter) AddForeignKey(fk *ForeignKeyBuilder) *TableAlter {
-	t.Queries = append(t.Queries, &Wrapper{"ADD CONSTRAINT %s", fk})
+	if t.Dialect() == dialect.SQLite {
+		t.Queries = append(t.Queries, &Wrapper{"ADD COLUMN %s", fk})
+	} else {
+		t.Queries = append(t.Queries, &Wrapper{"ADD CONSTRAINT %s", fk})
+	}
```
a8m: IMO this logic should reside in the schema migration package.

I'm thinking about 2 options (but you're welcome to suggest other alternatives):
1. Add an `addFK` method to the `sqlDialect` interface that generates a different command for the `sqlite3` dialect.
2. Change `fkExist` logic in `sqlite3` to always returns `true` (as it was before), and append the [foreign-key-clause](https://sqlite.org/syntax/foreign-key-clause.html) to the new column we create. For example:
   ```sql
   ALTER table t1 ADD COLUMN c1 int REFERENCES t2(c2) ON DELETE SET NULL;
   ```

Let me know what do you think about it.

dialect/sql/builder.go
```diff
@@ -333,17 +337,24 @@ func (fk *ForeignKeyBuilder) OnUpdate(action string) *ForeignKeyBuilder {
...
+
+	if fk.Dialect() == dialect.SQLite {
+		fk.Pad().Ident(fk.columns[0])
+	} else {
+		fk.WriteString("FOREIGN KEY")
+		fk.Nested(func(b *Builder) {
+			b.IdentComma(fk.columns...)
+		})
+	}
+
```
a8m: ditto

dialect/sql/schema/migrate.go
```diff
@@ -235,6 +236,14 @@ type changes struct {
 // changeSet returns a changes object to be applied on existing table.
 // It fails if one of the changes is invalid.
 func (m *Migrate) changeSet(curr, new *Table) (*changes, error) {
+	if curr == nil {
+		return nil, errors.New("current state could not be determined during change set generation")
+	}
+
+	if new == nil {
+		return nil, errors.New("determined state could not be determined during change set generation")
+	}
```
a8m: Use this format: `sql/schema: determined state could ...`.
 
Also, maybe add the table name as well?

dialect/sql/schema/mysql.go
```diff
@@ -240,9 +240,8 @@ func (d *MySQL) scanColumn(c *Column, rows *sql.Rows) error {
 	if nullable.Valid {
 		c.Nullable = nullable.String == "YES"
 	}
-	switch parts := strings.FieldsFunc(c.typ, func(r rune) bool {
-		return r == '(' || r == ')' || r == ' ' || r == ','
-	}); parts[0] {
+
+	switch parts := typeFields(c.typ); parts[0] {
```
a8m: nice!

dialect/sql/schema/sqlite.go
```diff
@@ -119,6 +123,52 @@ func (*SQLite) cType(c *Column) (t string) {
...
+			if err != nil {
+				return fmt.Errorf("converting varchar size to int: %v", err)
+			}
+			c.Size = size
+		}
+	case "json":
+		c.Type = field.TypeJSON
+	case "uuid":
+		c.Type = field.TypeUUID
+	case "enum":
```
a8m: I don't think this clause is needed.

dialect/sql/schema/sqlite.go
```diff
@@ -119,6 +123,52 @@ func (*SQLite) cType(c *Column) (t string) {
...
+		c.Type = field.TypeInt32
+	case "smallint":
+		c.Type = field.TypeInt16
+	case "bigint":
+		c.Type = field.TypeInt64
+	case "tinyint":
+		c.Type = field.TypeInt8
+	case "double", "real":
+		c.Type = field.TypeFloat64
+	case "timestamp", "datetime":
```
a8m: > "timestamp"

What is usually used by other ORMs? I know that `date` and `datetime` types exist in `sqlite`, but not familiar with `timestamp`.

dialect/sql/schema/sqlite_test.go
```diff
@@ -128,6 +161,443 @@ func TestSQLite_Create(t *testing.T) {
...
+						AddRow(1, "name", "varchar", 0, nil, 0))
+				mock.ExpectQuery(escape("pragma index_list('users')")).
+					WillReturnRows(sqlmock.NewRows([]string{"seq", "name", "unique", "origin", "partial"}).
+						AddRow(0, "sqlite_autoindex_users_1", 1, "pk", 0))
+				mock.ExpectExec(escape("ALTER TABLE `users` ADD COLUMN `spouse_id` integer NULL")).
+					WillReturnResult(sqlmock.NewResult(0, 1))
+				mock.ExpectQuery(escape("pragma foreign_key_list('users')")).
+					WillReturnRows(sqlmock.NewRows([]string{"id", "seq", "table", "from", "to", "on_update", "on_delete", "match"}))
+				mock.ExpectExec("ALTER TABLE `users` ADD COLUMN `spouse_id` REFERENCES `users`\\(`id`\\) ON DELETE CASCADE").
+					WillReturnResult(sqlmock.NewResult(0, 1))
```
a8m: Great testing work!

dialect/sql/schema/mysql.go
```diff
@@ -240,9 +240,8 @@ func (d *MySQL) scanColumn(c *Column, rows *sql.Rows) error {
 	if nullable.Valid {
 		c.Nullable = nullable.String == "YES"
 	}
-	switch parts := strings.FieldsFunc(c.typ, func(r rune) bool {
-		return r == '(' || r == ')' || r == ' ' || r == ','
-	}); parts[0] {
+
```
a8m:
```suggestion
```

dialect/sql/schema/migrate.go
```diff
@@ -235,6 +236,14 @@ type changes struct {
 // changeSet returns a changes object to be applied on existing table.
 // It fails if one of the changes is invalid.
 func (m *Migrate) changeSet(curr, new *Table) (*changes, error) {
+	if curr == nil {
+		return nil, errors.New("current state could not be determined during change set generation")
+	}
+
```
a8m:
```suggestion
```

dialect/sql/schema/sqlite.go
```diff
@@ -148,5 +198,143 @@ func (d *SQLite) dropIndex(ctx context.Context, tx dialect.Tx, idx *Index, table
...
+			onUpdate  string
+			onDelete  string
+			match     string
+		)
+		if err := rows.Scan(&id, &seq, &tableName, &from, &to, &onUpdate, &onDelete, &match); err != nil {
+			return false, errors.Wrap(err, "while querying foreign keys")
+		}
+
+		if tableName == fk.RefTable.Name && fk.Columns[0].Name == from && fk.RefColumns[0].Name == to {
+			return true, rows.Close()
```
a8m: Why not:
```suggestion
			return true, nil
```

You already call `defer rows.Close()` above.

dialect/sql/schema/sqlite.go
```diff
@@ -148,5 +198,143 @@ func (d *SQLite) dropIndex(ctx context.Context, tx dialect.Tx, idx *Index, table
...
+		}
+	}
+
+	return false, rows.Close()
```
a8m: ditto

erikh:
ok. it's a holiday here in the US so I'll be addressing this later this
week; thanks for the review!

dialect/sql/builder.go
erikh: ok, that makes sense. I think I get the general gist of what your'e suggesting:

```go
func (s *SQLite) addFK(t *Table, *fk ForeignKey) string {
   // construct string to add foreign key
}

func (s *SQLite) fkExist(t *Table, *fk ForeignKey) bool {
   return true
}
```

dialect/sql/schema/migrate.go
erikh: ok. Would you like wrappers for these package name identifiers usable for `github.com/pkg/errors.Wrap`? Could do this in a later PR.

dialect/sql/schema/sqlite.go
erikh: timestamp is used in postgres at least -- it's SQL 92 IIRC. There is also `timestamp with time zone`. the NOW() function typically returns a timestamp of some fashion in SQL.

I need to do a second pass and make sure all the types line up, I think I took this from the mysql driver. Sorry if reviewing this caused you a headache.

erikh:
Thanks for the review. I'll try to turn this around over the weekend. I assume I probably won't hear from you until after the new year. ;) Enjoy your time.

erikh:
I haven't forgotten about this; just haven't had much time this week. Thanks and sorry.

dialect/sql/builder.go
erikh: @a8m ok after some noodling on this (and re-reading what you said, multiple times, sorry about missing the message there!):

* addFK leads to import cycles best I can see without creating a third package, but I'm still playing around with this approach and it might prove fruitful. I can't share the migrate package with the builder directly because of the relationship that already exists. I'm guessing I'm missing something here and will report back after more searching around.

* setting things to true works fine, but then the alter cannot work at all unless I am still misunderstanding? I'm not sure is the best approach to adding edges to existing tables. I've implemented most of this and the clear problem with it is that it will never work on the second migration to the same table as the user probably would expect (with fks populated).

I think what might make sense here (but need to examine a little more) is altering the addColumn interface method to accept a foreignkey struct, or otherwise relate the foreign keys to the columns being currently built (which is true for FKs but not the columns that they are related to). This way I can deliberately say "I want to ensure my foreign key relationships are always added as a part of my column mutations to tables, no other way". And I think, but am not 100% certain, this way the work in fk.DSL() can either be simplified or removed for SQLite's case and we can just match/query the foreign keys directly in the Column portion of the migration in the tBuilder method and simplify the building process here.

WDYT?

Also sincerely sorry for the uber-high latency on this patch, I'll try to button things up this week.

a8m: > I think what might make sense here (but need to examine a little more) is altering the addColumn interface method to accept a foreignkey struct, or otherwise relate the foreign keys to the columns being currently built (which is true for FKs but not the columns that they are related to).

I think that's fine (as I mentioned in https://github.com/facebookincubator/ent/pull/221#discussion_r361163803). If you want to go with this approach, I can help with that and add the code that links the columns with their foreign-keys.  

WDYT?

a8m:
> I haven't forgotten about this; just haven't had much time this week. Thanks and sorry.

I can help you with that if you want. Just split it to multiple PRs, so we can merge all the ready parts.

erikh:
Between a variety of personal conflicts I can't really spend too much more time on this, and I feel bad just leaving it here. Happy to split it up over the next week if you want to take pieces of it, or you can cherry-pick what you would like to keep. I really do need this behavior though, so I'd encourage you to persist with the features this patch was intended to deliver.

Sorry man. I hate doing this to people but life is just getting too busy.

erikh:
Let me know how you want it split up and I will do the work.

Sloaix:
How much time will it take to merge this pr?

This feature is important for me, i'm using ent instead of xorm.

vtolstov:
now this pr needs to be fixed as it have merge conflicts, but yes, how we can help to get this merged?

a8m:
Thanks for the great work @erikh, and sorry for my delayed response. 

I'll check if I can split (and edit) it by myself and merge parts of this to `ent`. I want to make sure you get the credit for your work.

Thanks again and don't feel bad about it!

erikh:
Don't sweat credit. I appreciate it, but if the code isn't fitting well
into the current state of things, just scrap it.

Discovery notes from working on this:

I think the biggest problem with ent and sqlite revolve around the foreign
key migrations, it's going to lead to some surprising behavior as currently
implemented, when the keys just don't get managed in the destructive
migration types (the state of this patch). I would strongly suggest
implementing the table replacement procedure that sqlite specifies;
realistically, shared instances of sqlite are not going to happen often
enough for concurrency to be an issue, and I don't think most production
uses are going to be sqlite-backed anyway. For a solution though, some
other gate keeper could suffice as an out-of-band lock for controlling the
access at the right time when those situations do come, but there's no need
to one-shot a patch to this behavior. I think most sqlite users will be
satisfied with being able to migrate at all, and I don't think most of them
are using shared sqlite databases, but that's why I'm posting this here. :)

Anyways hope the flavor is welcome, and sorry I haven't had time. I'm
between a lot of things personally and I just kept putting it off because
other things were taking control, and it seems like a foregone conclusion
now. I hope there's no frustration on the lack of delivery.

Thanks,

@erikh

a8m:
#428 was merged and it adds the first basic migration functionality for SQLite (adding resources).
Stuff like dropping/modifying columns or constraints are currently disabled in SQLite and will be added in the future. 

I want to thank you @erikh for this amazing work and for being active and involved in this project!

PR #245, opened 2019-12-15
dialect/sql/schema: drop unique constraint on pg migration

### LINKED ISSUE #235: SQL contraints are not removed when field.Unique() is removed
If you define a field with a unique constraint it will properly set to the data-model
```
return []ent.Field{
   field.String("name").Unique(),
}
```

Now, remove the constraint:
```
return []ent.Field{
   field.String("name"),
}
```
That change will not be propagated to the database. You can still see the constraint defined:
```
CONSTRAINT users_name_key UNIQUE (name)
```
I observed this behavior with postgres. Not tested with Mysql.

a8m:
Fixes #235

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -258,6 +258,12 @@ func (t *TableAlter) AddForeignKey(fk *ForeignKeyBuilder) *TableAlter {
 	return t
 }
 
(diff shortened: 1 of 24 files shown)


chris-rock:
Great work @a8m Thank you!

Issue #215, opened 2019-12-03
Feature request: Cursor-based pagination


chris-rock:
The tool already includes support for pagination, but its limited to offset pagination which is known to be less optimal for fast-changing datasets. Support for cursor-based pagination would be awesome.

a8m:
Hey @chris-rock, 
We actually discussed about this internally a few months and kinda agreed that we want to add a cursor-based API.

I think we'll address this after [v1](https://github.com/facebookincubator/ent/issues/46) is finished.

chris-rock:
That sounds awesome. Thank you @a8m

chris-rock:
@a8m I implemented [keyset pagination](https://use-the-index-luke.com/no-offset) with ent by using the following code:

```
func CompositeWhere(operator string, columns []string, args ...interface{}) func(*sql.Selector) {
	return func(s *sql.Selector) {
		total := s.Total()

		b := &sql.Predicate{}
		b.SetDialect(s.Dialect())
		b.Nested(func(nb *sql.Builder) {
			nb.IdentComma(columns...)
		})
		b.WriteString(operator)
		b.WriteString("(")
		b.Args(args...)
		b.WriteString(")")

		s.Where(b)
		s.SetTotal(total + b.Total())
	}
}

func CursorAsc(columns []string, args ...interface{}) func(s *sql.Selector) {
	const operator = " > "
	return CompositeWhere(operator, columns, args...)
}

func CursorDesc(columns []string, args ...interface{}) func(s *sql.Selector) {
	const operator = " < "
	return CompositeWhere(operator, columns, args...)
}
```
The code can be used then as follows:
```
// fetch cursor
cursorUser, err = client.Advisory.Query().Where(user.SubjectEQ(q.Page.Cursor)).Only(ctx)

// set fields for order and determine the query for the cursor
fieldOrder = client.Desc(user.FieldName)
idOrder = client.Desc(user.FieldSubject)
cursor = CursorDesc(Columns(user.FieldName, user.FieldId), cursorUser.Name, cursorUser.Subject)

data, err := client.Advisory.Query().
  Where(cursor).
  Order(fieldOrder, idOrder).
  Limit(pageSize).
  All(ctx)
```

I am happy to contribute if you think this is useful for this project.

a8m:
Wonderful! I'm going to give it a look this weekend!
Thanks :)

sohaha:
Are there any new developments?

a8m:
[...]

PR #264, opened 2019-12-30
add predicates for composite where query

chris-rock:
Implements the predicates for composite where queries. It is the first step towards key-set pagination support as discussed in #215

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -1069,9 +1069,7 @@ func CompositeLT(columns []string, args ...interface{}) *Predicate {
 	return (&Predicate{}).CompositeLT(columns, args...)
 }
 
-// GT returns a composite ">" predicate.
-func (p *Predicate) CompositeGT(columns []string, args ...interface{}) *Predicate {
-	const operator = " > "
+func (p *Predicate) compositeP(operator string, columns []string, args ...interface{}) *Predicate {
 	return p.append(func(b *Builder) {
 		b.Nested(func(nb *Builder) {
 			nb.IdentComma(columns...)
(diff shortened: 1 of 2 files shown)


chris-rock:
CLA is signed, I am not sure why the fb is not posting an update here

a8m:
Hey @chris-rock, really thanks for the contribution. 
I'll review this later this week (I hope tomorrow).

dialect/sql/builder.go
```diff
@@ -1061,6 +1061,46 @@ func (p *Predicate) ContainsFold(col, sub string) *Predicate {
 	})
 }
 
+func Columns(columns ...string) []string {
```
chris-rock: Should we remove `Columns`? This could probably be done by:

```
t1 := Table("users")
t1.Columns("name", "age")
```

dialect/sql/builder_test.go
```diff
@@ -1063,6 +1063,50 @@ func TestBuilder(t *testing.T) {
...
 		},
+		{
+			input: func() Querier {
+				t1 := Table("users")
+				return Dialect(dialect.Postgres).
+					Select().
+					From(t1).
+					Where(CompositeGT(t1.Columns("id", "name"), 1, "Ariel"))
+			}(),
+			wantQuery: `SELECT * FROM "users" WHERE ("users"."id", "users"."name") > ($1, $2)`,
```
a8m: @chris-rock, I'm not familiar with this syntax - `(columns..ץ) OP (args...)`.
Would you mind add here a reference to docs, or explain how does it work (or what does it mean)? Also, the support in the different dialects (SQLite, MySQL and PostgreSQL).

dialect/sql/builder.go
```diff
@@ -1061,6 +1061,42 @@ func (p *Predicate) ContainsFold(col, sub string) *Predicate {
...
+		b.Nested(func(nb *Builder) {
+			nb.IdentComma(columns...)
+		})
+		b.WriteString(operator)
+		b.WriteString("(")
+		b.Args(args...)
+		b.WriteString(")")
+	})
+}
+
```
a8m: Maybe, create a `compositeP` function (like you did in #215), and apply it with the different operator?

dialect/sql/builder_test.go
chris-rock: The inspiration for the implementation came from https://use-the-index-luke.com/no-offset. I have tested the syntax in Sqlite and Postgres. [SQLite Example](https://www.sqlite.org/rowvalue.html#scrolling_window_queries):
```
SELECT * FROM contacts
 WHERE (lastname,firstname) > (?1,?2)
 ORDER BY lastname, firstname
 LIMIT 7;
```

chris-rock: The syntax is available for a long time, another good example: https://www.jooq.org/doc/3.12/manual/sql-building/sql-statements/select-statement/seek-clause/

a8m:
Thanks for your contribution @chris-rock!

jufemaiz:
QQ: how does one use this in anger?

PR #268, opened 2020-01-05
fix meta.tmpl code generating incomplete issue

kidlj:
In meta.tmpl, there is an issue when mixins field has validators, the generated code is incomplete.

mixins_validator

The screenshot shows the generated code taking the [official mixins demo code](https://entgo.io/docs/schema-mixin/) as an example.

This pull request is a fix.


a8m:
Thanks for your contribution @kidlj!

PR #276, opened 2020-01-08
Reserved keywords when using enums

danf0rth:
Fixing https://github.com/facebookincubator/ent/issues/275

Probably the best solution what i found to keep names and database entities as they are and fix arguments or types with suffix.


entc/gen/template/meta.tmpl
```diff
@@ -124,13 +123,14 @@ var (
 		}
 
 		{{ $name := $f.Validator -}}
-		// {{ $name }} is a validator for the "{{ $f.Name }}" field enum values. It is called by the builders before save.
-		func {{ $name }}({{ $f.Name }} {{ $enum }}) error {
-			switch {{ $f.Name }} {
+		{{ $arg := sanitize $f.Name }}
```
a8m:
```suggestion
		{{ $arg := receiver $f.Name }}
```

Use the existing `receiver` function. It should cover these cases.

danf0rth:
@a8m btw, `receiver` will not helps when someone specify "go" in enum name (and maybe some other keywords also).

a8m:
Oh, I see. But other validators use the `BuilderField` function. See: [meta.tmpl](https://github.com/facebookincubator/ent/blob/master/entc/gen/template/meta.tmpl#L80)

danf0rth:
It seems that this is the best solution, so i removed unused function that i added in my first commit. Hope is it ok now :) Thanks for your advices.

entc/gen/template/meta.tmpl
```diff
@@ -100,7 +100,6 @@ var (
 		{{ $enum := trimPackage $f.Type.String $.Package }}
 		// {{ $enum }} defines the type for the {{ $f.Name }} enum field.
 		type {{ $enum }} string
-
```
a8m: Can you add back please this line? Thanks

danf0rth: Added back :)

PR #294, opened 2020-01-18
add json tag to edges

aca:
Adding json tag for "Edge" field, just like "id" field (which is both auto-generated by ent)  will be really useful when building json rest api.

Furthermore, It will be really helpful it we can customize struct tag for id, edge field just like other fields and even edges.


aca:
cla signed. Don't now how I get it updated

aca:
@hantmac Can you elaborate it more?. JSON tag for fields is already available by default and we can even customize it with structtag method. (except for id field which is hard corded). And even edges also have structtag method.

I believe "Edges" should have proper json tag just like "ID" field. just like this.

```
// Repository is the model entity for the Repository schema.
type Repository struct {
	config `json:"-"`
	// ID of the ent.
	ID int `json:"id,omitempty"`
	// Name holds the value of the "name" field.
	Name string `json:"name,omitempty"`
	// Description holds the value of the "description" field.
	Description *string `json:"description,omitempty"`
	Edges struct {
		// Namespace holds the value of the namespace edge.
		Namespace *Namespace `json:"namespace"`
	} `json:"edges,omitempty"`
	namespace_id       *int
}
```

hantmac:
> @hantmac Can you elaborate it more?. JSON tag for fields is already available by default and we can even customize it with structtag method. (except for id field which is hard corded). And even edges also have structtag method.
> 
> I believe "Edges" should have proper json tag just like "ID" field. just like this.
> 
> ```
> // Repository is the model entity for the Repository schema.
> type Repository struct {
> 	config `json:"-"`
> 	// ID of the ent.
> 	ID int `json:"id,omitempty"`
> 	// Name holds the value of the "name" field.
> 	Name string `json:"name,omitempty"`
> 	// Description holds the value of the "description" field.
> 	Description *string `json:"description,omitempty"`
> 	Edges struct {
> 		// Namespace holds the value of the namespace edge.
> 		Namespace *Namespace `json:"namespace"`
> 	} `json:"edges,omitempty"`
> 	namespace_id       *int
> }
> ```

I got you, you are right but golang  structure combination is not standard. It Should like this:
type Repository struct {
	config `json:"-"`
	// ID of the ent.
	ID int `json:"id,omitempty"`
	// Name holds the value of the "name" field.
	Name string `json:"name,omitempty"`
	// Description holds the value of the "description" field.
	Description *string `json:"description,omitempty"`
	Edges
	namespace_id       *int
}

I think golang's structure combination method should be better like this.

aca:
@hantmac 
I don't understand what you meant. 
If "structure combination" you mentioned is about inline type definitions, It is widely used to express nested data structure (as far as I know).
Ent already generates models like above and populate edge data when eager loading is used.

for example
```
{
  "headers": {
    "Accept": "application/json"
  },
  "origin": "180.231.153.82, 180.231.153.82",
  "url": "https://httpbin.org/get"
}
```
```
type AutoGenerated struct {
	Headers struct {
		Accept string `json:"Accept"`
	} `json:"headers"`
	Origin string `json:"origin"`
	URL    string `json:"url"`
}
```

hantmac:
> @hantmac
> I don't understand what you meant.
> If "structure combination" you mentioned is about inline type definitions, It is widely used to express nested data structure (as far as I know).
> Ent already generates models like above and populate edge data when eager loading is used.
> 
> for example https://mholt.github.io/json-to-go/
> 
> If I convert entity to json, It currently generates "Edges" field even without json tag (As it is default by encoding/json)
> 
> In my case, I generate openapi component directly from ent model. It fails to parse Edge field as it dosn't include json tag.

Got it! I misunderstood you at first.

alexsn:
What's the current marshaling look like without adding those tags?

entc/gen/template/ent.tmpl
```diff
@@ -28,7 +28,7 @@ type {{ $.Name }} struct {
 				// {{ $e.StructField }} holds the value of the {{ $e.Name }} edge.
 				{{ $e.StructField }} {{ if not $e.Unique }}[]{{ end }}*{{ $e.Type.Name }} {{ with $e.StructTag }}`{{ . }}`{{ end }}
 			{{- end }}
-		}
+		} `json:"edges,omitempty"`
```
a8m: `omitempty` is irrelevant here, because `Edges` is a non-empty value (even if no edges were loaded).
```suggestion
		} `json:"edges"`
```

aca: Missed it. removed it.

aca:
@alexsn yes, sort of consistency.

So let me elaborate my use case.
In my project, I build JSON rest api service.
Instead of wrapping or customizing api response, I just use the query database and response directly using ent generated model.

And to use support openapi spec, I'm converting "entc generated model" to "openapi component" with this library. https://github.com/getkin/kin-openapi
It inspect struct and json tag to generate openapi spec component.

if we marshal it to json output without struct tag, go create tag with same name. "Edges".
But that library just ignore the field without struct tag.

I know it may sound like that it should be handled from that library. 
But as ent supports json struct tag by default. It will be very helpful if "edges" field also has json tag.

aca:
currently
```
{
    "Edges": {},
    "id": 2,
    "name": "fb"
}
```

changes to
```
{
    "edges": {},
    "id": 2,
    "name": "fb"
}
```

PR #295, opened 2020-01-18
add function to get all enum values

danf0rth:
Added small helper function to get all enum values.


alexsn:
Please explain the use case here. Why is this needed?

danf0rth:
@alexsn because enums is constrained list of known values, this can be used to find exclusions of not-yet-used enum values. E.g., if you have list of topics the user is following, and you add another one topic (enum value), you can filter out which topic user not yet subscribed. You just query all topics user subscribed, and then exclude one-by-one from entire list of values. This is not so frequent case, but if it happens, there is only two (okay, three) ways to get list of all possible enum values:

1. Have access to automatically generated list (such as in this PR)
2. Manually maintained list (with all the pros and cons)
3. Use `SHOW COLUMNS FROM table LIKE column`, and parse it to get actual list (but why?)

For my case, i am using days of week, stored in enums, so it's not so difficult to create such function with hardcoded values because days of weeks are never changed.

If this is not so required feature, it's okay to close this PR to keep API more cleaner and do not overload.

Thanks!

entc/gen/template/meta.tmpl
```diff
@@ -134,6 +134,15 @@ var (
...
+        // {{$enum}}All returns all values of {{$enum}} enum
+        func {{$enum}}All() []{{$enum}} {
+                return []{{$enum}}{
+                    {{- range $_, $e := $f.Enums}}
+                        {{ $f.EnumName $e }},
+                    {{- end}}
+                }
+        }
 	{{ end }}
 {{ end }}
```
a8m: If we decide to go with this, I think it should be an array of values and not a function. 
If we won't, you can always add this tiny logic to your codegen by using templates (the one below:  `"meta/additional"`). 

Thanks for your contribution!

danf0rth: So, based on your advice codegen should look like:

```
var StateAll = [...]State{
    StateOn,
    StateOff,
}
```
?

The reason is why this was wrapped into function is because slices and arrays are mutable, so them can accidentally be changed.

Also not sure about naming. Is it okay to just suffix a enum name with `All`?

a8m: > The reason is why this was wrapped into function is because slices and arrays are mutable, so them can accidentally be changed.

I agree. 


> You can always add this tiny logic to your codegen by using templates

About my previous comment -  I don't know what @alexsn thinks about it, but it looks to me as something that can be solved easily with an external template. If you are not sure how to use it, here's an example - [entc/integration/template](https://github.com/facebookincubator/ent/blob/master/entc/integration/template/ent/generate.go).

danf0rth: @a8m okay, i will close this, better to keep this for users implementation.

a8m: Thanks for the understanding @danf0rth 
I usually try to push stuff to the base of the codegen after I found them widely used.

I can share with you that @alexsn and I have discussions about creating an `ent-integration` repo for `ent`, so we can share all the other stuff we do that no part of the base codegen. For example, GraphQL Node API, [Relay pagination](https://facebook.github.io/relay/graphql/connections.htm), etc.  

This one, can reside there as well.

PR #298, opened 2020-01-20
fix eagerloading

ichord:
while parent nodes is zero.

~~~go
query.Where(predicate.{{ $e.Type.Name }}(func(s *sql.Selector) {
    s.Where(sql.InValues({{ $.Package }}.{{ $e.ColumnConstant }}, fks...))
}))
neighbors, err := query.All(ctx)
~~~

https://github.com/facebookincubator/ent/blob/259bf07b5d30dbf3a104631245375b51b91cf75e/entc/gen/template/dialect/sql/query.tmpl#L286~#L289

`query.All(ctx)` in above lines will return an error `pq: syntax error at end of input`.
The deeper issue is that `sql.InValues` doesn't generate a valid sql while `fks` is empty.

I think it's better to no reach to the database  if there are no parent nodes, so that just skip it.
I can't figure out where to write the test, please tell me.

And If you are going to fix it on your own, just reply and close it, so that I don't have to sign the CLA.


a8m:
> And If you are going to fix it on your own, just reply and close it, so that I don't have to sign the CLA.

I would be happy to see you contribute this code.
Thanks again!

ichord:
> I would be happy to see you contribute this code.
> Thanks again!

ok, I'm on it.
I like this project. 😁

ichord:
signed CLA.

entc/gen/template/dialect/sql/query.tmpl
```diff
@@ -55,6 +55,11 @@ func ({{ $receiver }} *{{ $builder }}) sqlAll(ctx context.Context) ([]*{{ $.Name
 	if err := sqlgraph.QueryNodes(ctx, {{ $receiver }}.driver, _spec); err != nil {
 		return nil, err
 	}
+
```
a8m: Remove these spaces please.

```suggestion
```

entc/gen/template/dialect/sql/query.tmpl
```diff
@@ -55,6 +55,11 @@ func ({{ $receiver }} *{{ $builder }}) sqlAll(ctx context.Context) ([]*{{ $.Name
...
+	if len(nodes) == 0 {
+		return nodes, nil
+	}
+
```
a8m:
```suggestion
```

PR #299, opened 2020-01-20
fix error description in MaxLen

### LINKED ISSUE #291: Invalid error description for MaxLen validator
```
// MaxLen adds a length validator for this field.
// Operation fails if the length of the string is greater than the given value.
func (b *stringBuilder) MaxLen(i int) *stringBuilder {
	b.desc.Size = i
	b.desc.Validators = append(b.desc.Validators, func(v string) error {
		if len(v) > i {
			return errors.New("value is less than the required length")
		}
		return nil
	})
	return b
}

```

danf0rth:
Fixes #291


a8m:
Thanks @danf0rth!

PR #303, opened 2020-01-21
Provide a flag to disable example test generation

hantmac:
Provide a flag to disable example test generation，solve https://github.com/facebookincubator/ent/issues/297.
In my opinion,that is a very unattractive feature of a code generator to generate example_test file.

### DIFF
--- a/cmd/entc/entc.go
+++ b/cmd/entc/entc.go
@@ -113,6 +113,7 @@ func main() {
 			cmd.Flags().StringVar(&cfg.Header, "header", "", "override codegen header")
 			cmd.Flags().StringVar(&cfg.Target, "target", "", "target directory for codegen")
 			cmd.Flags().StringSliceVarP(&template, "template", "", nil, "external templates to execute")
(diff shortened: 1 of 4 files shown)


a8m:
Thanks for your contribution @hantmac, but I'm in the same opinion as the [comment from yesterday](https://github.com/facebookincubator/ent/issues/297#issuecomment-576454659):

> I'm actually in favor of deleting this example_test file completely from ent. It was created before we had a proper documentation (entgo.io).

It's a legacy file that was created for internal use, and IMO it's irrelevant anymore. I'll be happy to see this file deleted from the templates.

hantmac:
> Thanks for your contribution @hantmac, but I'm in the same opinion as the [comment from yesterday](https://github.com/facebookincubator/ent/issues/297#issuecomment-576454659):
> 
> > I'm actually in favor of deleting this example_test file completely from ent. It was created before we had a proper documentation (entgo.io).
> 
> It's a legacy file that was created for internal use, and IMO it's irrelevant anymore. I'll be happy to see this file deleted from the templates.

After thinking for a while, I think I agree with you

PR #310, opened 2020-01-22
proposal: empty slice instead of nil slice in sqlAll

aca:
When sqlAll fails to match any data, it returns nil slice (with no error).
But when marshaling nil slice to json, it returns 'null'
So, When you are writing json rest api which expects response in array. You have to do something like this.

```
func (s *server) ListOrganization(w http.ResponseWriter, r *http.Request) {
	ctx := r.Context()
	orgs, err := s.db.Organization.Query().All(ctx)
	if err != nil {
		render.Render(w, r, ErrServerError(r, err))
		return
	}
	if len(orgs) == 0 {
		render.JSON(w, r, []*ent.Organization{})
	}
	render.JSON(w, r, orgs)
}
```

It would be very helpful it sqlAll just returns empty slice instead of nil slice.


aca:
I will check the test failure later.

a8m:
LGTM! Please rebase and regenerate assets:
```
go generate ./entc/gen/...
go generate ./...
```

aca:
@a8m rebased!

aca:
hmm I will check the test

aca:
fixed.

a8m:
Thanks @aca!

PR #316, opened 2020-01-25
delete TMPDIR/ent generated by go test

hantmac:
Solve https://github.com/facebookincubator/ent/issues/307
After running go test -v to test the project, the dir ent will be generated in $TMPDIR, I think we should delete the ent after test just in case, or it will affect the test result next time in some unknown cases.


a8m:
Thanks for the PR @hantmac, I’ll review this later.

a8m:
> or it will affect the test result next time in some unknown cases.

I don't have any issue with the current test, but don't mind adding this if this bother others.

Thanks

PR #317, opened 2020-01-25
Add DebugWithContext for contextual logging 

### LINKED ISSUE #312: Feature Request: contextual logging
Sometimes it is useful to support logging with context.
E.g. I want to distinguish logs that belong to single request.

Could be achieved by allowing a logging function to receive a context:
```go
func LogCtx(fn func(ctx context.Context, i ...interface{})) Option {
	return func(c *config) {
		c.logCtx = fn
	}
}
```
Or by something like `func getLog(ctx context.Context) func(...Interface{})` function that will return logging function for provided context.

Btw, thank you for your great project.

aca:
fix #312 
This introduce new logging driver, config to enable contextual logging.
Need to configure logger which can handle context.Context with ent.LogCtx(logger).
If not, it behaves the same as Debug() with log.Println.

Examples using [zerolog](https://github.com/rs/zerolog)
```
logCtx := func(ctx context.Context, i ...interface{}) {
	log.Ctx(ctx).Print(i)
}

drv, _ := sql.Open("sqlite3", "file:ent?mode=memory&cache=shared&_fk=1")
dbgDrv := dialect.DebugWithContext(drv, logCtx)
client := ent.NewClient(ent.Driver(dbgDrv))
```

Results would look like this.
```
DBG [driver.Tx(c0ecdcd4-f94f-40b4-bee0-cc19721ae326): started] req_id=bom5rksd4kprp6u9fhp0 request="GET /todo/2"
DBG [Tx(c0ecdcd4-f94f-40b4-bee0-cc19721ae326).Query: query=SELECT DISTINCT `todos`.`id`, `todos`.`name`, `todos`.`contents` FROM `todos` WHERE `todos`.`id` = ? LIMIT ? args=[1 2]] req_id=bom5rksd4kprp6u9fhp0 request="GET /todo/2"
DBG [Tx(c0ecdcd4-f94f-40b4-bee0-cc19721ae326): rollbacked] req_id=bom5rksd4kprp6u9fhp0 request="GET /todo/2"
```


a8m:
Thanks for the PR @aca, I’ll review this later.

dialect/dialect.go
```diff
@@ -68,6 +68,12 @@ type DebugDriver struct {
 	log    func(...interface{}) // log function. defaults to log.Println.
 }
 
+// DebugDriverWithContext is a driver that logs all driver operations with context.
+type DebugDriverWithContext struct {
+	Driver                                       // underlying driver.
+	logCtx func(context.Context, ...interface{}) // log function.
+}
+
```
a8m: Maybe we can merge these 2 drivers into one as follows:

```go
// DebugDriver is a driver that logs all driver operations.
type DebugDriver struct {
	Driver                                       // underlying driver.
	log    func(context.Context, ...interface{}) // log function. defaults to log.Println.
}

// Debug gets a driver and an optional logging function, and returns
// a new debugged-driver that prints all outgoing operations.
func Debug(d Driver, logger ...func(...interface{})) Driver {
	logf := log.Println
	if len(logger) == 1 {
		logf = logger[0]
	}
	return &DebugDriver{d, func(_ context.Context, v ...interface{}) { logf(v...) }}
}

// DebugWithContext gets a driver and an logging function, and returns
// a new debugged-driver that prints all outgoing operations with context.
func DebugWithContext(d Driver, logger func(context.Context, ...interface{})) Driver {
	return &DebugDriver{d, logger}
}
```

entc/gen/template/client.tmpl
```diff
@@ -106,6 +111,42 @@ func (c *Client) Debug() *Client {
...
+//	}
+//
+//	client, _ := ent.Open("sqlite3", "file:ent?mode=memory&cache=shared&_fk=1", ent.LogCtx(logCtx))
+//
+//	client.DebugWithContext().
+//		{{ (index $.Nodes 0).Name }}.
+//		Query().
+//		Count(ctx)
+//
+func (c *Client) DebugWithContext() *Client {
```
a8m: I'm not sure I like it. 
Since, this is not so common, we can keep the API as it is today, and let people using it as follows: 

```go
func main() {
    drv, err := sql.Open("<dialect>", "<mysql-dsn>")
    if err != nil {
        return nil, err
    }
    drv := dialect.DebugWithContext(drv, logf)
    client := ent.NewClient(ent.Driver(drv))
}
```

dialect/dialect.go
aca: I was thinking the same. 
I just didn't know if it was ok to change struct signature.
I'll try to fix this.

aca:
wip

entc/gen/template/client.tmpl
aca: That would be enough. Removed config.

dialect/dialect.go
```diff
@@ -64,29 +64,37 @@ func NopTx(d Driver) Tx {
...
+	logf := log.Println
 	if len(logger) == 1 {
-		drv.log = logger[0]
+		logf = logger[0]
 	}
+	drv := &DebugDriver{d, func(_ context.Context, v ...interface{}) { logf(v...) }}
+	return drv
+}
+
+// DebugWithContext gets a driver and an logging function, and returns
```
a8m:
```suggestion
// DebugWithContext gets a driver and a logging function, and returns
```

PR #328, opened 2020-02-05
dialect/sql/schema: new mysql int column type formats

qystishere:


a8m:
I guess this is the latest mysql version (8.0.19), right?

qystishere:
Yes, latest from docker.

a8m:
Thanks for your contribution @qystishere!

> Yes, latest from docker.

So, that's not enough, because the latest version of MySQL changed the format in `INFORMATION_SCHEMA.COLUMNS`, and `COLUMN_TYPE` can be one of these formats:
```
tinyint
tinyint (X)
tinyint unsigned
tinyint (X) unsigned
```
(It's true for all numeric types).

Right now we're using `8.0.18` as latest, but I planned to fix this soon. You are welcome to extend this PR and add this support.

dialect/sql/schema/mysql.go
```diff
@@ -449,6 +435,39 @@ func (d *MySQL) tableSchema() sql.Querier {
 	return sql.Raw("(SELECT DATABASE())")
 }
 
+// columnData returns column data: parts, size and signedness
+func columnData(typ string) ([]string, int64, bool, error) {
```
a8m: `parseColumn` is more meaningful IMO. 
```suggestion
func parseColumn(typ string) ([]string, int64, bool, error) {
```

dialect/sql/schema/mysql.go
```diff
@@ -449,6 +435,39 @@ func (d *MySQL) tableSchema() sql.Querier {
...
+				return parts, 0, false, fmt.Errorf("converting %s size to int: %w", parts[0], err)
+			}
+			return parts, size, false, nil
+		case len(parts) == 3: // int(10) unsigned
+			size, err := strconv.ParseInt(parts[1], 10, 0)
+			if err != nil {
+				return parts, 0, true, fmt.Errorf("converting %s size to int: %w", parts[0], err)
+			}
+			return parts, size, true, nil
+		}
```
a8m: Let's avoid the duplication here, and change it to something like:
```go
var (
  size int
  unsigned bool
)
switch {
	case len(parts) == 2 && parts[1] == "unsigned": // int unsigned
		unsigned = true
	case len(parts) == 3: // int(10) unsigned
		unsigned = true
		fallthrough 
	case len(parts) == 2: // int(10)
		size, err = strconv.ParseInt(parts[1], 10, 0)
		if err != nil {
			return parts, 0, false, fmt.Errorf("converting %s size to int: %v", parts[0], err)
		}
}
```

(`size` and `unsigned` can be used by the second switch as well).

dialect/sql/schema/mysql.go
```diff
@@ -449,6 +435,39 @@ func (d *MySQL) tableSchema() sql.Querier {
...
+		case len(parts) == 2: // int(10)
+			size, err := strconv.ParseInt(parts[1], 10, 0)
+			if err != nil {
+				return parts, 0, false, fmt.Errorf("converting %s size to int: %w", parts[0], err)
+			}
+			return parts, size, false, nil
+		case len(parts) == 3: // int(10) unsigned
+			size, err := strconv.ParseInt(parts[1], 10, 0)
+			if err != nil {
+				return parts, 0, true, fmt.Errorf("converting %s size to int: %w", parts[0], err)
```
a8m: Please use `%v` since we still support Go 1.12.
```suggestion
				return parts, 0, true, fmt.Errorf("converting %s size to int: %v", parts[0], err)
```

qystishere:
@a8m Thanks, I fixed, why CI runs on old code? 🤔

a8m:
> @a8m Thanks, I fixed, why CI runs on old code? 

The migration test ensures that the generated API does not break when upgrading the `ent` version.
However, I forgot that `master` doesn't support `mysql:8.0.19`, so let's change back the CI to run on `8.0.18`, and I'll change it afterwards. 

Thanks!

a8m:
Thanks @qystishere!

PR #343, opened 2020-02-12
wip: custom dialect types

cliedeman:


a8m:
Hey, and thanks for your contribution.

I know it's a draft PR, but since we kinda have a design idea for this feature, I'll comment here later this week to share our thoughts.

Sorry for the delay, we're mainly focused now on adding hooks and authorization support for `ent`.

cliedeman:
@a8m Not a problem. I really like what you are doing with this library so I am excited so see the design you come up with. There PR's really help me get familiar with the guts of the project so I don't feel like its a waste if nothing comes from it.

Ciaran

a8m:
Hey @cliedeman, sorry for the delay, I'm focused on the mutation and hooks layers and didn't have time to finalize the custom fields interface.

I thought about creating an interface that will provides the following values: database-type, Go-type and predicates (maybe create a separate package named `fieldtype`). 

When creating a new `fieldtype`, a user needs to define its database type, Go-type (that implements the `driver.Valuer`, and the `sql.Scanner` interfaces). If the defined type is supporting predicates, they can be defined as well - we'll need to provide some interface for it as well (like UnaryOp, BinaryOp, etc). 

IMO, we should migrate all existing fields to it (without breaking existing programs), and support the popular database types per dialect (like pg.hstore, mysql.datetime, etc). 

One of the advantages of migrating existing types, is to allow users define database primitive types as Go typedefs. For example, use this Go type: `type StatusCode string`, and a string field in the database.

PR #352, opened 2020-02-18
Added user defined type column

cliedeman:
Add user defined column to info schema query. 
This is not consumed anywhere yet but will be used for enum types or custom types.

### DIFF
--- a/dialect/sql/schema/postgres.go
+++ b/dialect/sql/schema/postgres.go
@@ -77,7 +77,7 @@ func (d *Postgres) setRange(ctx context.Context, tx dialect.Tx, t *Table, value 
 func (d *Postgres) table(ctx context.Context, tx dialect.Tx, name string) (*Table, error) {
 	rows := &sql.Rows{}
 	query, args := sql.Dialect(dialect.Postgres).
-		Select("column_name", "data_type", "is_nullable", "column_default").
+		Select("column_name", "data_type", "is_nullable", "column_default", "udt_name").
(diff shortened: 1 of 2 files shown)


a8m:
Thanks for your contribution @cliedeman.

I'm not sure I get why it's needed, and prefer to see an implementation that uses it before adding it. 

Also, I tried to avoid the usage of `CREATE TYPE` for enums in PostgreSQL when I added its support (although, it's the right way), because it's too complicated to maintain (adding or removing values).

cliedeman:
Hi @a8m 

I put this in its own PR to keep follow up PR's smaller. I have some local code but its still very wip

So there are 2 use cases that will use this. 

Built in types:
[Range](https://www.postgresql.org/docs/11/rangetypes.html) types and other custom types like [hstore](https://www.postgresql.org/docs/11/hstore.html).

This is quite simple, the udt pretty much just replaces data_type and you fall into the standard ent column migration flow (ignoring the custom query predicates that new types would require)

Enums:
This is quite tricky to say the least.
I have implemented auto enum migration before in sqlalchemy and found it to be doable.

(I was writing a long drawn out comment about how I would implement the enum migrations but decided its not relevant. It is doable but quite complicated.)

Would you be open to adding enum support without migrations?
I make use of both in projects I would like to try using with ent. I actually prefer writing migrations by hand and also use some more unconventional gin indexes I don't ever expect  ent to handle

```    
ALTER TABLE some_table
add constraint some_table_duration_overlap
EXCLUDE USING gist (name WITH =, duration WITH &&)
```

Issue #355, opened 2020-02-21
Adding a Custom Predicate


cliedeman:
Is it possible to add custom predicates?

The append method on predicate is private

```
func CustomPredicate(key string) predicate.User {
	return predicate.User(func(s *sql.Selector) {
		p := &sql.Predicate{}
		... some custom sql...
		s.Where(p)
	})
}
```

a8m:
Hey @cliedeman, 
You're welcome to add a PR and change this.

It's likely going to change soon with the custom types support (custom types need to support user-defined predicates).

cliedeman:
This was implemented

PR #361, opened 2020-02-26
Expose custom predicate function

cliedeman:
First attempt at exposing a CustomPredicate for #355

I wanted to restrict the methods available in the custom predicate to a subset of the Builder. 
I tried to do it with an interface but I got errors due to the fluent return types


dialect/sql/builder.go
```diff
@@ -1815,6 +1821,38 @@ type Builder struct {
 	total        int           // total number of parameters in query tree.
 }
 
+// PredicateBuilder Limited builder for use in custom predicates
+type PredicateBuilder struct {
```
cliedeman: Is there a better way to do this? I tried an interface but could not get it to work

entc/integration/json/ent/user/where.go
```diff
@@ -209,3 +209,12 @@ func Not(p predicate.User) predicate.User {
 		p(s.Not())
 	})
 }
+
+// Not applies the not operator on the given predicate.
+func CustomPredicate(f func(builder sql.PredicateBuilder)) predicate.User {
+	return predicate.User(func(s *sql.Selector) {
+		s.Where(sql.P().CustomPredicate(func(builder *sql.Builder) {
+			f(sql.NewPredicateBuilder(builder))
```
cliedeman: Probably don't want to initialise this here

a8m:
If you expose (or create another) `Append` function to the `Predicate` builder, I'm pretty sure this can solve the issue:

```go
n, err = client.User.Update().Where(func(s *sql.Selector) {
    p := sql.P().Append(func(b *sql.Builder) {
        b.Ident("column").WriteString(" ILIKE ")
        b.Arg("value")
    })
    s.Where(p)
})
```

You can generate your custom predicates (for all types) using templates. Currently, there's one custom template named `"meta/additional"` that appends custom code to type's package. I think we can create another one (named `"where/additional"`) for the predicates (just for convenience).

WDYT?

entc/gen/template/where.tmpl
```diff
@@ -124,4 +124,13 @@ func Not(p predicate.{{ $.Name }}) predicate.{{ $.Name }} {
 	)
 }
 
+// CustomPredicate allows the user to write a custom predicate
```
cliedeman: Is there any to make this only generated for sql clients?

cliedeman:
So my first iteration actually looked exactly like your example 

```
n, err = client.User.Update().Where(func(s *sql.Selector) {
    p := sql.P().Append(func(b *sql.Builder) {
        b.Ident("column").WriteString(" ILIKE ")
        b.Arg("value")
    })
    s.Where(p)
})
```

but passing in the entire builder makes me nervous because its such a big api surface. 
I think this will be the first place where the user has raw access to the builder?

> You can generate your custom predicates (for all types) using templates. Currently, there's one custom template named "meta/additional" that appends custom code to type's package. I think we can create another one (named "where/additional") for the predicates (just for convenience).

This is definitely on my radar. But ideally we should be able to generate specific predicates for tables and or specific fields. 
So in my hstore example I would ideally like to have the predicates generated automatically but we don't have that level of control... yet.

cliedeman:
So in a perfect world... append would remain private and any custom predicates would be generated directly into where.go

cliedeman:
@a8m I'd like to continue working on this but its going to overlap a bit with https://github.com/facebookincubator/ent/pull/343#issuecomment-587388684.

Is there a design ready to share?

a8m:
> This is definitely on my radar. But ideally we should be able to generate specific predicates for tables and or specific fields.
So in my hstore example I would ideally like to have the predicates generated automatically but we don't have that level of control... yet.

Why not generating this using custom template? 

> Is there a design ready to share?

I'll write my comment there.

Issue #360, opened 2020-02-26
Proposal: streaming Select


covrom:
In the current version (v0.1.2), data is selected (by Select clause) into the data slice. This increases memory allocation without any reasons. In most cases, including in CRUD, streaming data further into the processing chain is sufficient. For example, CRUD in most cases transmits data only to the external interface (frontend or grpc), and if we use streaming, we can send it as a stream using the http.Flusher interface. To do this, I propose implementing a data channel return in Select clause (with errors and using the canceled context) instead of a slice.

a8m:
Thanks for the proposal @covrom. @danf0rth also [suggested](https://github.com/facebookincubator/ent/issues/46#issuecomment-576586784) to add an iterator API, and this one can be built on top of it.

We're currently focused on V1, but we can start a design discussion about it here.

covrom:
API integration is a good alternative, but it uses a single-threaded processing style, i.e. after reading the next record, the connection to the database will be idle for the duration of processing. I implemented parallelism with a buffered channel, which will be more efficient. See https://github.com/facebookincubator/ent/pull/362

hantmac:
> Iterators are good, but they use a single-threaded processing style. After reading the next record, the connection to the database will be idle for the duration of processing. I implemented concurrency with a buffered channel, which will be more efficient. See #362

How about adding a bench-mark to make a comparison?

danf0rth:
> After reading the next record, the connection to the database will be idle for the duration of processing.

Yes, this is a main feature of (lazy) iterators: load/scan data only when it is required. Not sure, but isn't the same buffers can be implemented on client side of API?

a8m:
> Iterators are good, but they use a single-threaded processing style. 

When using an iterator like interface, concurrency is just a user decision:
```go
nodes := make(chan []*ent.User, 1)
go listen(nodes)
for it.Next() {
  nodes <- it.One()
}
```

Anyway, I think both ideas are fine, and we need to figure out the right way, but we're planning to finish the hooks and the privacy support before.

covrom:
well, iterators, which are sent to the channel, sound good!

ernado:
> When using an iterator like interface, concurrency is just a user decision:
So every iteration new struct is allocated? 

E.g. `bufio.Scanner` does not allow retaining of `Bytes()` result and reuses byte slice instead of allocating it, making possible to implement a memory-efficient scanning.

If we allow retaining `it.One` result we will limit memory efficiency.

[...]

PR #375, opened 2020-03-04
Feature/iterators

covrom:
Alternative to iterators - walk over results.

#46
#360


alexsn:
Thanks for the PR @covrom, we'll review it next week.

a8m:
Hey @covrom and thanks for your contribution - Let's discuss about this proposal here (or in #360). 

When @alexsn and I discussed about this feature, we were thinking about something like the `bufio.Scanner` API (with a few changes), but `filepath.Walk` like  API is another interesting option and we need to discuss which one is preferred. So, let's talk first about pros/cons, and I'd be happy to hear more thoughts except @alexsn and I.

Also, IMO eager-loading should be supported.

PR #386, opened 2020-03-13
entc cli command to remove generated code

paulvollmer:
hello,

i added a command to the `entc` cli to remove the generated files from the given directory. for the `clean` command the path it is not the `ent/schema` directory (as it is for `entc generate`  because it will also delete the `client.go`, `config.go` etc. files.

the command will read the `schema` directory and create a list of files that need to be removed based on the files found at the directory.


a8m:
Hey @paulvollmer, and thanks for your contribution!
You know, that's crazy - I have a similar command (`entc clean`) in my `git stash`!  

Super happy to see this contribution, and I'm going to review this tomorrow. Thanks!

cmd/entc/entc.go
```diff
@@ -115,6 +115,43 @@ func main() {
...
+			Args: cobra.ExactArgs(1),
+			Run: func(cmd *cobra.Command, path []string) {
+				cleanup := make([]string, 0)
+				cleanup = append(cleanup, filepath.Join(path[0], "migrate"))
+				cleanup = append(cleanup, filepath.Join(path[0], "predicate"))
+				cleanup = append(cleanup, filepath.Join(path[0], "client.go"))
+				cleanup = append(cleanup, filepath.Join(path[0], "config.go"))
+				cleanup = append(cleanup, filepath.Join(path[0], "context.go"))
+				cleanup = append(cleanup, filepath.Join(path[0], "ent.go"))
+				cleanup = append(cleanup, filepath.Join(path[0], "tx.go"))
```
a8m: Let's change this to `filepath.Walk` and remove all files that were generated by `entc` (`go generate` signature matching).
It will be easier to maintain, and it will support deleting files that were generated with external templates as well. 

Thanks

paulvollmer:
@a8m good to hear that the PR is welcome.

at the moment the linter throw an error becasue the `main` func has more than 100 loc.

it looks like i need to move one more command  `Run` func out of it.

or any other ideas?


https://circleci.com/gh/facebookincubator/ent/3057?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

cmd/entc/entc.go
paulvollmer: okay. i didn't thought about files generated by external  templates. i start working on that.

paulvollmer: done

a8m: Please address this to all files.

You should use `filepath.Walk` under the given directory and check that the file was generated by `entc` (on package comment).

paulvollmer: what do you mean with "on package comment"? checking if the file has a line `// Code generated by entc, DO NOT EDIT.` ?

is it not better to use the `Templates` and `GraphTemplates ` array from the  `entc/gen/template.go` source as the list of files need to be deleted?

https://github.com/facebookincubator/ent/blob/dc222c0b15343920adac37c2c60d292514b50708/entc/gen/template.go#L39-L106

a8m: >  checking if the file has a line // Code generated by entc, DO NOT EDIT. ?

Yes. 

> is it not better to use the Templates and GraphTemplates array from the entc/gen/template.go source as the list of files need to be deleted?

Custom templates generate additional files (and directories).

paulvollmer: > checking if the file has a line // Code generated by entc, DO NOT EDIT. ?

i'm not sure if this works well because if a custom template do not write that line it will not be deleted. for the build in templates it is more expensive to open and check the file. we know the files the `generate` command has created.

my initial strategy was to read the `ent/schema/` directory, find all files with a `.go` extension and then checking the `ent/` directory for files that start with that name. like it is defined at the Templates array (`%s_create.go`, `%s_update.go`, etc.)


a strategy for custom templates could be to add a `templates` flag, similar to the flag available at the `generate` command to know the files we need to delete.

a8m: Sorry, I've missed this comment. 

> i'm not sure if this works well because if a custom template do not write that line it will not be deleted. 

I'm fine if it won't work for these case - it's Go convention and just needs to be documented. All the external templates we provide to users have the proper headers. 

> we know the files the generate command has created.
>
> my initial strategy was to read the ent/schema/ directory, find all files with a .go extension and then checking the ent/ directory for files that start with that name. like it is defined at the Templates array (%s_create.go, %s_update.go, etc.)

Schema files can have more than one definition - valid case. In this case, you'll need to read the schema files, find all declarations and then find their generated files.

a8m:
Hey @paulvollmer, any update with this PR?

paulvollmer:
Hi @a8m, i hadn't time the last weeks to work on the feature but now i implemented the fileWalk.

i do the following:

walk over the input path and delete all .go files with the header `// Code generated by entc, DO NOT EDIT.`

after that walk i to a second walk to delete the empty directories (`mutate`, `predicate` etc.).

paulvollmer:
@a8m after a long break at this pr i solved the merge conflict.
at the moment the linter fail becasue the main func has too many loc.

what do you think about refactoring the `cobra.Command Run` func's out of the main?

paulvollmer:
@luantranminh  @a8m  i have solved the merge conflicts and the PR is now ready to get merged.

frederikhors:
What is blocking this merge?

a8m:
Thanks for this contribution @paulvollmer, it's really appreciated. I decided to have this logic enabled by default instead of adding a new command to `ent`. See #2612

frederikhors:
You make my dreams come true, @a8m. 😄

Issue #409, opened 2020-03-29
invalid (json != longtext) when creating schema resources in mariadb 10.5


mrasong:
In mariadb 10.5, type `JSON` is an alias of `LONGTEXT`

once I defined the schema with `field.JSON`, and run client.Schema.Create(ctx), the schemas created successfully.

```go
// Fields of the Source.
func (Source) Fields() []ent.Field {
	return []ent.Field{
		field.JSON("images", []string{}).StructTag(`json:"images"`),
	}
}
```

Howerver, when I added a few fields and running `client.Schema.Create(ctx)` again, it paniced.

```
failed creating schema resources: sql/schema: changing column type for "images" is invalid (json != longtext)
```

a8m:
Hey @mrasong and thanks for raising up this issue.

We didn't test `ent` migration on MariaDB, but we should fix this. I'll try to address this ASAP.

mrasong:
thx

PR #410, opened 2020-03-29
fixed: json as an alias of longtext in mariadb

mrasong:
fixed: #409


a8m:
Thanks for your contribution @mrasong, I’ll review this later today.

dialect/sql/schema/schema.go
```diff
@@ -170,6 +170,9 @@ func (c *Column) ConvertibleTo(d *Column) bool {
 	case c.UintType() && d.IntType():
 		// uintX can not be converted to intY, when X > Y.
 		return c.Type-field.TypeUint8 <= d.Type-field.TypeInt8
+	// mariadb json == longtext
+	case c.Type == field.TypeString && d.Type == field.TypeJSON:
+		return true
```
a8m: I prefer the dialect-specific logic to be in its own file (`mysql.go`). Also, IMO it should be handled when reading the column information.

We can detect if this is a MariaDB database by parsing the `version`.

mrasong: done

mrasong:
```
&{content string longtext  2147483647  false false false <nil> [] []}
&{images string longtext  2147483647  false false false <nil> [] []}
````

it seems not works well.
If content defiened to `longtext` and images to `json`, we got both longtext.
so the last commit which can due to normal `longtext` field to `json`.
this could be a bug.

**Comment by a8m** [comment]
Hey @mrasong, please rebase your changes.

Also, according to [MariaDB docs](https://mariadb.com/kb/en/json-data-type/), it adds a `CHECK Constraints` for JSON type (`json_valid(...)`) - we can use this to check if the `longtext` column is a JSON alias.

**Comment by mrasong** [comment]
May still have a problem.

the fields in schema defined like this:

```
field.Text("content").StructTag(`json:"content"`),
field.JSON("images", []string{}).StructTag(`json:"images"`),
```

the mariadb table field like this:
```
+-------------------------+---------------+------+-----+---------+----------------+
| Field                   | Type          | Null | Key | Default | Extra          |
+-------------------------+---------------+------+-----+---------+----------------+
| id                      | bigint(20)    | NO   | PRI | NULL    | auto_increment |
...
| content                 | longtext      | NO   |     | NULL    |                |
| images                  | longtext      | NO   |     | NULL    |                |
+-------------------------+---------------+------+-----+---------+----------------+
```

when we set the `content` with a json string `{"foo":"bar"}`, the `images` with `["IMG1","IMG2"]`

```
MariaDB [test]> update source set content = '{"foo":"bar"}', images = '["IMG1","IMG2"]' where id = 3;
Query OK, 1 row affected (0.012 sec)
Rows matched: 1  Changed: 1  Warnings: 0
```

and then run the below sql.

```
MariaDB [test]> select json_valid(content),content,json_valid(images),images from source where id = 3;
+---------------------+---------------+--------------------+-----------------+
| json_valid(content) | content       | json_valid(images) | images          |
+---------------------+---------------+--------------------+-----------------+
|                   1 | {"foo":"bar"} |                  1 | ["img1","img2"] |
+---------------------+---------------+--------------------+-----------------+
1 row in set (0.008 sec)
```

we got both json valid of the two fields , actually we just want to the `images` field valid.

**Comment by mrasong** [comment]
find another way,  we can convert mariadb JSON to longtext

in `mysql.go`
```
	case field.TypeJSON:
		t = "json"
		if compareVersions(d.version, "5.7.8") == -1 {
			t = "longblob"
		}
		// mariadb json to longtext
		if strings.Contains(strings.ToLower(d.version), "mariadb") {
			t = "longtext"
		}
```

PR #455, opened 2020-04-29
entc/gen: Fix import for ID type

htdvisser:
I encountered some issues when adding UUID fields with type `ulid.ULID` imported from [github.com/oklog/ulid/v2](https://pkg.go.dev/github.com/oklog/ulid/v2?tab=doc). ~This pull request aliases imports like this if a type uses a different short package name (`ulid` in my case) than the basename of the full package path (`v2` in my case).~ This pull request fixes the missing import for the ID type.


a8m:
Hey @htdvisser, thanks for your contribution and it's great to see you here again.

It looks like something that need to be solved and I'll review it in a few hours.

schema/field/type.go
```diff
@@ -74,6 +78,22 @@ type TypeInfo struct {
...
+	case t.PkgPath == "":
+		return ""
+	case t.Ident != "":
+		if idx := strings.IndexByte(t.Ident, '.'); idx != -1 {
+			if pkgIdent := t.Ident[:idx]; pkgIdent != filepath.Base(t.PkgPath) {
+				return fmt.Sprintf(`%s "%s"`, pkgIdent, t.PkgPath)
+			}
+		}
+	}
+	return fmt.Sprintf(`"%s"`, t.PkgPath)
```
a8m, line 94: ditto
```suggestion
	return fmt.Sprintf("%q", t.PkgPath)
```

schema/field/type.go
```diff
@@ -74,6 +78,22 @@ type TypeInfo struct {
...
+// Import returns the import for the type.
+// The import has an alias if needed. The package path is wrapped in quotes.
+func (t TypeInfo) Import() string {
+	switch {
+	case t.PkgPath == "":
+		return ""
+	case t.Ident != "":
+		if idx := strings.IndexByte(t.Ident, '.'); idx != -1 {
+			if pkgIdent := t.Ident[:idx]; pkgIdent != filepath.Base(t.PkgPath) {
+				return fmt.Sprintf(`%s "%s"`, pkgIdent, t.PkgPath)
```
a8m, line 90: Just a suggestion. 
```suggestion
				return fmt.Sprintf("%s %q", pkgIdent, t.PkgPath)
```

schema/field/type.go
```diff
@@ -74,6 +78,22 @@ type TypeInfo struct {
...
 
+// Import returns the import for the type.
+// The import has an alias if needed. The package path is wrapped in quotes.
+func (t TypeInfo) Import() string {
+	switch {
+	case t.PkgPath == "":
+		return ""
+	case t.Ident != "":
+		if idx := strings.IndexByte(t.Ident, '.'); idx != -1 {
+			if pkgIdent := t.Ident[:idx]; pkgIdent != filepath.Base(t.PkgPath) {
```
a8m, line 89: package filepath uses backslashes for Windows, use package path instead.

a8m:
Hey @htdvisser, 
Did you see my previous comment?

> Can you elaborate what is not working as expected?

htdvisser:
Yes, I saw it. I suspect the issue I experienced came from some tooling not understanding that `ulid.ULID` comes from an import ending in `/v2`. I'll dig into it some more when I have time.

a8m:
> I'll dig into it some more when I have time.

Sure, do you want to give me more content? what didn't work? maybe I can help with that.

htdvisser:
Turns out the problem was actually a missing import for the type of the ID. I added an additional commit to the pull request to fix that.

htdvisser:
I pushed a third commit to remove the alias I added initially. I'll leave it up to you to decide if you want an `import "github.com/oklog/ulid/v2"` (with all 3 commits in this PR) or `import ulid "github.com/oklog/ulid/v2"` (the first 2 commits).

a8m:
> Turns out the problem was actually a missing import for the type of the ID. I added an additional commit to the pull request to fix that.

Got it. Thanks for your contribution!

PR #531, opened 2020-06-08
entc/mutation: expose builder api for getting mutations

### LINKED ISSUE #529: Interface for Create/Update functions
Ent already generates similar functions for Create/Update functions. If you have to write a bit more complex code to handle create and update scenarios you'll end up writing a create and update function for ent, even though the logic may be similar. Consider the following example:

create a new one:

```
a8m, err := client.User.    // UserClient.
    Create().               // User create builder.
    SetName("a8m").         // Set field value.
    SetAge(30).    // Avoid nil checks.
    Save(ctx)               // Create and return.
```

handle existing entry:

```
a8m, err = a8m.Update().    // User update builder.
    SetName("a8m").         // Set field value.
    SetAge(31).             // Set field value
    Save(ctx)               // Save and return.
```

If ent would generate an interface for functions we could write the following:

```
entry, err := client.User.Create()
// gather existing
entry, err :=  a8m.Update()

entry.SetName("a8m"). 
SetAge(30). 
Save(ctx) 
```

At this point those interfaces can't be manually created since the return object for the methods is not an interface. Are there any other options that I missed?

a8m:
Fixed #529


chris-rock:
Super cool!!!!!! Thank you @a8m

a8m:
Hehe, I'm glad you liked it @chris-rock. Please see an example [here](https://github.com/facebookincubator/ent/blob/85612b308f957ae644683b370ba9dd7cc7ea85d2/entc/integration/integration_test.go#L1084).

I'm going to update the docs later this week, but for now, just note that you have a mutation object for each type (as attached in the link), but if you want to set a field/edge of different types, you can use the `ent.Mutation` object or to type-assert the specific method (e.g. [SetName](https://github.com/facebookincubator/ent/blob/85612b308f957ae644683b370ba9dd7cc7ea85d2/entc/integration/integration_test.go#L1071)).

Thanks for raising this issue, I actually wanted to add it a long time ago.

PR #533, opened 2020-06-09
Bugfix: eager lodaing when using uuid for foreign keys failing if loading more than a single relation (#532)

### LINKED ISSUE #532: Multiple foreign key UUID failing with `WithX()` eager loading
Use of UUID foreign keys is failing when using eager loading. Index of foreign keys explicitly set to 0 for all foreign keys.

https://github.com/facebookincubator/ent/blob/master/entc/gen/template/dialect/sql/decode.tmpl#L61-L63

```go
						{{- with extend $ "Idx" 0 "Field" $f "Rec" $receiver "StructField" $f.Name }}
							{{ template "dialect/sql/decode/field" . }}
						{{- end }}
```

Should be amended to:

```go
						{{- with extend $ "Idx" $i "Field" $f "Rec" $receiver "StructField" $f.Name }}
							{{ template "dialect/sql/decode/field" . }}
						{{- end }}
```

jufemaiz:
Resolves #532


a8m:
Thanks for your contribution @jufemaiz!

PR #545, opened 2020-06-11
schema/edge: an storage-key option for edges

### LINKED ISSUE #126: proposal: custom column name for edges
Currently there's a way to customize the Table name. Can we do the same for column names (including edges)? 

Thanks

a8m:
Fixed #126


a8m:
Docs will be updated tomorrow.

PR #547, opened 2020-06-13
Feature: introduce validation errors

jelmersnoeck:
This introduces the concept of validation errors, where we have a high
level validation error which wraps a more detailed error message.

The higher level `ValidationError` is set up in the generated files, much
like the `NotFoundError` and `ConstraintError` and is accompanied by an
`IsValidationError` check method. Thus, it can be used as follows:

```go
t, err := tx.Team.Create().SetName(input.Name).Save(ctx)
if ent.IsValidationError(err) {
        // handle validation error response
}
```


a8m:
Hey @jelmersnoeck and thanks for your contribution.

The general idea is good and looks fine, but I have some comments.

> It also introduces the notion of more detailed errors so that the
consumer of the API can use the more detailed error information to
determine what to do next.

I'm not sure I understand why it's needed (and afraid I'll need to maintain these error types). 
If it's not so common, users can always write their own validators, and return any error type they want. 

> The only backwards incompatible change is that the missing edge/field
error is now in the format of ent: validator failed for edge "<name>": missing required edge "<name>" instead of the simpler ent: missing required edge "<name>".

Can we avoid this?

aca:
Thanks @jelmersnoeck !
I have no right on this project. But I think it would be better to seperate pr.
One for introducing ```ValidationError``` types and one for ```core validator error types```.
And let maintainers decide whether to accept proposal or not for each feature.

jelmersnoeck:
Hey @a8m and @aca thanks for your feedback!

You're right, it's 2 concerns. I've split it out and this now only has the ValidationError type (and the check).

I've also opened a new issue describing the use case for more detailed errors with a revamped idea on how to approach it, which is more in line with what you suggested around using the `Validate()` method: https://github.com/facebookincubator/ent/issues/549

a8m:
@jelmersnoeck let me know when it's ready for a review (click "Request Review").
Thanks for your contribution.

jelmersnoeck:
Hey @a8m, I don't see another "Request Review" button (I think it's because it's not a draft anymore), but it's ready! I just ran the latest against our codebase and all seems good.

entc/gen/template/builder/create.tmpl
```diff
@@ -61,7 +61,7 @@ func ({{ $receiver }} *{{ $builder }}) Save(ctx context.Context) (*{{ $.Name }},
 			{{- else }}
 				if len({{ $mutation }}.{{ $e.StructField }}IDs()) == 0 {
 			{{- end }}
-				return nil, errors.New("{{ $pkg }}: missing required edge \"{{ $e.Name }}\"")
+				return nil, &ValidationError{Field: "{{ $e.Name }}", err: errors.New("{{ $pkg }}: missing required edge \"{{ $e.Name }}\"")}
```
a8m, line 64: Field?

schema/field/gen/numeric.tmpl
```diff
@@ -6,7 +6,7 @@
 package field
 
 import (
-	"errors"
+ 	"errors"
```
a8m, line 9: Is it needed?

entc/gen/template/base.tmpl
```diff
@@ -71,6 +71,31 @@ func As(fn AggregateFunc, end string) AggregateFunc {
 	}
 {{ end }}
 
+// ValidationError returns when validating a field fails.
+type ValidationError struct {
+    Field string
```
a8m, line 76: Do we need this? I'm fine with leaving this type just with `err`, since a `ValidationError` is returned for both edge and field validators.

jelmersnoeck: The `Unwrap` would allow to get the actual error message (coming from the underlying validate function), with a separate `Field`, it is now also possible to say exactly which field this error is for.

It is possible to get this out with a regexp (as the error message contains this), but seeing that the underlying error is accessible without regexp parsing (using Unwrap), it'd be great to be able to get the field out too without having to do that.

schema/field/gen/numeric.tmpl
jelmersnoeck: No! Sorry, didn't catch this when reviewing. Updated now.

entc/gen/template/builder/create.tmpl
jelmersnoeck: Apart from the discussion below wrgt the reasoning behind this field, would "Label" make more sense to not confuse it between fields/edges?

entc/gen/template/base.tmpl
a8m: I understand the use-case, but have problem with using edges as fields here.

What do you think about:

```go
type ValidationError struct {
    Field string // Field name; may be empty for edge errors.
    Edge string  // Edge name; may be empty for field errors.
    err error
}
```
Or:
```go
type ValidationError struct {
    Name string // Field or edge name.
    err error
}
```
WDYT? I prefer the second option.

jelmersnoeck: 👍  on the second one. I'll change!

jelmersnoeck: Updated and re-ran generation.

Issue #559, opened 2020-06-19
can not define ID as uint64


TargetLiu:
https://github.com/facebookincubator/ent/blob/93c78aeda6e736b08da5a00a3ee8167854540af7/entc/gen/type.go#L768-L775

I want to define ID as uint64 without auto increment in order to use distributed id generation such as snowflake etc.

But according to the above code I cannot do this.

What is the reason why the type cannot be changed？

a8m:
Hey @TargetLiu,
There's no real reason, I was created this way because we needed to support string/UUID IDs.
If we change it, we need to make sure we provide users an API to set the ID field to be incremented.

TargetLiu:
@a8m So do you have plans to do this?

a8m:
Hey @TargetLiu,
This issue can be addressed with additional field on the `entsql` annotation. See https://github.com/facebook/ent/pull/947 for a  reference. 

If you want to contribute this change, add another field `Increment *bool` to the `entsql.Annotation` and use it in [entc/type.go](https://github.com/facebook/ent/blob/master/entc/gen/type.go) as follows:

```go
func (f Field) PK() *schema.Column {
	c := &schema.Column{
		Name:      f.StorageKey(),
		Type:      f.Type.Type,
		Key:       schema.PrimaryKey,
		Increment: true,
	}
	if ant := f.EntSQL(); ant != nil && ant.Increment != nil {
		c.Increment = *ant.Increment
	}
	// ...
	return c
}
```

I suggested `*bool` for the field type, because `false` is also a valid value.

saantiaguilera:
Hi! Coming from #1073 which has a similar issue. Was there any follow up on this thread? I can make a PR including this behavior without trouble.

On a side note, how's the release process for the module? Do you do it on demand based on new features or there's a schedule?

Thanks!!

a8m:
> Hi! Coming from #1073 which has a similar issue. Was there any follow up on this thread? I can make a PR including this behavior without trouble.

Sure, let me know if you need some info/guidance and I can help with that (it will be tomorrow morning, because it's the EOD for me). 

> On a side note, how's the release process for the module? Do you do it on demand based on new features or there's a schedule?

No special process for this package, I'm releasing a new version when there are enough changes for a release or when someone asks for it in the issue-tracker or in [Slack](https://app.slack.com/client/T029RQSE6/C01FMSQDT53). But, users can always pull latest version using:
```
go get github.com/facebook/ent@master
```

[...]

Issue #562, opened 2020-06-21
Adding support for fields/edges dialect-specific DDLs


nilpntr:
Would you like to add support for ddl's to add default values for the time fields?

a8m:
Pasting here my comment from #558:

> I've landed 3e7481c3a015400fd5f407aba1c57f35731601dc that ignores the MySQL default values for time fields, and it should not apply any changes for the second time the migration runs.
> 
> However, it won't work with the current way you declare the schema, because you add additional parameters to the schema datatype (the create/update default values).
> 
> I suggest to close this issue and open a new one for adding support for fields/edges dialect-specific DDLs. For example:
> 
> ~~~go
> // Add abstraction for common DDLs.
> field.Time("create_time").
>     DDL(
>         field.Default("CURRENT_TIMESTAMP"),
>         field.OnUpdate("CURRENT_TIMESTAMP")
>     )
> 
> // Add additional information for the given column. The EXTRA in the
> // INFORMATION_SCHEMA table. For example, the "ON UPDATE" clause.
> field.Time("update_time").
>     SchemaDef(map[string]string{
>         MySQL: "DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP",
>     })
> 
> // Completely override the ent-migration column definition.   
> field.Time("create_time").
>     SchemaDef(map[string]string{
>         MySQL: "TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL",
>     })
> ~~~
> 

Feedback or thoughts are welcome.

sj8m:
That syntax would be nice :)

a8m:
Closing, as this was implemented by [entsql.Annotation](https://pkg.go.dev/entgo.io/ent/dialect/entsql). Thanks 🙏

PR #565, opened 2020-06-24
add: change id type and increment method for numeric (#559)

TargetLiu:


a8m:
Hey @TargetLiu and thanks for your contribution.

I was thinking about a different API for customize the column definition. Something as follows:

```go
// 1:
field.Int("id").
    SchemaOptions(field.Increment(false))

// 2:
field.Int("id").
    SchemaOptions(field.Increment(true), field.InitialValue(1000))

// 3:
field.Time("updated_at").
    SchemaOptions(field.Default("CURRENT_TIMESTAMP"), field.OnUpdate("CURRENT_TIMESTAMP"))
```

It will also solve #562 and can be extended with more options in the future.

TargetLiu:
@a8m What you think is more comprehensive than mine, looking forward to your completion.

PR #575, opened 2020-07-01
entc/gen: add support for tx hooks

a8m:


entc/gen/template/client.tmpl
```diff
@@ -79,6 +80,7 @@ func (c *Client) Tx(ctx context.Context) (*Tx, error) {
 	}
 	cfg := config{driver: tx, log: c.log, debug: c.debug, hooks: c.hooks}
 	return &Tx{
+		ctx: ctx,
```
alexsn, line 83: If we store a context in tx maybe we should add `Context() / WithContext()` so users can modify it

entc/gen/template/tx.tmpl
```diff
@@ -29,25 +29,67 @@ type Tx struct {
...
+		txDriver := tx.config.driver.(*txDriver)
+		var fn {{ $iface }} = {{ $func }}Func(func(context.Context, *Tx) error {
+			return txDriver.tx.{{ $func }}()
+		})
 		tx.mu.Lock()
-		defer tx.mu.Unlock()
-		for _, f := range tx.{{ $onFuncs }} {
-			f(err)
+		hooks := make([]{{ $func }}Hook, len(tx.{{ $onFuncs }}))
+		copy(hooks, tx.{{ $onFuncs }})
```
alexsn, line 83: hooks := append([]{{ $func }}Hook(nil), tx.{{ $onFuncs }}...)

entc/integration/integration_test.go
```diff
@@ -842,7 +860,13 @@ func Tx(t *testing.T, client *ent.Client) {
...
 		m.On("onRollback", nil).Once()
 		defer m.AssertExpectations(t)
-		tx.OnRollback(m.onRollback)
+		tx.OnRollback(func(next ent.Rollbacker) ent.Rollbacker {
+			return ent.RollbackFunc(func(ctx context.Context, tx *ent.Tx) error {
+				err := next.Rollback(ctx, tx)
+				m.onRollback(err)
+				return err
+			})
+		})
```
alexsn, line 869: This is a copy of the above, define once.

entc/gen/template/tx.tmpl
```diff
@@ -29,25 +29,67 @@ type Tx struct {
...
+	onCommit   []CommitHook
+	onRollback []RollbackHook
+
+	// ctx lives for the life of the transaction. It is
+	// the same context used by the underlying connection.
+	ctx context.Context
 }
 
-{{- range $func := list "Commit" "Rollback" }}
+{{- range $func, $iface := dict "Commit" "Committer" "Rollback" "Rollbacker" }}
```
alexsn, line 40: dict doesn't define order on keys so we might get generated changes without touching this template. Use `keys` for iteration which keep the keys sorted and get the value with `get`.

PR #579, opened 2020-07-02
schema/field: add the Unique option to the UUID field builder

moh-abk:
resolves - https://github.com/facebookincubator/ent/issues/578

### DIFF
--- a/schema/field/field_test.go
+++ b/schema/field/field_test.go
@@ -341,10 +341,11 @@ func TestField_Enums(t *testing.T) {
 }
 
 func TestField_UUID(t *testing.T) {
-	fd := field.UUID("id", uuid.UUID{}).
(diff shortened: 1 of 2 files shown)


schema/field/field_test.go
```diff
@@ -341,10 +341,11 @@ func TestField_Enums(t *testing.T) {
 }
 
 func TestField_UUID(t *testing.T) {
-	fd := field.UUID("id", uuid.UUID{}).
+	fd := field.UUID("id", uuid.UUID{}).Unique().
 		Default(uuid.New).
```
a8m, line 345:
```suggestion
	fd := field.UUID("id", uuid.UUID{}).
		Unique().
		Default(uuid.New).
```

PR #585, opened 2020-07-05
cmd/entc: add support for generate.go when default target argument is not applied

roger2hk:
Resolve https://github.com/facebookincubator/ent/issues/566.

### DIFF
--- a/cmd/entc/entc.go
+++ b/cmd/entc/entc.go
@@ -194,25 +194,25 @@ func createDir(target string) error {
 	if err := os.MkdirAll(target, os.ModePerm); err != nil {
 		return fmt.Errorf("creating schema directory: %w", err)
 	}
-	if err := ioutil.WriteFile(target+"/../generate.go", []byte(getGenerateFileData(target)), 0644); err != nil {
+	if err := ioutil.WriteFile(filepath.Join(target, "/../generate.go"), getGenerateFileData(target), 0644); err != nil {
 		return fmt.Errorf("creating generate.go file: %w", err)
 	}
 	return nil
 }
 
 // ent/generate.go file used for "go generate" command.
-func getGenerateFileData(target string) string {
+func getGenerateFileData(target string) []byte {
 	pathArr := strings.Split(target, "/")
 
 	var sb strings.Builder
 	sb.Grow(32)
 
 	sb.WriteString("package ent\n\n")
-	sb.WriteString("//go:generate go run github.com/facebookincubator/ent/cmd/entc generate ")
-	sb.WriteString("./" + pathArr[len(pathArr)-1])
+	sb.WriteString("//go:generate go run github.com/facebookincubator/ent/cmd/entc generate ./")
+	sb.WriteString(pathArr[len(pathArr)-1])
 	sb.WriteString("\n")
 
-	return sb.String()
+	return []byte(sb.String())
 }
 
 // schema template for the "init" command.
--- a/cmd/entc/entc.go
+++ b/cmd/entc/entc.go
@@ -202,14 +202,14 @@ func createDir(target string) error {
 
 // ent/generate.go file used for "go generate" command.
 func getGenerateFileData(target string) []byte {
-	pathArr := strings.Split(target, "/")
-
 	var sb strings.Builder
 	sb.Grow(32)
 
-	sb.WriteString("package ent\n\n")
+	sb.WriteString("package ")
+	sb.WriteString(filepath.Base(filepath.Dir(target)))
+	sb.WriteString("\n\n")
 	sb.WriteString("//go:generate go run github.com/facebookincubator/ent/cmd/entc generate ./")
-	sb.WriteString(pathArr[len(pathArr)-1])
+	sb.WriteString(filepath.Base(target))
 	sb.WriteString("\n")
 
 	return []byte(sb.String())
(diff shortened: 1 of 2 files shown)


a8m:
Thanks for your contribution @roger2hk 😃 
I'll review it later today.

cmd/entc/entc.go
```diff
@@ -194,15 +194,27 @@ func createDir(target string) error {
 	if err := os.MkdirAll(target, os.ModePerm); err != nil {
 		return fmt.Errorf("creating schema directory: %w", err)
 	}
-	if target != defaultSchema {
-		return nil
-	}
-	if err := ioutil.WriteFile("ent/generate.go", []byte(genFile), 0644); err != nil {
+	if err := ioutil.WriteFile(target+"/../generate.go", []byte(getGenerateFileData(target)), 0644); err != nil {
```
a8m, line 197: Use `filepath.Dir` and `filepath.Join` for assembling the file path (it won't work in Windows for example).

cmd/entc/entc.go
```diff
@@ -194,15 +194,27 @@ func createDir(target string) error {
...
+	var sb strings.Builder
+	sb.Grow(32)
+
+	sb.WriteString("package ent\n\n")
+	sb.WriteString("//go:generate go run github.com/facebookincubator/ent/cmd/entc generate ")
+	sb.WriteString("./" + pathArr[len(pathArr)-1])
+	sb.WriteString("\n")
+
+	return sb.String()
+}
```
a8m, line 216: Use `filepath` functions (as mentioned above) and return a `[]byte` because conversion above is redundant.

cmd/entc/entc_test.go
```diff
@@ -33,3 +33,27 @@ func TestCmd(t *testing.T) {
...
+	require.NoError(t, err)
+
+	cmd = exec.Command("go", "run", "github.com/facebookincubator/ent/cmd/entc", "generate", "./"+targetBasePath+"/"+targetSchemaFolderName)
+	stderr = bytes.NewBuffer(nil)
+	cmd.Stderr = stderr
+	require.NoError(t, cmd.Run(), stderr.String())
+
+	_, err = os.Stat(targetBasePath + "/user.go")
+	require.NoError(t, err)
+}
```
a8m, line 59: Please use shorter names for the local variables in this function.

cmd/entc/entc.go
roger2hk: Code updated in https://github.com/facebookincubator/ent/pull/585/commits/59dcd397fd020d5bbc111f6c26756f213a39dc78.

cmd/entc/entc.go
roger2hk: Code updated in https://github.com/facebookincubator/ent/pull/585/commits/59dcd397fd020d5bbc111f6c26756f213a39dc78.

I'm not sure if there is a better way to convert `strings.Builder` to a `[]byte`.

cmd/entc/entc_test.go
roger2hk: Code updated in https://github.com/facebookincubator/ent/pull/585/commits/59dcd397fd020d5bbc111f6c26756f213a39dc78.

roger2hk:
@a8m Thanks for the feedback. There might be flaws in the code as I started the Go journey a month ago. Hope the PR is good to Go now.

cmd/entc/entc.go
```diff
@@ -194,15 +194,27 @@ func createDir(target string) error {
 	if err := os.MkdirAll(target, os.ModePerm); err != nil {
 		return fmt.Errorf("creating schema directory: %w", err)
 	}
-	if target != defaultSchema {
-		return nil
-	}
-	if err := ioutil.WriteFile("ent/generate.go", []byte(genFile), 0644); err != nil {
+	if err := ioutil.WriteFile(filepath.Join(target, "/../generate.go"), getGenerateFileData(target), 0644); err != nil {
```
a8m, line 197: Use:
```go
filepath.Join(filepath.Dir(target), "generate.go")
```

cmd/entc/entc.go
```diff
@@ -194,15 +194,27 @@ func createDir(target string) error {
...
 }
 
+// ent/generate.go file used for "go generate" command.
+func getGenerateFileData(target string) []byte {
+	pathArr := strings.Split(target, "/")
+
+	var sb strings.Builder
+	sb.Grow(32)
+
+	sb.WriteString("package ent\n\n")
```
a8m, line 210: The generated package name is derived from the target directory (i.e. it's not `package ent`) - see the code in `entc/entc.go`. Maybe we need to export this logic.

roger2hk: See if 73ebac3 works as expected.

roger2hk:
> The reason I didn't add the `generate.go` file to cases that use `--target`, is because that sometimes users structure their packages differently (search in previous issues). For example:
> 
> ```
> .
> ├── schema
> │   ├── user.go
> │   ├── ....
> │   └── group.go
> ├── entity
> │   ├── user_query.go
> │   ├── ....
> │   └── user_create.go
> └── project-root.md
> ```
> 
> I think that if we want to add support for this, we should address ~all custom cases.

If we want to add support for this, we will need two command parameters for both schema path and generate target path.

roger2hk:
@a8m Would adding two command parameters for both schema path and generate target path help? Any missing scenario?

roger2hk:
This PR is now closed as it is already behind a lot of commits. https://github.com/facebook/ent/pull/827 is also a PR for the same issue.

PR #587, opened 2020-07-05
dialect/sql/schema: accept convert from string2enum and enum2stirng

a8m:


dialect/sql/schema/schema.go
```diff
@@ -171,6 +171,9 @@ func (c *Column) ConvertibleTo(d *Column) bool {
 	case c.UintType() && d.IntType():
 		// uintX can not be converted to intY, when X > Y.
 		return c.Type-field.TypeUint8 <= d.Type-field.TypeInt8
+	case c.Type == field.TypeString && d.Type == field.TypeEnum ||
+		c.Type == field.TypeEnum && d.Type == field.TypeString:
```
alexsn, line 175: Lets stay consistent and add helpers as we do for numeric columns:
```
// StringType reports of the given type is a string type.
func (c Column) StringType() bool { return c.Type == field.TypeString }

// EnumType reports of the given type is an enum type.
func (c Column) EnumType() bool { return c.Type == field.TypeEnum }
```

PR #631, opened 2020-07-21
Split strings in pascal/camcelCase conversion on - in addition to _

errorhandler:
I've been evaluating Ent for production use and ran into a small issue with the code gen and enums. If an enum string value is `kebab-case` the generated go code isn't valid. For example:

```go
...
		field.Enum("role").
			Values("user", "admin", "free-user").
			Default("user")
...
```

Generates:

```go
...
// Role values.
const (
		RoleUser Role = "user"
		RoleAdmin Role = "admin"
		RoleFree-user Role = "free-user"
)
...
```

The issue is `RoleFree-user` isn't a valid Go variable name.

I've enhanced the Pascal/camelCase string conversion functions to split on `-` as well as `_`. The resulting code gen is more expected:

```go
// Role values.
const (
	RoleUser     Role = "user"
	RoleAdmin    Role = "admin"
	RoleUser     Role = "user"
	RoleAdmin    Role = "admin"
	RoleFreeUser Role = "free-user"
)
```

### DIFF
--- a/entc/gen/func.go
+++ b/entc/gen/func.go
@@ -108,14 +108,11 @@ func plural(name string) string {
 	return p
 }
 
-// pascal converts the given column name into a PascalCase.
-//
-//	user_info => UserInfo
-//	full_name => FullName
-//	user_id   => UserID
-//
-func pascal(s string) string {
-	words := strings.Split(s, "_")
+func isSeparator(r rune) bool {
+	return r == '_' || r == '-'
+}
+
+func pascalWords(words []string) string {
 	for i, w := range words {
 		upper := strings.ToUpper(w)
 		if _, ok := acronyms[upper]; ok {
@@ -127,18 +124,31 @@ func pascal(s string) string {
 	return strings.Join(words, "")
 }
 
-// camel converts the given column name into a camelCase.
+// pascal converts the given name into a PascalCase.
 //
-//	user_info => userInfo
-//	full_name => fullName
-//	user_id   => userID
+//	user_info => UserInfo
+//	full_name => FullName
+//	user_id   => UserID
+//	full-admin   => FullAdmin
+//
+func pascal(s string) string {
+	words := strings.FieldsFunc(s, isSeparator)
+	return pascalWords(words)
+}
+
+// camel converts the given name into a camelCase.
+//
+//	user_info  => userInfo
+//	full_name  => fullName
+//	user_id    => userID
+//	full-admin => fullAdmin
 //
 func camel(s string) string {
-	words := strings.SplitN(s, "_", 2)
+	words := strings.FieldsFunc(s, isSeparator)
 	if len(words) == 1 {
 		return strings.ToLower(words[0])
 	}
-	return strings.ToLower(words[0]) + pascal(words[1])
+	return strings.ToLower(words[0]) + pascalWords(words[1:])
 }
 
 // snake converts the given struct or field name into a snake_case.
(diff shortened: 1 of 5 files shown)


entc/gen/func.go
```diff
@@ -127,18 +124,31 @@ func pascal(s string) string {
...
-//	full_name => fullName
-//	user_id   => userID
+//	user_info  => userInfo
+//	full_name  => fullName
+//	user_id    => userID
+//	full-admin => fullAdmin
 //
 func camel(s string) string {
-	words := strings.SplitN(s, "_", 2)
+	words := strings.FieldsFunc(s, isSeparatorRune)
```
errorhandler, line 147: Sadly, `strings.FieldsFunc` doesn't have a `SplitN` equivalent so I've split out the `pascalWords` function which assumes the string is already split.

entc/gen/func.go
```diff
@@ -108,14 +108,11 @@ func plural(name string) string {
...
-// pascal converts the given column name into a PascalCase.
-//
-//	user_info => UserInfo
-//	full_name => FullName
-//	user_id   => UserID
-//
-func pascal(s string) string {
-	words := strings.Split(s, "_")
+func isSeparatorRune(r rune) bool {
+	return r == '_' || r == '-'
```
errorhandler, line 112: This is the key change here - I've only added `-`, if anyone can think of other separators that make sense I'm happy to add them here.

entc/gen/func.go
```diff
@@ -127,18 +124,31 @@ func pascal(s string) string {
 	return strings.Join(words, "")
 }
 
-// camel converts the given column name into a camelCase.
+// pascal converts the given name into a PascalCase.
```
errorhandler, line 127: I've updated this comment to reflect (as per my limited understanding) that this function is also used for more than column name conversion, such as enum values.

a8m:
Hey @errorhandler and thanks for your contribution.

> I've been evaluating Ent for production use and ran into a small issue with the code gen and enums.

It' great to hear this and you can feel free to ask here questions if you need any help.


About your use case, we had ~similar issue with GraphQL enums - GraphQL requires uppercased enums and that caused our constants to be not "Go-ish". Therefore, we've just added an option to control the Go name for enums. See #635.

The API looks as follows: 
```go
field.Enum("type").
	// A map from Go name to the actual enum value.
	ValueMap(map[string]string{
		"PNG": "png",
		"SVG": "svg",
		"JPG": "jpg",
	}).
```

Does it solve your problem?
If you still think some of the code in this PR should be added to ent, let me know and I'll review it.

errorhandler:
Hey @a8m! That will fix the issue I've encountered, however, personally I think it would be best to also avoid generating invalid Go code as much as possible. Perfectly happy to just use `ValueMap` but I also think there would be value in handling kebab-case out of the box.

a8m:
> Hey @a8m! That will fix the issue I've encountered, however, personally I think it would be best to also avoid generating invalid Go code as much as possible. Perfectly happy to just use `ValueMap` but I also think there would be value in handling kebab-case out of the box.

I don't mind adding this. If you want to rebase your branch and resolve the conflicts, I'll review it. If not, that's OK, just let me know and I'll add it a different PR. 

Thanks for your contribution!

errorhandler:
@a8m Thanks! I've fixed up the conflicts!

entc/gen/func.go
```diff
@@ -108,14 +108,11 @@ func plural(name string) string {
...
 
-// pascal converts the given column name into a PascalCase.
-//
-//	user_info => UserInfo
-//	full_name => FullName
-//	user_id   => UserID
-//
-func pascal(s string) string {
-	words := strings.Split(s, "_")
+func isSeparatorRune(r rune) bool {
```
a8m, line 111: Don't need to add the type to the func name. Let's remove it.

```suggestion
func isSeparator(r rune) bool {
```

errorhandler: Great catch - updated

a8m:
Thanks for your contribution @errorhandler

:shipit:

Issue #127, opened 2019-10-25
override id column


marwan-at-work:
Is there a way to redeclare the id column as a custom string type? 

If I put the following in my schema: 

`field.String("id").Unique()`

The generated code ends up broken because it creates two declarations of "id", but furthermore it ignores the fact that it's a string, and declares it as an integer instead. 

This might be related to the UUID discussions, but it would be great if we can define our own custom primary keys as strings that we provide at runtime.

alexsn:
Currently the "id" field is implied so there's no way to override the underlying storage key. The above setting should give a better error message where a reserved field name is used.

alexsn:
Are you looking to rename the "id" column name, provide your own PKs or both?

a8m:
This is part of our roadmap and it should be addressed in the upcoming week (or 2).
I linked this issue to the roadmap item.

marwan-at-work:
@alexsn im not looking to rename "id". I'm looking to keep it "id" but change its type from integer to string, and also be able to provide my own value when i create a row. 

Thanks :v:

BitPhinix:
Seems to work with the lastest branch, but there isn't a uuid type available. Is there any way to use a Postgres UUID primary key with ent?

a8m:
Hey @BitPhinix, this feature is still in development. 
I'll update the issue (and the roadmap) once we finish the implementation.

aca:
@a8m Feature Proposal 
I think it is related to this, so I write it here.
Can we customize struct tag for field "id" ?

I'm currently using type generated by ent, and I want to make json reponse with no ID field.
Which is achievable by Sensitive() for other fields.
But It seems it is impossible to make edit struct tag for ID field.

```
type Project struct {
    ent.Project
    ID `json:"-"`
}
```
I thought I could solve this by doing something like this. But didn't worked well.

a8m:
Hey @aca, did you try to override the struct tag? See - https://entgo.io/docs/schema-fields/#struct-tags

aca:
@a8m I meant ID field which is generated by entc not by defining ent.Field
```
// Fields of the Namespace.
func (Namespace) Fields() []ent.Field {
	return []ent.Field{
		field.String("name").Unique().Match(regexp.MustCompile("[a-z0-9]+(?:[._-][a-z0-9]+)*")).NotEmpty(),
		field.Int64("id").StructTag("-").Unique(),
	}
}
```
[...]

PR #634, opened 2020-07-22
use strconv.FormatInt for string id

stevenferrer:
Use `strconv.FormatInt` instead of type casting string ID


a8m:
Hey @stevenferrer and thanks for your contribution. 

Can you give more context for this PR? I assume it's related to previous functionality mentioned here - https://github.com/facebookincubator/ent/issues/127#issuecomment-583834177

stevenferrer:
Hi @a8m, likewise, thank you very much for this amazing project!

I just want to address a small issue when using `--idtype string` specifically a line generated in [`assignValues`](https://github.com/stevenferrer/ent-demo/blob/master/ent/user.go#L40) and [`[Entity]Create.sqlSave`](https://github.com/stevenferrer/ent-demo/blob/master/ent/user_create.go#L83).

The difference demonstrated below:

Current: `fmt.Printf("%#v", string(1))` produces `"\x01"`
Proposed: `fmt.Printf("%#v", strconv.FormatInt(1, 10)` produces `"1"`

I'm not sure though if #127 is related.

a8m:
Can you elaborate how do you use `string` id-type with SQL? IDs are stored as int in the database? If yes, what's the reason for this?

As mentioned in the linked comment:

> This (legacy) option was added to allow using the same generate ent package for both AWS Neptune (gremlin) and SQL dialects. Neptune uses UUID strings for its IDs, and MySQL is configured with numeric types for its PKs (in our deployment). Therefore, **the type in the code must be string, and we convert it to int before/after the data were loaded.**

stevenferrer:
An example is when you use `json.Marshal`:

```go
brown := client.User.Create().SetName("brown").SaveX(ctx)
b, _ := json.Marshal(brown)
fmt.Println(string(b))
```
Outputs:
```json
{"id":"\u0001","name":"brown"}
```

a8m:
Thanks for sharing the example @stevenferrer. Can you please answer my previous comment to help me get more context?

> Can you elaborate how do you use string id-type with SQL? IDs are stored as int in the database? If yes, what's the reason for this?

stevenferrer:
We're using `PostgreSQL` and before migrating some other parts of our system to ent, we're already using `bigserial` as our ID type and we're mapping it as `string` in our Go structs. 

For example, the below table:

```sql
CREATE TABLE address (
    id bigserial NOT NULL,
    unit_no varchar NOT NULL,
    barangay varchar NOT NULL,
    province varchar NOT NULL,
    city varchar NOT NULL,
    updated_at timestamp,
    created_at timestamp NOT NULL DEFAULT now()
);
```

is mapped to this struct:

```go
type Address struct {
	ID string `json:"id"`
	UnitNo string `json:"unitNo"`
	Barangay string `json:"barangay"`
	Province string `json:"province"`
	City string `json:"city"`
	UpdatedAt *time.Time `json:"-"`
	CreatedAt *time.Time `json:"createdAt"`
}
```

a8m:
> We're using PostgreSQL and before migrating some other parts of our system to ent, we're already using bigserial as our ID type and we're mapping it as string in our Go structs.

Thanks for sharing this information. 

There's no way to auto-map numeric value to string type in ent. Your PR fixes one problem with this, but there are so many other places that need to be changed in order to support this.

The [GoType](https://entgo.io/docs/schema-fields/#go-type) option is used exactly for cases like this, but it doesn't support ID fields as the moment. Once I'll add the support for this, you will be able to use it as follows:

```go
type StringID string // Need to implement the schema/field.ValueScanner interface.

	return []ent.Field{
		field.Int("id").
			GoType(new(StringID)),
	}
}
```

If you don't want to be blocked on this, you can use [external-templates](https://entgo.io/docs/templates/) for now (to override the existing behavior) and update your code in the future.

stevenferrer:
Ok, got it. Will probably use a workaround for now, thanks for the suggestion! Closing this PR then.

a8m:
I'll update this PR with the new API soon @stevenferrer.
Thanks for your contribution :)

PR #660, opened 2020-08-06
all: facebookincubator/ent => facebook/ent

a8m:
ent repository is going to be migrated to facebook organization


a8m:
Command I used in MacOS to find+replace all "facebookincubator" imports to "facebook". 
```
find . -type f -name '*.go' -exec sed -i '' 's/facebookincubator/facebook/g' {} +
```

Issue #702, opened 2020-08-27
ent/examples/edgeindex/edgeindex.go error


peanut-cc:
in `ent/examples/edgeindex/edgeindex.go ` example

In the following code：
```
	_, err := client.Street.
		Create().
		SetName("ST").
		SetCity(tlv).
		Save(ctx)
	if err == nil {
		return fmt.Errorf("expecting creation to fail")
	}
```

Should be as follows：
```
	_, err := client.Street.
		Create().
		SetName("ST").
		SetCity(tlv).
		Save(ctx)
	if err != nil {
		return fmt.Errorf("expecting creation to fail")
	}
```
？？？

a8m:
Hey @peanut-cc!

No, that's fine. See the comment above the code block:
![image](https://user-images.githubusercontent.com/7413593/91476104-2ff42a00-e8a5-11ea-981c-a1382f365bf0.png)

peanut-cc:
ok

a8m:
> ok

Did you understand why it's OK? as you can see, in the first block [here](https://github.com/facebook/ent/blob/master/examples/edgeindex/edgeindex.go#L43-L48), we're adding a street named "ST" to the city TLV, and since the street name must be unique under the city edge, the [next block](https://github.com/facebook/ent/blob/master/examples/edgeindex/edgeindex.go#L51-L55) should fail because they are duplicates.

PR #703, opened 2020-08-27
fix(examples/edgeindex/edgeindex.go): edgeindex.go about error judge bug

peanut-cc:
fix https://github.com/facebook/ent/issues/702


a8m:
Thanks for your contribution @peanut-cc, but that's an expected behavior. See my comment in #702 😃

PR #777, opened 2020-09-17
/examples/traversal Invoke deferred call client.Close() before log.Fatalf exits immediately.

ProximaB:
os.Exit() exits immediately and no deferred statements can be evaluated.

```
// Fatalf is equivalent to Printf() followed by a call to os.Exit(1).
func Fatalf(format string, v ...interface{}) {
	std.Output(2, fmt.Sprintf(format, v...))
	os.Exit(1)
}
```

### DIFF
--- a/examples/traversal/example_test.go
+++ b/examples/traversal/example_test.go
@@ -18,39 +18,47 @@ import (
 	"github.com/pkg/errors"
 )
 
-func Example_Traversal() {
+func Example_traversal() {
 	client, err := ent.Open("sqlite3", "file:ent?mode=memory&cache=shared&_fk=1")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


a8m:
Hey @ProximaB and thanks for your contribution.

> os.Exit() exits immediately and no deferred statements can be evaluated.

You are right, and I agree that this is "more correct".  However, IMO it makes the example-program less friendly to read, and in terms of correctness - the in-memory database will be closed anyway if the program exits, so there's no really need for closing it. Calling `client.Close()` is necessary when there are several test-cases that open the same database.

ProximaB:
Yeah, I see your point. Anyway If you care about closing resources why would you use Fatalf in first place.

Zenix27 (approved):

PR #814, opened 2020-10-03
doc: add rune length validator example

roger2hk:
Add rune length validator example in doc for https://github.com/facebook/ent/issues/784.

### DIFF
--- a/doc/md/schema-fields.md
+++ b/doc/md/schema-fields.md
@@ -275,6 +275,24 @@ func (Group) Fields() []ent.Field {
 }
 ```
 
+Here is another example for validating the rune length by using the [unicode/utf8](https://golang.org/pkg/unicode/utf8/) package.
+
+```go
+func MaxRuneCount(maxLen int) func(s string) error {
+	return func(s string) error {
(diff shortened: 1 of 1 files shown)


doc/md/schema-fields.md
```diff
@@ -275,6 +275,24 @@ func (Group) Fields() []ent.Field {
...
+		}
+		return nil
+	}
+}
+
+field.String("name").
+	Validate(MaxRuneCount(10))
+field.String("nickname").
+	Validate(MaxRuneCount(20))
+```
```
a8m, line 294:
```suggestion
Here is another example for writing a reusable validator:

```go
// MaxRuneCount validates the rune length of a string by using the unicode/utf8 package.
func MaxRuneCount(maxLen int) func(s string) error {
	return func(s string) error {
		if utf8.RuneCountInString(s) > maxLen {
			return errors.New("value is more than the max length")
		}
		return nil
	}
}

field.String("name").
	Validate(MaxRuneCount(10))
field.String("nickname").
	Validate(MaxRuneCount(20))
```
```

a8m (approved):
Minor doc change. Please address and I'll merge it.

PR #836, opened 2020-10-11
ent/privacy: initial privacy package

a8m:


entc/gen/template/privacy/privacy.tmpl
```diff
@@ -35,52 +39,22 @@ var (
...
-	decision, ok := ctx.Value(decisionCtxKey{}).(error)
-	if ok && errors.Is(decision, Allow) {
-		decision = nil
-	}
-	return decision, ok
-}
+var (
+	// DecisionContext creates a new context from the given parent context with
+	// a policy decision attach to it.
+	DecisionContext = privacy.DecisionContext
```
alexsn, line 45: Since go doesn't support `const` we shouldn't use `var` but a forward function instead.

entc/gen/template/privacy/privacy.tmpl
```diff
@@ -35,52 +39,22 @@ var (
...
+	// DecisionFromContext retrieves the policy decision from the context.
+	DecisionFromContext = privacy.DecisionFromContext
```
alexsn, line 47: Ditto

alexsn (changes requested):

PR #861, opened 2020-10-19
`notIn` support Selector like `in`

maxiloEmmmm:


dialect/sql/builder.go
```diff
@@ -1057,7 +1057,11 @@ func (p *Predicate) NotIn(col string, args ...interface{}) *Predicate {
 	return p.Append(func(b *Builder) {
 		b.Ident(col).WriteOp(OpNotIn)
 		b.Nested(func(b *Builder) {
-			b.Args(args...)
```
a8m, line 1060: Please add the same check with do in line 1007.

a8m (changes requested):

a8m (approved):

a8m:
Thanks for your contribution @maxiloEmmmm 😃

PR #874, opened 2020-10-21
Fix naming conflict for 'Client'

### LINKED ISSUE #873: Bug: Creating an entity named `Client` doesn't go well
### What happened
I created a new entity named `Client` and ran `go generate ./ent` successfully, but the resulting code had errors, so the application won't build. On some investigation I found that we are using a struct named `Client` in the generated code already which was getting mixed with the `Client` entity.

### How to reproduce
- Create an entity named `Client`, `entc init Client`
- Generate code, `go generate ./ent`
- Now try building the code

### Expected behavior
`entc` should generate code which is works in this case. If this is not feasible then we can atleast throw a warning during `entc generate` or `entc init` if the entity name equals one of the "restricted" names.

prmsrswt:
Added `'Client'` in the `globalIdent` list. Fixes #873

### DIFF
--- a/entc/gen/type.go
+++ b/entc/gen/type.go
@@ -1254,6 +1254,7 @@ var (
 		"AggregateFunc",
 		"As",
 		"Asc",
+		"Client",
 		"Count",
 		"Debug",
 		"Desc",


a8m:
Thanks for the fix @prmsrswt!

PR #882, opened 2020-10-25
doc: replace CreateX reference to SaveX

ernado:
Looks like `CreateX` was deprecated and this comment is slightly outdated.


a8m (approved):

a8m:
Thanks for your contribution @ernado

PR #893, opened 2020-10-28
feat: Postgres Native Enums

cliedeman:
Resolves https://github.com/facebook/ent/issues/872

```go
func (Person) Fields() []ent.Field {
	return []ent.Field{
		field.Int("id"),
		field.Enum("mood").GoType(mood.Mood("mood")).SchemaType(map[string]string{dialect.Postgres: "mood"}),
	}
}
```

Notes
- alter type is not used because it cannot run in a transaction and cannot remove types
- the current implementation will drop enum types that ent does not know about
- migration to and from enums are supported but this assumes that in the varchar -> enum case all the values are valid enum values


dialect/sql/schema/migrate.go
```diff
@@ -603,6 +647,13 @@ type sqlDialect interface {
 	alterColumns(table string, add, modify, drop []*Column) sql.Queries
 }
 
+type nativeEnumer interface {
```
cliedeman, line 650: could not think of a better name....

dialect/sql/schema/postgres_test.go
```diff
@@ -45,6 +45,8 @@ func TestPostgres_Create(t *testing.T) {
 			name: "no tables",
```
cliedeman, line 45: More tests need to be added for enum use cases

dialect/sql/schema/postgres.go
```diff
@@ -332,15 +347,241 @@ func (d *Postgres) addColumn(c *Column) *sql.ColumnBuilder {
...
+     pg_enum ON pg_enum.enumtypid = pg_type.oid
+         INNER JOIN pg_namespace on pg_namespace.oid = pg_type.typnamespace
+WHERE pg_namespace.nspname = current_schema()
+ORDER BY pg_type.typname, pg_enum.enumlabel`
+
+func (d *Postgres) enumChanges(ctx context.Context, tx dialect.Tx, tables ...*Table) (*enumChanges, error) {
+	expectedEnums := make(map[string]*NativeEnum)
+
+	for _, t := range tables {
+		// How do we retrieve annotation here?
```
cliedeman, line 380: We could alter behaviour using annotations but they do not seem to be copied from field to column

cliedeman:
Not sure why generate is failing

a8m:
Hey @cliedeman and thanks for adding this! It's really amazing.

I'll start giving it a look tomorrow, but I can say now, that I prefer to merge it in small chunks (split it into multiple PRs) - I can help you with that tomorrow. 

BTW, it is ready for review, or you still work on it?

cliedeman:
Hi @a8m It is ready for review. I only expect to add code. More unit tests and some validation of the enums (E.g. a user should not be allowed to specify the same enum on 2 different tables with differing values).

Happy to split it into chunks as you see fit

dialect/sql/schema/postgres.go
```diff
@@ -332,15 +347,241 @@ func (d *Postgres) addColumn(c *Column) *sql.ColumnBuilder {
...
+	for typeName := range enumsInUse {
+		var result sql.Result
+		err := tx.Exec(ctx, fmt.Sprintf("ALTER TYPE %s RENAME TO %s", typeName, typeTempName(typeName)), []interface{}{}, result)
+
+		if err != nil {
+			return err
+		}
+	}
+
+	// Step 2. Create new Enums
```
ernado, line 533: Can we use "new enum" strategy only as a fallback?

In situation where only new values are added to already existing enum, we can use just `ALTER TYPE ADD VALUE` .

cliedeman: alter type add value cannot run in a transaction. So if the migration failed we will leave the database in a 'dirty' state

ernado: Seems like it was changed recently?

> If ALTER TYPE ... ADD VALUE (the form that adds a new value to an enum type) is executed inside a transaction block, the new value cannot be used until after the transaction has been committed.

https://www.postgresql.org/docs/current/sql-altertype.html

Probably we can leave it as is and introduce opt-in optimization for that.

cliedeman: @ernado thats great information. Looks like its been added since 12.

cliedeman: I will be happy to add that in a followup pr. this one is already quite big

cliedeman:
@a8m I have converted this pr to draft while I split this out into smaller pr's.

I should get to it in the next few days

mvrahden:
Hi @cliedeman was there any update in the meantime on this topic?

juicycleff:
Really need this feature, worked around it by writing custom migrations just for enums, but now Atlas complains with the below error `ariga.io/atlas/sql/postgres.(*state).mayDropEnum` because I guess it does not know how to drop the enum

mvrahden:
> Really need this feature, worked around it by writing custom migrations just for enums, but now Atlas complains with the below error `ariga.io/atlas/sql/postgres.(*state).mayDropEnum` because I guess it does not know how to drop the enum

Try out https://github.com/mvrahden/go-enumer to generate enums. It has a compatibility option with `ent`

PR #894, opened 2020-10-28
To snake case file name - when we use **entc init CamelCaseEntity**

haoxins:


a8m:
Hey @haoxins and thanks for your contribution.

I actually prefer to stick with the same format that `entc` uses for codegen (lowercased names). I also don't like the idea that `init` and `gen` will use 2 different formats (`schema/user_group.go` and `ent/usergroup.go`).

haoxins:
make sense :)

haoxins:
@a8m a quick question, how about change the `gen template` filenames to snake case also?

`user_group_delete.go`

PR #904, opened 2020-10-30
entc/gen: fix typo in "MaskNotFound" func comment

ernado:
Found another typo ^^

Looks like after `go generate ./...` the `bindata.go` changed from ` os.FileMode(420)` to `os.FileMode(436)` which is weird, should I fix it?


a8m:
Hey @ernado and thanks for fixing this.

Please re-run `go generate ./...` from the root of the project. Regarding binddata, I'm not sure what's the problem, but if you find some, feel free to send a patch.

ernado:
```console
ernado@nexus:/src/ent$ go generate ./...
ernado@nexus:/src/ent$ git diff
```
I'm getting zero diff after running `go geneate ./...` in root and `git status` is not reporting any changed files :thinking:

ernado:
Probably os.FileMode(420) is causing that issue, I'll investigate and ping you when pipeline is green again.

ernado:
PTAL @a8m, I've fixed pipeline issues.

Note that I've updated bindata flags so this won't happen again for anyone.

a8m:
Thanks for your contribution @ernado 😃

PR #921, opened 2020-11-05
Fix 917

masseelch:
Currently model/stringer is defined as TypeTemplate so the String() method gets attached to the same file it was in before.

I can make a GraphTemplate out of it. What do you prefer?


entc/gen/template.go
```diff
@@ -61,6 +61,9 @@ var (
 		{
 			Name:   "model",
 			Format: pkgf("%s.go"),
+			ExtendPatterns: []string{
+				"model/stringer",
+			},
```
a8m, line 66: You can delete this as it's not a pattern.

a8m (approved):
Looks good except one minor comment. Please fix and I'll merge it.

a8m:
Wonderful @masseelch! Thanks for your contribution.

PR #926, opened 2020-11-07
fix: pk forces the Type of the numeric to be TypeInt so that the user defined id type is invalid

TargetLiu:
fix: pk forces the Type of the numeric to be TypeInt so that the user defined id type is invalid


a8m (approved):
Thanks for the fix @TargetLiu

PR #938, opened 2020-11-10
feat: Added Annotation Options

### LINKED ISSUE #924: Set ROCKSDB Engine for MySQL Table Migration
Hi, I am trying to use this tool to create a database in MySQL. But, I want to use ROCKSDB as engine. I couldn't find any setting to set the engine when running migration.

GORM Does it like this - db.Set("gorm:table_options", "ENGINE=InnoDB").AutoMigrate(&User{})
https://gorm.io/docs/migration.html#content-inner

dooven:
resolves #924 

Added `Option` field in the `Annotations` struct so that users can specify additional options like `ENGINE = INNODB` when creating a table

### DIFF
--- a/dialect/entsql/annotation.go
+++ b/dialect/entsql/annotation.go
@@ -34,6 +34,14 @@ type Annotation struct {
 	//	}
 	//
 	Collation string `json:"collation,omitempty"`
+
+	// Options defines the additional table options. For example:
+	//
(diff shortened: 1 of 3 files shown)


a8m (approved):

a8m:
Looks good @dooven! Thanks for adding this.

PR #949, opened 2020-11-14
fix invalid type, add an postgre type 'timestamp without time zone' to the scan colum

ssoor:
When the table already exists and has column the type is "timestamp without time zone", using migrate will return an error.

```
2020/11/14 05:09:34 failed printing schema changes: sql/schema: invalid type "timestamp without time zone" for column "create_time"
```
This is because when scanning column types, the corresponding ones are not processed.

### DIFF
--- a/dialect/sql/schema/postgres.go
+++ b/dialect/sql/schema/postgres.go
@@ -241,7 +241,7 @@ func (d *Postgres) scanColumn(c *Column, rows *sql.Rows) error {
 		c.Size = maxCharSize + 1
 	case "character", "character varying":
 		c.Type = field.TypeString
-	case "date", "time", "timestamp", "timestamp with time zone":
+	case "date", "time", "timestamp", "timestamp with time zone", "timestamp without time zone":
 		c.Type = field.TypeTime
 	case "bytea":
 		c.Type = field.TypeBytes


dialect/sql/schema/postgres.go
```diff
@@ -241,7 +241,7 @@ func (d *Postgres) scanColumn(c *Column, rows *sql.Rows) error {
 		c.Size = maxCharSize + 1
 	case "character", "character varying":
 		c.Type = field.TypeString
-	case "date", "time", "timestamp", "timestamp with time zone":
+	case "date", "time", "timestamp", "timestamp with time zone", "timestamp without time zone":
```
cliedeman, line 244: should timestamptz variations also be included?

a8m: I'm not sure. We need to check what is returned from `INFORMATION_SCHEMA` if the column was created with the `timestamptz` abbreviation.

a8m (approved):
Thanks for your contribution @ssoor

dialect/sql/schema/postgres.go
a8m: @ssoor, you're welcome to check that and contribute this in a different PR if it's needed.

PR #954, opened 2020-11-16
Feat/integer to string part2

cliedeman:
Added conversions that account for string max length.
Had to make some compromises on the tests because postgres does not support unsigned and uses smallint for both int8 and int16


dialect/sql/schema/schema.go
```diff
@@ -186,13 +187,45 @@ func (c *Column) ConvertibleTo(d *Column) bool {
...
 		return true
-	// Allow conversions to arbitrary length strings
-	case c.Type.Integer() && d.Type == field.TypeString && d.Size == 0:
+	case c.Type.Integer() && d.Type == field.TypeString && NumericConvertibleToString(c.Type, d.Size):
 		return true
 	}
 	return c.FloatType() && d.FloatType()
 }
 
+func NumericConvertibleToString(numericType field.Type, stringSize int64) bool {
```
a8m, line 196: Do we need this? What happens if we not check this, does the database reject the operation or it truncate the data?

entc/integration/migrate/entv2/schema/conversion.go
```diff
@@ -0,0 +1,32 @@
...
+		// Postgres uses the same type for int8 and int16
+		// Postgres loses unsigned so we have assume value is signed
+		field.String("int8_to_string").MaxLen(6),
+		field.String("uint8_to_string").MaxLen(6),
+		field.String("int16_to_string").MaxLen(6),
+		field.String("uint16_to_string").MaxLen(6),
+		field.String("int32_to_string").MaxLen(12),
+		field.String("uint32_to_string").MaxLen(12),
+		field.String("int64_to_string").MaxLen(21),
+		field.String("uint64_to_string").MaxLen(21),
```
a8m, line 30: Use `entsql.Annotation{Size: X}` instead.

dialect/sql/schema/schema.go
cliedeman: Hi @a8m both postgres and myql fail with value too long errors. I should have tested this first.

We can remove these checks then.

```
-- Mysql
create table conversion (
    col1 smallint not null
);

insert into conversion values (0), (1), (32767);

alter table conversion modify column col1 varchar(1);
[2020-11-19 07:22:39] [22001][1406] Data truncation: Data too long for column 'col1' at row 3
[2020-11-19 07:22:39] [22001][1406] Data too long for column 'col1' at row 3
```

```
-- Postgres
create table conversion (
    col1 smallint not null
);

insert into conversion values (0), (1), (32767);

alter table conversion alter column col1 set data type varchar(1);
[2020-11-19 07:25:30] [22001] ERROR: value too long for type character varying(1)
```

entc/integration/migrate/entv2/schema/conversion.go
cliedeman: updated

dialect/sql/schema/schema.go
a8m: Let's remove it then. 

We'll do it the same way we migrate non-unique columns to unique - if the database accepts this (no duplicates values), we fine with this. The same way here, if the table is empty, or doesn't contain invalid values the change should be accepted.

cliedeman: Done 👍

a8m:
Thanks @cliedeman for your contribution!

PR #960, opened 2020-11-18
Add DefaultFunc for string field

haoxins:
https://github.com/facebook/ent/issues/935

### DIFF
--- a/schema/field/field.go
+++ b/schema/field/field.go
@@ -208,6 +208,12 @@ func (b *stringBuilder) Default(s string) *stringBuilder {
 	return b
 }
 
+// DefaultFunc sets the default value generating function of the field.
+func (b *stringBuilder) DefaultFunc(fn func() string) *stringBuilder {
+	b.desc.Default = fn
+	return b
+}
+
 // Nillable indicates that this field is a nillable.
 // Unlike "Optional" only fields, "Nillable" fields are pointers in the generated field.
 func (b *stringBuilder) Nillable() *stringBuilder {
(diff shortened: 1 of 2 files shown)


schema/field/field.go
```diff
@@ -208,6 +208,12 @@ func (b *stringBuilder) Default(s string) *stringBuilder {
 	return b
 }
 
+// DefaultFunc sets the default value generating function of the field.
+func (b *stringBuilder) DefaultFunc(fn func() string) *stringBuilder {
+	b.desc.Default = fn()
+	return b
+}
```
a8m, line 215: The function needs to be invoked by the creator and not by the schema.
```suggestion
func (b *stringBuilder) DefaultFunc(fn func() string) *stringBuilder {
	b.desc.Default = fn
	return b
}
```

Start with this change and I can help you to add it to the codegen templates.

haoxins: done

haoxins: cc @a8m

a8m (approved):
Thanks for adding this @haoxins 

Some stuff still miss in the templates, but I can add them.  I'll merge this later, and ping you to update the docs.

nolotz:
Hey @haoxins,

in #1086 I've made the missing changes needed in the code generation to get default functions work.

This ticket looks already abandoned, so I've included your changes as well. Please let me know if you would rather commit to these changes.
Otherwise, we should consider closing this ticket and merge my request as it offers a more general solution to the problem and will work with all Descriptors.

cc @a8m 

best regards

haoxins:
cool

PR #973, opened 2020-11-23
Added "inet" example in faq

hugmouse:
Requested here: https://github.com/facebook/ent/issues/963#issuecomment-731715046

But for this example to work we need to do `go get github.com/facebook/ent@c1de826324c57262c6477c51520f8effb0a925fc`


a8m (approved):
Thanks for adding this @hugmouse

PR #982, opened 2020-11-25
doc: minor change in getting-started and .gitignore update

kerbelp:
Updated doc and .gitignore
verified that docs looks good in localhost.

### DIFF
--- a/doc/md/getting-started.md
+++ b/doc/md/getting-started.md
@@ -24,6 +24,7 @@ go get github.com/facebook/ent/cmd/entc
 ```
 
 After installing `entc` (the code generator for `ent`), you should have it in your `PATH`.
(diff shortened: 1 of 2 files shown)


doc/md/getting-started.md
```diff
@@ -24,6 +24,7 @@ go get github.com/facebook/ent/cmd/entc
 ```
 
 After installing `entc` (the code generator for `ent`), you should have it in your `PATH`.
+If you don't find it your path, add: `PATH=$PATH:~/go/bin/` to `~/.bash_profile` (or `~/.profile`)
```
a8m, line 27: I think it's better to tell users to use `go run  github.com/facebook/ent/cmd/entc <command>` to make sure they are using the same ent version as defined in `go.mod` - the PATH can point to a older version of entc, and sometimes, we see users have troubles with it in the issue tracker.

PR #991, opened 2020-11-27
fix: unable to create index on an ID field

### LINKED ISSUE #988: bug: unknown field for id
I ran into this when trying to index the id field of all my schemas, which was derived from a base mixin.

I am getting `entc/gen: invalid index for schema "User": unknown index field "id"` for that field, even though I know it exists and so does the database.

I tested, and this is present when the `id` field is not provided (default), when it is explicitly defined in the schema, and when it is provided by a mixin. Other fields work fine.

<details>
<summary><code>mixin.go</code></summary>

```go
// BaseMixin holds the schema definition for the BaseMixin entity.
type BaseMixin struct {
	mixin.Schema
}

// Fields of the Mixin.
func (BaseMixin) Fields() []ent.Field {
	return []ent.Field{
		field.UUID("id", uuid.UUID{}).Default(uuid.New),
		field.String("someOtherField"),
	}
}

```

OR

```go
// BaseMixin holds the schema definition for the BaseMixin entity.
type BaseMixin struct {
	mixin.Schema
}

// Fields of the Mixin.
func (BaseMixin) Fields() []ent.Field {
	return []ent.Field{
		field.UUID("id", uuid.UUID{}).Default(uuid.New),
		field.String("someOtherField"),
	}
}

func (BaseMixin) Indexes() []ent.Index {
	return []ent.Index{
		index.Fields("id", "someOtherField"),
	}
}
```
</details>

<details>
<summary><code>user.go</code></summary>

```go
// User holds the schema definition for the User entity.
type User struct {
	ent.Schema
}

// Fields of the User.
	return []ent.Field{
		field.String("testField"),
	}
}

// Edges of the User.
	return nil
}

func (User) Indexes() []ent.Index {
	return []ent.Index{
		index.Fields("id"),
	}
}
```

OR

```go
// User holds the schema definition for the User entity.
type User struct {
	ent.Schema
}

func (User) Mixin() []ent.Mixin {
	return []ent.Mixin{
		BaseMixin{},
	}
}

// Fields of the User.
	return []ent.Field{
		field.String("testField"),
	}
}

// Edges of the User.
	return nil
}

func (User) Indexes() []ent.Index {
	return []ent.Index{
		index.Fields("id"),
	}
}
```
</details>

napei:
Co-authored-by: Ciaran Liedeman <ciaran@stackworx.io>

Closes #988 

Added a check inside to determine if an ID field is being used.
Added an integration test for indexes using non-default IDs provided by a mixin


entc/gen/type.go
```diff
@@ -484,10 +484,15 @@ func (t *Type) AddIndex(idx *load.Index) error {
...
+			// Might be the ID field, so check
+			if t.ID.Name == name {
+				f = t.ID
+			} else {
+				return fmt.Errorf("unknown index field %q", name)
+			}
 		}
-		if f.def.Size != nil && *f.def.Size > schema.DefaultStringLen {
+		if f.def != nil && f.def.Size != nil && *f.def.Size > schema.DefaultStringLen {
 			return fmt.Errorf("field %q exceeds the index size limit (%d)", name, schema.DefaultStringLen)
```
a8m, line 496: This one is not needed, just use `NewType` in the test. 

```suggestion
		if def.Size != nil && *f.def.Size > schema.DefaultStringLen {
			return fmt.Errorf("field %q exceeds the index size limit (%d)", name, schema.DefaultStringLen)
```

a8m (approved):
Looks good @napei.

Please add an integration-test for it as well. Thanks!

entc/gen/type.go
napei: Probably should be `f.def.Size` not `def.Size`.

Also, I added this in because when I added the change to use `t.ID` as `f`, `f.def` was `nil` and hence caused a panic.

Happy to remove it but from what I could tell, `def` is not assigned on the `ID` field.

napei:
> Looks good @napei.
> 
> Please add an integration-test for it as well. Thanks!

Slightly confused about integration tests for this one. Is this something to add in `entc/integration/index_test.go`? or a separate test suite just for testing indexes?

entc/gen/type.go
napei: So removing that again actually brings back that error. This is from the PR action, but I can also repro locally.

image

cliedeman:
I came across this issue too in testing a bug I found related to indexes on #988

@napei adding this to entc/integration/ent/schema/card.go

```go
func (Card) Indexes() []ent.Index {
	return []ent.Index{
		index.Fields("id", "name"),
		index.Fields("id", "name", "number"),
	}
}
```

Causes the error.

 A better fix might be to add some defaults to the default id field at creation time

https://github.com/facebook/ent/blob/master/entc/gen/type.go#L175

```go	
ID: &Field{
			Name:      "id",
			def: &load.Field{
				Name: "id",
			},
			Type:      idType,
			StructTag: structTag("id", ""),
}
```

napei:
> I came across this issue too in testing a bug I found related to indexes on #988

Did you open a PR to fix my issue #988 that has had a PR open for 20 days now or did you reference the wrong issue.

cliedeman:
@napei I missed your pr and only just realized its the same issue

napei:
> A better fix might be to add some defaults to the default id field at creation time
> 
> https://github.com/facebook/ent/blob/master/entc/gen/type.go#L175
> 
> ```go
> ID: &Field{
> 			Name:      "id",
> 			def: &load.Field{
> 				Name: "id",
> 			},
> 			Type:      idType,
> 			StructTag: structTag("id", ""),
> }
> ```

@a8m will adding this break anything or should `def` be derived some other way?

cliedeman:
Also there are 2 issues here.

#988 fixes the id column/indexes when the id is added via a mixin.

There is a second issue where you get the above error when you don't use a custom id

napei:
@cliedeman I'll add your integration test to this PR if you're happy with that, and I'll co-author you.

entc/gen/type.go
```diff
@@ -172,7 +172,10 @@ func NewType(c *Config, schema *load.Schema) (*Type, error) {
 	typ := &Type{
 		Config: c,
 		ID: &Field{
-			Name:      "id",
+			Name: "id",
```
cliedeman, line 175: You will need to an extra test case for this scenario, one where the default id field is used

entc/gen/type.go
```diff
@@ -484,8 +487,13 @@ func (t *Type) AddIndex(idx *load.Index) error {
 	}
 	for _, name := range idx.Fields {
 		f, ok := t.fields[name]
-		if !ok {
-			return fmt.Errorf("unknown index field %q", name)
+		if !ok || f == nil {
```
cliedeman, line 490: can you invert this check like in #988

entc/gen/type.go
napei: There is already one present in `type_test.go`

https://github.com/facebook/ent/pull/991/files#diff-5c305024b752f5d6230be7644ec718d9a78debb409e9bd56ebeb69dff6bbb34eR224-R226

Do you mean an extra integration test?

cliedeman: Yup. Like the card one mentioned below

napei: @cliedeman added in 083aeec and 19187e5

a8m (approved):
Thanks @cliedeman and @napei for the amazing collaboration.

PR #1014, opened 2020-12-03
entc/gen: add missing braces in mutation template

BrentChesny:


a8m (approved):
Thanks for fixing this @BrentChesny 🙂

PR #1022, opened 2020-12-07
entc/internal: fix snapshot schema merge

BrentChesny:


a8m (approved):
Thanks for fixing this @BrentChesny 😃

PR #1041, opened 2020-12-14
Repro/bug

cliedeman:
Came across this while trying to setup https://github.com/volatiletech/boilbench for ent

```
err := client.Node.Update().Where(node.ID(nd.ID)).Exec(ctx)
```

The code does not actually make sense but it exposes a weird bug.
I think it only happens on models with foreign keys

### DIFF
--- a/entc/integration/integration_test.go
+++ b/entc/integration/integration_test.go
@@ -135,6 +135,7 @@ var (
 		EagerLoading,
 		Mutation,
 		CreateBulk,
+		Issue1,
 	}
 )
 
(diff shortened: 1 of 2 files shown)


a8m:
Thanks for reporting this @cliedeman. Do you want me to take a look on this, or you work on a fix?

cliedeman:
@a8m I tried to but couldn't figure it out

I think its an edge case in this code block when multiple is false and the `ids` variable is never populated

https://github.com/facebook/ent/blob/master/dialect/sql/sqlgraph/graph.go#L659

a8m:
See https://github.com/facebook/ent/pull/1043

a8m:
Thanks for reporting this @cliedeman. You contribute a lot for this project and it's really appreciated!

Issue #869, opened 2020-10-21
proposal: add an optional to load edge-ids (foreign-keys) in query builder


aight8:
Maybe this is a uniqueness of relational DB's. But edge ID's are a part of a row (Assuming the foreign key is on the side of the rows table).
However this has big benefits:
- You can generally check if a relation exists
- I can use the relation ID for various things (where the ID only is enough, links/actions/other ID only deterministic things like loading an image, and the ID is a hash)

Is it possible somehow?

When not, maybe it would be nice to make this accessible. If graph DB's don't support this, maybe behind a function. Mostly those are 5 int/varchar fields to also select in a simple query. I really missing this one.

a8m:
Hey @aight8 and thanks for proposing this issue. 
It was actually raised before in the issue tracker a few times, so I think we need to address this.

I have a few ideas how such API should look like (and I'll write them later), but you're welcome to write here your suggestions as well.

aight8:
huck, sry I posted this issue by mistake... I wrote it as part of my other "Feedback" issue.

errorhandler:
I've run into this while moving an existing codebase to ent. Here are some ideas for how this could be solved.

## 1. Exposing the foreign keys

I believe being able to access the foreign key values is important so I would suggest exporting the foreign key values

```go
type Pet struct {
	config `json:"-"`
	// ID of the ent.
	ID int `json:"id,omitempty"`
	// Name holds the value of the "name" field.
	Name string `json:"name,omitempty"`
	// Edges holds the relations/edges for other nodes in the graph.
	// The values are being populated by the PetQuery when eager-loading is set.
	Edges     PetEdges `json:"edges"`
	UserPets *int
	UserTeam *int
}
```

## 2. Match nilability in struct

As far as I understand, all FKs in the generated struct are pointers as they are `nil` until the edges are eager-loaded. To make 1. actually usable I believe it should match the edge'se required state.

## 3. Expose predicates for querying foreign keys

1. and 2. are an approach to solve the read path I believe. The only way I've found to do the equivalent of:

```sql
SELECT * FROM `pets` WHERE user_pets = ?
```

is as follows:

```go
client.Pet.Query().Where(func(selector *sql.Selector) {
		selector.Where(sql.EQ(selector.C(pet.OwnerColumn), ownerID))
}).All(ctx)
```

I think there is value in adding generated predicates so you can write the following:

```go
client.Pet.Query().Where(pet.HasOwnerWithID(ownerID)).All(ctx)
```

[...]

PR #1046, opened 2020-12-15
Expose foreign keys via feature flag

errorhandler:
Following the discussion in #869 I've gone ahead and given 1. and 2. a shot, with 3. coming in a follow-up PR. I've added a basic test case and put this feature behind a feature flag.

Any feedback is greatly appreciated! Especially in the general approach.


entc/gen/feature.go
```diff
@@ -43,11 +43,20 @@ var (
 		},
 	}
 
+	FeatureExposeFKs = Feature{
+		Name:        "exposefks",
```
errorhandler, line 47: Any better names would be greatly appreciated.

entc/gen/template/dialect/sql/create.tmpl
```diff
@@ -67,7 +67,16 @@ func ({{ $receiver }} *{{ $builder }}) createSpec() (*{{ $.Name }}, *sqlgraph.Cr
 			_spec.Edges = append(_spec.Edges, edge)
 		}
 	{{- end }}
-	return _node, _spec
+	{{- if ($.FeatureEnabled "exposefks") }}
+	{{- range $fk := $.ForeignKeys }}
+	{{- with $e := $.EdgeForForeignKey $fk }}
+		if id, ok := {{ $mutation }}.{{ $e.StructField }}ID(); ok {
+			_node.{{ $fk.Field.StructField }} = {{ if $fk.Field.Nillable }}&{{ end }}id
```
errorhandler, line 74: If we're exposing the foreign keys, set the field if the foreign key was set in the mutation.

entc/gen/template/dialect/sql/decode.tmpl
```diff
@@ -67,8 +67,12 @@ func ({{ $receiver }} *{{ $.Name }}) assignValues(values ...interface{}) error {
 						if value, ok := values[{{ $i }}].(*sql.NullInt64); !ok {
 							return fmt.Errorf("unexpected type %T for edge-field {{ $f.Name}}", value)
 						} else if value.Valid {
-							{{ $receiver }}.{{ $f.Name }} = new({{ $f.Type }})
-							*{{ $receiver }}.{{ $f.Name }} = {{ $f.Type }}(value.Int64)
+							{{- if $f.Nillable }}
```
errorhandler, line 70: Handle both pointers and value foreign key fields.

entc/gen/template/dialect/sql/query.tmpl
```diff
@@ -7,7 +7,9 @@ in the LICENSE file in the root directory of this source tree.
 {{/* Additional fields for the builder. */}}
 {{ define "dialect/sql/query/fields" }}
 	{{- with $.ForeignKeys }}
-		withFKs bool
+		{{- if not ($.FeatureEnabled "exposefks") }}
+			withFKs bool
```
errorhandler, line 11: This is always true if the FF is enabled, so don't generated it if the FF is on.

entc/gen/template/dialect/sql/query.tmpl
```diff
@@ -40,15 +45,22 @@ func ({{ $receiver }} *{{ $builder }}) sqlAll(ctx context.Context) ([]*{{ $.Name
 			if withFKs {
 				_spec.Node.Columns = append(_spec.Node.Columns, {{ $.Package }}.ForeignKeys...)
 			}
+			{{- else }}
+			_spec.Node.Columns = append(_spec.Node.Columns, {{ $.Package }}.ForeignKeys...)
```
errorhandler, line 49: Always load the FK if the FF is set.

alexsn:
I dislike APIs where there are multiple ways to do the same thing. Currently if we want the edge id of an object be must do `<Object>.Query<Edge>.OnlyID()` and I'd prefer if this will remain the only way. So I'd suggest the FK remain unexported and the `.OnlyID` query finisher will check if we can fetch the ID without db call (i.e. see if it's cached on those unexported fields).

a8m:
> Following the discussion in #869 I've gone ahead and given 1. and 2. a shot, with 3. coming in a follow-up PR. I've added a basic test case and put this feature behind a feature flag.
> 
> Any feedback is greatly appreciated! Especially in the general approach.

I'd appreciate if you can add a summary how do you expect/plan the API to look like. Currently, I'm using custom templates to implement an API with similar functionality:
```go
client.User.
    Query().
    WithEdgeIDs() // We can rename it to "WithFKs".
    All(ctx)

// Then, use it as follows:
id, err := client.CardID()
```

a8m:
Also, it's part of a wider change that will allow selecting specific node fields when querying data - much more efficient in some cases (GraphQL in my case).

G-Bruin:
@a8m When will this new feature be available？

a8m:
> Also, it's part of a wider change that will allow selecting specific node fields when querying data - much more efficient in some cases (GraphQL in my case).

This change should be added with https://github.com/facebook/ent/pull/1075, and we can rebase the work here on top of it. 

Thanks for working on it @errorhandler

a8m:
> > Also, it's part of a wider change that will allow selecting specific node fields when querying data - much more efficient in some cases (GraphQL in my case).
> 
> This change should be added with #1075, and we can rebase the work here on top of it.
> 
> Thanks for working on it @errorhandler

Hey @errorhandler, any update on this PR? Do you need help with this? 
Let me know if you don't have time to finish it, and I can continue it. Thanks 😃

errorhandler:
> > > Also, it's part of a wider change that will allow selecting specific node fields when querying data - much more efficient in some cases (GraphQL in my case).
> > 
> > 
> > This change should be added with #1075, and we can rebase the work here on top of it.
> > Thanks for working on it @errorhandler
> 
> Hey @errorhandler, any update on this PR? Do you need help with this?
> Let me know if you don't have time to finish it, and I can continue it. Thanks 😃

Hi @a8m! I was going to finish it up after New Years, however I've had a death in the family which means I have not be able to work on this. I'll have to pass this over to you I'm afraid.

Thanks for all your hard work on ent @a8m!

a8m:
> Hi @a8m! I was going to finish it up after New Years, however I've had a death in the family which means I have not be able to work on this. I'll have to pass this over to you I'm afraid.
> 
> Thanks for all your hard work on ent @a8m!

Hey @errorhandler, it made me sad to read it yesterday 😢. I'm really sorry for your loss and hope you and your family are safe and healthy.

Thanks for taking the time to work on this. I really appreciate it 🙏

a8m:
Closing as this was addressed with #1213 

Thanks!

Issue #1056, opened 2020-12-18
mixin: ent generate usage code when use GoType in mixin


giautm:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/facebook/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

https://github.com/giautm/ent-bug/commit/5f09a8f7531ac075962f06299a208ea1333334e1#diff-11a3c2971f7c8e5caf8efd23382aa8b7d930011f8189d701b006d4bafadea043

<img width="779" alt="Screen Shot 2020-12-18 at 13 16 19" src="https://user-images.githubusercontent.com/12751435/102581497-47e10980-4133-11eb-89ef-91662dea3cfa.png">

## Expected Behavior 🤔

Dont generate usage code.

## Steps to Reproduce 🕹

Steps:

1. Create custom enum type
2. Create mixin contain enum in step 1)
3. Use mixin in scheme defination
4. Run `go generate`

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.15.4  |
| Ent         | 0.5.3   |
| Database    | Mysql   |
| Driver      | https://github.com/go-sql-driver/mysql        |

a8m:
Thanks for reporting, I’ll give it a look later today.

giautm:
I will sent a PR that update `template.impl` to add one line `_ = {{ $pkg }}MixinFields{{ $i }}`. Is it sound good?

PR #1057, opened 2020-12-18
fix(#1056): ignore non-usage code in runtime.go

giautm:


a8m:
Thanks for your contribution @giautm 😃

PR #1074, opened 2020-12-22
Allow non incremental PKs

saantiaguilera:
This pull requests adds a new feature to allow non autoincremental primary keys.

I've created a new annotation field in entc.Annotation type, of type `*bool`, given that it can be ommited (and it should behave differently according to each particular scenario (PKs vs normal columns).

I've also pondered about using a typed value, since using a bool ptr forces to create additional variables to reference:
```go
incrementalDisabled := false
return []ent.Field{
  field.Int("something").Annotations(entsql.Annotation{
    Incremental: &incrementalDisabled,
  }),
```
With typed values it looked better, but i've opted for the bool ptr because devs could make broken-state types if not.. Like this:
```go
const (
	// incrementalUndefined delegates the incremental property of this column to the internals of the module
	// applying the value that best fits the given scenario
	incrementalUndefined Incremental = iota
	// IncrementalEnabled enables the incremental property of a column
	IncrementalEnabled
	// IncrementalDisabled disables the incremental property of a column
	IncrementalDisabled
)
type Incremental uint8

func BrokenIncrementalValue() {
  b := Incremental(6)
}
```
Also, looked kinda weird given that it's a flag at the end of the day. And the auto-complete features knows it's an underlying uint, so it autocompletes as follows:
```go
a := entc.Annotation{
  Incremental: 0,
}
```

The patch request contains nothing else besides that and tests. I tried giving support also to non-PK columns that wish to be autoincremental, but I don't have enough time to support each of the dialects own caveats (eg. Mysql needing to declare autoincremental columns as keys, or sqlite not supporting them).


a8m:
Thanks for your contribution @saantiaguilera! I'll review this later today/tomorrow morning.

entc/integration/config/config_test.go
```diff
@@ -6,6 +6,9 @@ package template
 
 import (
 	"context"
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"testing"
```
a8m, line 12: Please sort imports.

entc/integration/config/config_test.go
```diff
@@ -41,7 +44,38 @@ func TestSchemaConfig(t *testing.T) {
...
 }
+
+func TestMySQL(t *testing.T) {
+	for version, port := range map[string]int{"56": 3306, "57": 3307, "8": 3308} {
+		t.Run(version, func(t *testing.T) {
+			root, err := sql.Open("mysql", fmt.Sprintf("root:pass@tcp(localhost:%d)/", port))
+			require.NoError(t, err)
+			defer root.Close()
+			ctx := context.Background()
+			err = root.Exec(ctx, "CREATE DATABASE IF NOT EXISTS migrate", []interface{}{}, new(sql.Result))
```
a8m, line 61:
```suggestion
			err = root.Exec(ctx, "CREATE DATABASE IF NOT EXISTS config", []interface{}{}, new(sql.Result))
```

entc/integration/config/config_test.go
```diff
@@ -41,7 +44,38 @@ func TestSchemaConfig(t *testing.T) {
...
+			require.NoError(t, err, "creating database")
+			defer root.Exec(ctx, "DROP DATABASE IF EXISTS migrate", []interface{}{}, new(sql.Result))
```
a8m, line 63:
```suggestion
			defer root.Exec(ctx, "DROP DATABASE IF EXISTS config", []interface{}{}, new(sql.Result))
```

entc/integration/config/config_test.go
```diff
@@ -41,7 +44,38 @@ func TestSchemaConfig(t *testing.T) {
...
+
+			drv, err := sql.Open("mysql", fmt.Sprintf("root:pass@tcp(localhost:%d)/migrate?parseTime=True", port))
```
a8m, line 65:
```suggestion
			drv, err := sql.Open("mysql", fmt.Sprintf("root:pass@tcp(localhost:%d)/config?parseTime=True", port))
```

entc/integration/config/config_test.go
```diff
@@ -41,7 +44,38 @@ func TestSchemaConfig(t *testing.T) {
...
+
+			// Run schema creation.
+			require.NoError(t, client.Schema.Create(ctx))
+
+			u, err := client.User.Create().SetID(200).Save(ctx)
+			require.NoError(t, err)
+			assert.Equal(t, 200, u.ID)
+
+			_, err = client.User.Create().Save(ctx)
+			assert.Error(t, err)
```
a8m, line 78:
```suggestion
			require.NoError(t, err, "connecting to migrate database")

			client := ent.NewClient(ent.Driver(drv))
			// Run schema creation.
			require.NoError(t, client.Schema.Create(ctx))

			u, err := client.User.Create().SetID(200).Save(ctx)
			require.NoError(t, err)
			assert.Equal(t, 200, u.ID)
			_, err = client.User.Create().Save(ctx)
			assert.Error(t, err)
```

a8m (commented):
Looks great @saantiaguilera. Thanks for adding this!

I added a few minor style comments. Please address them and I'll land it.

a8m (approved):

a8m:
Thanks for your contribution @saantiaguilera! It's appreciated.

PR #1085, opened 2020-12-26
dialect/sql/schema: add mysql binary support for native uuid handling

nolotz:
In MySQL 8.0 UUID's are handled via binary columns.


a8m (approved):
Thanks for adding this @nolotz 😃

PR #1102, opened 2020-12-29
entc/gen: handle default funcs in runtime tempalte 

a8m:


a8m:
CI is failing due to some PostgreSQL 11 segfault issue in CircleCI version (latest?) 

image

PR #1104, opened 2020-12-30
fix: Allow Postgres arbitrary last insert id type

cliedeman:


cliedeman:
@a8m I think I will need a postgres only integration suite to test this

Maybe this [example](https://stackoverflow.com/a/27121638)

a8m (approved):
Wonderful. Thanks for adding this @cliedeman.

a8m:
> @a8m I think I will need a postgres only integration suite to test this
> 
> Maybe this [example](https://stackoverflow.com/a/27121638)

We can add a PosgreSQL-only test in `graph_test` in another diff.

PR #1110, opened 2021-01-01
entc/init: name conflict checks

### LINKED ISSUE #1109: add name conflict detection to ent init

## Summary 💡

When running `ent init Client` in the CLI, currently there is no detection for reserved words, allowing this schema to be created. It's not until running `go generate ./...` that the error is raised. 

```
entc/gen: create type Client: schema name conflicts with ent predeclared identifier "Client"
exit status 1
ent/generate.go:3: running "go": exit status 1
```

Perhaps adding in these checks to `ent init` would make it nicer on the user. For example, I run `ent init Client ...........` and then proceed to write the schemas out only to find that there are name conflicts and I have to change things.

napei:
Adds in name conflict checking when running `ent init` to provide better feedback to the user when generating new schemas.
Also made error handling more concise for `init` to show what went wrong when

Closes #1109 

New Error: `ent/init: init schema Client: schema name conflicts with ent predeclared identifier "Client"`

### DIFF
--- a/cmd/internal/base/base.go
+++ b/cmd/internal/base/base.go
@@ -83,7 +83,7 @@ func InitCmd() *cobra.Command {
 		},
 		Run: func(cmd *cobra.Command, names []string) {
 			if err := initEnv(target, names); err != nil {
-				log.Fatalln(err)
+				log.Fatalln(fmt.Errorf("ent/init: %w", err))
 			}
 		},
 	}
@@ -179,16 +179,19 @@ func GenerateCmd(postRun ...func(*gen.Config)) *cobra.Command {
 // initEnv initialize an environment for ent codegen.
 func initEnv(target string, names []string) error {
 	if err := createDir(target); err != nil {
-		return err
+		return fmt.Errorf("create dir %s: %w", target, err)
 	}
 	for _, name := range names {
+		if err := gen.CheckNameConflicts(name); err != nil {
+			return fmt.Errorf("init schema %s: %w", name, err)
+		}
 		b := bytes.NewBuffer(nil)
 		if err := tmpl.Execute(b, name); err != nil {
-			log.Fatalln(err)
+			return fmt.Errorf("executing template %s: %w", name, err)
 		}
-		target := filepath.Join(target, strings.ToLower(name+".go"))
-		if err := ioutil.WriteFile(target, b.Bytes(), 0644); err != nil {
-			log.Fatalln(err)
+		newFileTarget := filepath.Join(target, strings.ToLower(name+".go"))
+		if err := ioutil.WriteFile(newFileTarget, b.Bytes(), 0644); err != nil {
+			return fmt.Errorf("writing file %s: %w", newFileTarget, err)
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


cmd/internal/base/base.go
```diff
@@ -179,16 +179,19 @@ func GenerateCmd(postRun ...func(*gen.Config)) *cobra.Command {
...
 		if err := tmpl.Execute(b, name); err != nil {
-			log.Fatalln(err)
+			return fmt.Errorf("executing template %s: %w", name, err)
 		}
-		target := filepath.Join(target, strings.ToLower(name+".go"))
-		if err := ioutil.WriteFile(target, b.Bytes(), 0644); err != nil {
-			log.Fatalln(err)
+		newFileTarget := filepath.Join(target, strings.ToLower(name+".go"))
+		if err := ioutil.WriteFile(newFileTarget, b.Bytes(), 0644); err != nil {
+			return fmt.Errorf("writing file %s: %w", newFileTarget, err)
```
a8m, line 194: Please keep the previous version and just update the error message.

entc/gen/type.go
```diff
@@ -680,21 +680,29 @@ func (t Type) RelatedTypes() []*Type {
...
+		return fmt.Errorf("schema name conflicts with ent predeclared identifier %q", name)
 	}
 	return nil
 }
 
+// check checks the schema type.
+func (t *Type) check() error {
+	return CheckNameConflicts(t.Name)
+}
+
```
a8m, line 705: You can remove this one and just use the exported function.

entc/gen/type.go
```diff
@@ -680,21 +680,29 @@ func (t Type) RelatedTypes() []*Type {
...
-func (t *Type) check() error {
-	pkg := t.Package()
-	if token.Lookup(pkg).IsKeyword() {
-		return fmt.Errorf("schema lowercase name conflicts with Go keyword %q", pkg)
+// CheckNameConflicts will determine if a name is going to conflict with any
+// pre-defined names
+func CheckNameConflicts(name string) error {
+	// Golang keywords and identifiers are lower case
+	lowerName := strings.ToLower(name)
+
```
a8m, line 688:
```suggestion
	// schema package is lower-cased (see Type.Package).
	pkg := strings.ToLower(name)
```

I prefer to keep it as `pkg` because it explains why this lower-cased identifier is needed.

entc/gen/type.go
```diff
@@ -680,21 +680,29 @@ func (t Type) RelatedTypes() []*Type {
...
 }
 
-// check checks the schema type.
-func (t *Type) check() error {
-	pkg := t.Package()
-	if token.Lookup(pkg).IsKeyword() {
-		return fmt.Errorf("schema lowercase name conflicts with Go keyword %q", pkg)
+// CheckNameConflicts will determine if a name is going to conflict with any
+// pre-defined names
+func CheckNameConflicts(name string) error {
```
a8m, line 685: Maybe use something like `ValidSchemaName` / `CheckSchemaName` / `CheckIdent`? 
```suggestion
func ValidSchemaName(name string) error {
```

cmd/internal/base/base.go
napei: In regards to `target -> newFileTarget`? I only changed that as that `target` variable shadows the `target` parameter in the function.

a8m:
Thanks for adding this @napei, it's really appreciated.

PR #1114, opened 2021-01-02
entc/gen: run goimports concurrently

napei:
This PR changes formatting for generated assets to run concurrently to save time. Personally on a 34-file ent directory the formatting time dropped from 2.8 seconds to 387ms which is a large improvement considering how ent's files increase in size and number based on number of schemas.


entc/gen/graph.go
```diff
@@ -643,17 +644,25 @@ func (a assets) write() error {
...
-			return fmt.Errorf("format file %s: %v", path, err)
-		}
-		if err := ioutil.WriteFile(path, src, 0644); err != nil {
-			return fmt.Errorf("write file %s: %v", path, err)
-		}
+	var eg errgroup.Group
+
+	for i := range a.files {
+		idx := i
+		f := a.files[idx]
```
tmc, line 651: it's a bit odd to declare this outside the func below.

entc/gen/graph.go
```diff
@@ -643,17 +644,25 @@ func (a assets) write() error {
...
+		eg.Go(func() error {
+			path := f.path
+			src, err := imports.Process(path, f.content, nil)
+			if err != nil {
+				return fmt.Errorf("format file %s: %v", path, err)
```
tmc, line 656: I think a better format here might be something like "issue with processing path '%v': %w", path, err

entc/gen/graph.go
```diff
@@ -643,17 +644,25 @@ func (a assets) write() error {
...
+			}
+			if err := ioutil.WriteFile(path, src, 0644); err != nil {
+				return fmt.Errorf("write file %s: %v", path, err)
```
tmc, line 659: similar to above, a more clear message might be "issue writing file '%v': %w", ...

entc/gen/graph.go
napei: Has to happen to support errorgroup, the index has to be redeclared each iteration outside the goroutine. Same for the file element. It's just because it's not a `go func(f file){}(a.files[i])`

entc/gen/graph.go
napei: It's actually formatting a file, not processing a path. Perhaps a better alternative is `formatting file %s: %w` which also includes error unwrapping which is supposed to be there anyway. Generally error messages, because they're wrapped, don't include "error doing xyz" as that's extraneous if they all  have that.

Another PR which I'll open will improve the error wrapping in entc, as currently they're just logged with fatal.

https://github.com/facebook/ent/blob/8b9479599d754d6e66ab5bc19885b4091c660b3f/cmd/internal/base/base.go#L162

entc/gen/graph.go
napei: But yes, it can be shortened to `f := a.files[i]`

a8m:
Thanks for adding this @napei (and thanks @tmc for the CR).

I'm going to test it on a few projects next week and I'll update you.

a8m:
> Thanks for adding this @napei (and thanks @tmc for the CR).
> 
> I'm going to test it on a few projects next week and I'll update you.

I ran this on my local machine on 2 projects with hundreds of files, and didn't see any improvement. But when I limited the concurrent goroutines with a semaphore (to 20), I actually saw some improvement (of 20%).

I'll try to investigate it more and update. Like I mentioned in the Slack, I already tried this, and thought about just changing the format logic to write (and format) only files that were changed.

PR #1116, opened 2021-01-02
cmd/ent: fix invalid comment

napei:
There was a comment that was bugging me. 😄


a8m (approved):
Thanks for fixing this @napei 😃

Issue #139, opened 2019-11-04
Feature Request: Upsert


mk0a1a:
Is it possible to have atomic upsert functionality similar to postresql's [INSERT ON CONFLICT](http://www.postgresqltutorial.com/postgresql-upsert/) or mysql's [INSERT ON DUPLICATE KEY UPDATE](https://www.techbeamers.com/mysql-upsert/) statement? Use case is to save roundtrips on remote databases and to minimize lines of code.

a8m:
This issue somehow depends on #127. Although, it's possible to define the conflict on any constraint.

I don't see any problem with adding this after #127 will be addressed.

chris-rock:
@a8m do you have already a syntax for upsert in mind? Since upsert is different in Postgres/Sqlite and Mysql, Ent should probably abstract that common case.

Just a draft of my current thinking:

```
pedro, err := client.Pet.
    UpsertOneID(id).
    Create(
	pet.Create().SetName("pedro")
    ).
    Update(
        pet.Update().SetOwnerID(owner)
    ).
    Save(ctx)
```
This would allow us to reuse the Settters for Create and Update and ensure that Create and Update can include different properties.

jufemaiz:
This looks rather promising.

moh-abk:
Any ETA on this? :)

tomtwinkle:
I want one.

clearcodecn:
I need it.

yuvalt7:
yes please

moh-abk:
Sorry to bug but is this being worked on? 🤦🏾‍♂️

ivanvanderbyl:
Has anyone made a start on this? I'd like to contribute!

ivanvanderbyl:
A few thoughts on this — at least from a Postgres standpoint — using `ON CONFLICT` requires some sort of annotation to indicate that a field is unique (already supported in the schema) or supply it with a named constraint (also generated by using `index.Field(...).Unique()`. Beyond that we need to update the graph inserter with the necessary pieces to write the conflict statement. Assuming something like @chris-rock suggested above, I've started to put this together: 

```go
// Pet holds the schema definition for the Pet entity.
type Pet struct {
	ent.Schema
}

// Fields of the Pet.
func (Pet) Fields() []ent.Field {
	return []ent.Field{
		field.String("name"),
	}
}

func (Pet) Indexes() []ent.Index {
	return []ent.Index{
		index.Fields("name").Unique(),
	}
}

// Usage:
[...]

PR #1121, opened 2021-01-03
(WIP) Upserts (Resolving #139)

ivanvanderbyl:
This PR aims to implement database level upserts using `ON CONFLICT ...` (Postgres) and equivalent in MySQL, adding new DSL methods for declaring an Upsert operation:

```go
// Pet holds the schema definition for the Pet entity.
type Pet struct {
	ent.Schema
}

// Fields of the Pet.
	return []ent.Field{
		field.String("name"),
	}
}

func (Pet) Indexes() []ent.Index {
	return []ent.Index{
		index.Fields("name").Unique(),
	}
}

// Usage:
pedro, err := client.Pet.
    Upsert().
    Create().SetName("pedro").
    Update().SetName("pedro-2").
    Save(ctx)

```

Resolving #139

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -620,6 +620,11 @@ type InsertBuilder struct {
 	defaults  string
 	returning []string
 	values    [][]interface{}
+
+	// Upsert
+	conflictColumns []string
+	updateColumns   []string
+	updateValues    []interface{}
 }
 
 // Insert creates a builder for the `INSERT INTO` statement.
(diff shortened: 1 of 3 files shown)


cliedeman:
@ivanvanderbyl I started work on it here https://github.com/cliedeman/ent/tree/feat/upsert but don't plan on continuing with it.

I think the `.Upsert` api you are using is the correct one

dialect/sql/builder.go
```diff
@@ -692,6 +710,27 @@ func (i *InsertBuilder) Query() (string, []interface{}) {
...
+		for j, c := range i.updateColumns {
+			if j > 0 {
+				i.Comma()
+			}
+			i.Ident(c).WriteString(" = ")
+			switch v := i.updateValues[j].(type) {
+			case Querier:
+				i.Join(v)
+			default:
+				i.Arg(v)
```
cliedeman, line 729: Instead of using arg you should use the dialect specific update syntax so that this will work with multiple rows.

Postgres
```
insert into segments(id, departure_hour) 
values 
    (1153, 2), 
    (1156, 4), 
    (1154, 2) 
on conflict do update set 
    departure_hour = excluded.departure_hour;
```

Mysql
https://stackoverflow.com/questions/2714587/mysql-on-duplicate-key-update-for-multiple-rows-insert-in-single-query

Not sure if sqlite has an equivalent

ivanvanderbyl: Thanks for the input, I'll take a look at that, and I like the approach you took to supporting multiple dialects.

ivanvanderbyl:
After thinking about this overnight, I think @a8m's syntax would be cleaner and simpler to implement:

```go
pedro, err := client.Pet.
    Create().
    SetName("pedro").
    OnConflict().
    SetName("pedro-2").
    Save(ctx)
```

There's a slight difference between how MySQL and Postgres handle conflicts: in MySQL it is assumed that you will conflict on the primary key and any unique index, so specifying conflict targets is optional. In Postres it's required.

There are multiple ways we could handle this: 
- Be explicit like Postgres:`OnConflict(pet.NameField)`
- Use Indexes and Primary keys by default if they have a value set in the mutation
- Database annotations

ivanvanderbyl:
Duplicated by #1325

PR #1122, opened 2021-01-03
schema/field: relax signature of DefaultFunc for String() and Bytes()

tmc:
Using `GoType()` will generate code that expects the provided type throughout generated code -- it's not appropriate to restrict the signature of DefaultFunc in these cases.

This change relaxes the signature of DefaultFunc for `stringBuilder` and `bytesBuilder` and does a reflect-based check inside of `Descriptor()`.

### DIFF
--- a/schema/field/field.go
+++ b/schema/field/field.go
@@ -600,7 +600,8 @@ func (b *bytesBuilder) Descriptor() *Descriptor {
 	// If the Default is present and a function, check that it's signtaure is appropriate.
 	if b.desc.Default != nil {
 		typ := reflect.TypeOf(b.desc.Default)
-		if typ.Kind() == reflect.Func && (typ.NumIn() != 0 || typ.NumOut() != 1 || typ.Out(0).String() != b.desc.Info.String()) {
+		// We check against "[]uint8" here because the reflect package treats byte as an alias to uint8.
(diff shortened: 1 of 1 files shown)


schema/field/field.go
```diff
@@ -298,6 +298,27 @@ func (b *stringBuilder) Descriptor() *Descriptor {
...
+// of the field on creation. For example:
+//
+//	field.String("cuid").
+//		DefaultFunc(cuid.New)
+//
+func (b *customTypeStringBuilder) DefaultFunc(fn interface{}) *customTypeStringBuilder {
+	typ := reflect.TypeOf(fn)
+	if typ.Kind() != reflect.Func || typ.NumIn() != 0 || typ.NumOut() != 1 || typ.Out(0).String() != b.desc.Info.String() {
+		b.desc.Err = fmt.Errorf("expect type (func() %s) for default value", b.desc.Info)
+	}
```
a8m, line 317: What do you think about moving this check to the `Descriptor` method, and changing the function signature of the `stringBuilder`? I think we need to run this check even if the `GoType` option is called after the default function.

tmc: That sounds cleaner to me, let me prep that.

tmc: As I'm adding this it seems as though this should exist on the other builder types as well..

a8m:
Hey @tmc, I merged this PR to https://github.com/facebook/ent/pull/1125, fixed a an issue and added tests for it.

PR #1129, opened 2021-01-05
Allow construction time schema options

marwan-at-work:
This PR allows ent users to pass alternate database names for individual tables, such as: 

```golang
client, _ := ent.Open(
  "mysql",
  "root@password:localhost:3306/db1",
  ent.WithUserSchema("db2",
  ent.WithCarSchema("db3"),
)
```

This will end up sending queries to the DB server as such: 

```sql
SELECT * FROM `db2`.`users`;
INSERT INTO `db3`.`cars` ...;
```

Note that I have not written tests yet for a couple of reasons: 

1. Would love to get the main logic reviewed first to ensure it looks okay before writing any tests. 
2. I haven't actually taken the time to look at how testing is done. 

If things look okay in this PR, I'd be happy to add tests. 

PS. I've separated the "go generate" command into a separate commit so that this PR can be reviewed more easily. 

Fixes https://github.com/facebook/ent/issues/663


dialect/sql/builder.go
```diff
@@ -616,6 +616,7 @@ func (d *DropIndexBuilder) Query() (string, []interface{}) {
 type InsertBuilder struct {
 	Builder
 	table     string
+	schema    string
```
a8m, line 619: Let's move all the changes in this file to a separate PR, and add tests for it in the `builder_test.go` file.

dialect/sql/sqlgraph/graph.go
```diff
@@ -825,7 +827,7 @@ type creator struct {
 func (c *creator) node(ctx context.Context, tx dialect.ExecQuerier) error {
 	var (
 		edges  = EdgeSpecs(c.Edges).GroupRel()
-		insert = c.builder.Insert(c.Table).Default()
+		insert = c.builder.Insert(c.Table).Schema(c.Schema).Default()
```
a8m, line 830: Let's move all changes in this file to a separate PR with tests in `graph_test.go` - ping me in Slack if you need help with this. 

Also, some parts are missing here, but I can help with this or can guide you.

entc/gen/template/config.tmpl
```diff
@@ -24,6 +24,10 @@ type config struct {
 	log func(...interface{})
 	// hooks to execute on mutations.
 	hooks *hooks
+	{{- range $.Nodes }}
+		// optional schema name for the {{ .Name }} table.
+		{{ .Name }}Schema string
+	{{- end }}
```
a8m, line 30: 3 comments about this - 
1. I think this should be storage-specific configuration (should move to `dialect/sql/config.tmpl`)
2. I'm not sure this should be added by default to the codegen. There are 2 options: 1) provide a feature-flag to enable it to all schemas 2) add a new annotation to `entsql` that lets users to enable it to specific schemas (shared mixin configuration for all schemas)
3. Struct fields should be private

WDYT?

a8m (commented):
wow

PR #1134, opened 2021-01-05
Add schema options for sql builders

marwan-at-work:
This PR adds schema options to all the SQL builders (insert, delete, update etc) to pave the way for users to customize the database name for each table. 


a8m (approved):
Looks good, thanks for adding this @marwan-at-work 😃

PR #1135, opened 2021-01-05
Skip schema prefix when dialect is SQLite

marwan-at-work:
SQLite does not have schema prefixes when referencing a table, make sure the Builders account for this case.

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -1819,7 +1819,6 @@ func (s *Selector) Query() (string, []interface{}) {
 	b.WriteString(" FROM ")
 	switch t := s.from.(type) {
 	case *SelectTable:
-		fmt.Println("ONE", s.dialect)
 		t.SetDialect(s.dialect)
 		b.WriteString(t.ref())
 	case *Selector:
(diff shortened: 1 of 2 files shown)


dialect/sql/builder.go
```diff
@@ -1819,6 +1819,7 @@ func (s *Selector) Query() (string, []interface{}) {
 	b.WriteString(" FROM ")
 	switch t := s.from.(type) {
 	case *SelectTable:
+		fmt.Println("ONE", s.dialect)
```
a8m, line 1822:
```suggestion
```

Remove please.

dialect/sql/builder.go
```diff
@@ -682,7 +682,7 @@ func (i *InsertBuilder) Returning(columns ...string) *InsertBuilder {
 // Query returns query representation of an `INSERT INTO` statement.
 func (i *InsertBuilder) Query() (string, []interface{}) {
 	i.WriteString("INSERT INTO ")
-	if i.schema != "" {
+	if i.schema != "" && i.dialect != dialect.SQLite {
 		i.Ident(i.schema).WriteByte('.')
 	}
```
a8m, line 687: Maybe moving this code block to the `Builder`, and then calling it from all builders as follows:

```go
func (b *Builder) writeSchema(s schema)  {
	if schema != "" && u.dialect != dialect.SQLite {
		b.Ident(u.schema).WriteByte('.')
	}
}

// ..

u.writeSchema(u.schema)
```

marwan-at-work: @a8m done 👍🏼

a8m (approved):
Thanks for your contribution @marwan-at-work 😃

PR #1136, opened 2021-01-05
Add schema options for sql graph

marwan-at-work:
This PR adds schema options to all graph operations

### DIFF
--- a/dialect/sql/sqlgraph/graph.go
+++ b/dialect/sql/sqlgraph/graph.go
@@ -117,22 +117,6 @@ func To(table, column string) StepOption {
 	}
 }
 
-// ToSchema sets the db name of the From
-// step
-func ToSchema(name string) StepOption {
-	return func(s *Step) {
-		s.To.Schema = name
-	}
-}
-
-// EdgeSchema sets the db name of the EDge
-// table
-func EdgeSchema(name string) StepOption {
-	return func(s *Step) {
-		s.Edge.Schema = name
-	}
-}
-
 // Edge sets the edge info for getting the neighbors.
 func Edge(rel Rel, inverse bool, table string, columns ...string) StepOption {
 	return func(s *Step) {
--- a/dialect/sql/sqlgraph/graph.go
+++ b/dialect/sql/sqlgraph/graph.go
@@ -71,6 +71,9 @@ type Step struct {
 	Edge struct {
 		// Rel of the edge.
 		Rel Rel
+		// Schema is an optional name of the database
+		// where the table is defined.
+		Schema string
 		// Table name of where this edge columns reside.
 		Table string
 		// Columns of the edge.
@@ -84,6 +87,9 @@ type Step struct {
 	To struct {
 		// Table holds the table name of the neighbors (to).
 		Table string
+		// Schema is an optional name of the database
+		// where the table is defined.
+		Schema string
 		// Column to join with. Usually the "id" column.
 		Column string
 	}
@@ -108,6 +114,22 @@ func To(table, column string) StepOption {
 	return func(s *Step) {
 		s.To.Table = table
 		s.To.Column = column
+	}
+}
+
+// ToSchema sets the db name of the From
+// step
+func ToSchema(name string) StepOption {
+	return func(s *Step) {
+		s.To.Schema = name
+	}
+}
+
+// EdgeSchema sets the db name of the EDge
+// table
+func EdgeSchema(name string) StepOption {
+	return func(s *Step) {
+		s.Edge.Schema = name
 	}
 }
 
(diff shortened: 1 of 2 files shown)


dialect/sql/sqlgraph/graph.go
```diff
@@ -84,6 +87,9 @@ type Step struct {
 	To struct {
 		// Table holds the table name of the neighbors (to).
 		Table string
+		// Schema is an optional name of the database
+		// where the table is defined.
+		Schema string
```
a8m, line 92: You need to add the `Schema` field to the `Form` struct as well.

dialect/sql/sqlgraph/graph.go
```diff
@@ -111,6 +117,22 @@ func To(table, column string) StepOption {
...
+}
+
+// EdgeSchema sets the db name of the EDge
+// table
+func EdgeSchema(name string) StepOption {
+	return func(s *Step) {
+		s.Edge.Schema = name
+	}
+}
+
```
a8m, line 135: Let's start without these options for now, and just change the actual `Step` in the tests. For example:
```go
input: func() *Step {
	step :=  NewStep(
		From("nodes", "id", 1),
		To("nodes", "id"),
		Edge(O2O, true, "nodes", "prev_id"),
	),
	step.To.Schema = "db1"
	return step
}()
```

I want to see how we're going to use it in the codegen, because I don't really like these options (prefer to just change the actual ones).

dialect/sql/sqlgraph/graph.go
```diff
@@ -157,17 +179,17 @@ func Neighbors(dialect string, s *Step) (q *sql.Selector) {
 			Join(match).
 			On(to.C(s.To.Column), match.C(pk1))
 	case r == M2O || (r == O2O && s.Edge.Inverse):
-		t1 := builder.Table(s.To.Table)
+		t1 := builder.Table(s.To.Table).Schema(s.To.Schema)
 		t2 := builder.Select(s.Edge.Columns[0]).
-			From(builder.Table(s.Edge.Table)).
+			From(builder.Table(s.Edge.Table).Schema(s.Edge.Schema)).
 			Where(sql.EQ(s.From.Column, s.From.V))
```
a8m, line 185:
```suggestion
			From(builder.Table(s.Edge.Table).
			Schema(s.Edge.Schema)).
			Where(sql.EQ(s.From.Column, s.From.V))
```

Add \n (for other places as well)

a8m (commented):
Amazing work @marwan-at-work. Thanks a lot for your contribution 😃 

Please fix this minor comments and I'll re-review this.

dialect/sql/sqlgraph/graph.go
marwan-at-work: @a8m from all the tests that I saw, the `Table` field in the `From` struct never actually ends up in the final SQL statement...This is mainly because the "From" comes from a `*sql.Selector` which already has a `sql.TableView` with the Table/Schema inside. 

I'm happy to ad the Schema field but I just wanted to note that in case I'm missing something :)

dialect/sql/sqlgraph/graph.go
marwan-at-work: @a8m for sure, just FYI gofmt will not end up formatting as you suggested, but more like this: 


```golang
    From(builder.Table(s.Edge.Table).
        Schema(s.Edge.Schema)).
    Where(sql.EQ(s.From.Column, s.From.V))
```

Since `.Schema`  is part of the From input, and not the From return. Let me know if that's okay and I'll change all of them :)

dialect/sql/sqlgraph/graph.go
marwan-at-work: Sounds good, just FYI this is what it looks like so far: https://github.com/marwan-at-work/ent/compare/schemagraph...marwan-at-work:schematemplate?expand=1

dialect/sql/sqlgraph/graph.go
a8m: You're right. Ignore this

dialect/sql/sqlgraph/graph.go
a8m: Let's re-added them after working on the codegen. I still think on other alternative. Sounds good?

dialect/sql/sqlgraph/graph.go
a8m: Oh, right. I missed this. Thanks

a8m (approved):
Wonderful @marwan-at-work! Let's fix the small comment and I'll land this.

dialect/sql/sqlgraph/graph.go
marwan-at-work: @a8m i already have the tests adjusted to the alternative, so i can just push that up :)

dialect/sql/sqlgraph/graph.go
marwan-at-work: Cool cool, will keep them as is then.

marwan-at-work:
@a8m I updated the PR with your suggestion, happy to make more changes if need be. Thanks!

a8m (approved):
🥳

a8m:
Great @marwan-at-work! Let's discuss on the codegen part on Slack, cool?

PR #1142, opened 2021-01-06
add an Annotation to toggle the QuerySpec.Unique for a entity.

rubensayshi:
solves https://github.com/facebook/ent/issues/999

DISTINCT can give very poor performance for big result sets and more expensive to compare columns. 

DISTINCT also doesn't work for some column types which aren't comparable, I was having an issue with using ent to query a postgres view with a `hstore_to_json` which wouldn't work because of the DISTINCT.


rubensayshi:
I've been running this in production for a while, I can rebase if there's actual interest to merge ;)

rubensayshi:
@a8m I still think this is a good PR, we've been using it for a while now and it's simple and a very low cost IMO to have both camps what they want ;)

Antonboom:
+1

mgabeler-lee-6rs:
A new reason for this: there is at least one common type, at least in postgresql, that is not comparable, and thus any table using it will _error_ if you attempt to `SELECT DISTINCT` on it: `json`

Issue #1107, opened 2020-12-30
Naming convention for multi-word fields / edges 


adayNU:
Is there a standard naming convention for fields and/or edges which are more than one word?

Just wondering since I _believe_ its standard to use `snake_case` for things like database columns, but Go uses `camelCase`. My hunch is to use `snake_case` but couldn't find that explicitly called out anywhere.

notzheng:
I hope this can help you.


[](https://github.com/facebook/ent/issues/1069)

a8m:
Hey @adayNU! 

> Is there a standard naming convention for fields and/or edges which are more than one word?

Yes, we use `snake_case` for both fields and edges, and ent automatically "pascalize" them in Go code when it's needed (e.g. struct fields or methods). The database columns (and struct tags) will be the same as you wrote the fields/edge in the schema unless stated otherwise (e.g. manually configured the column and the struct-tag using the `StorageKey` and `StructTag` options). 

Closing, but please feel free to contribute a change for adding the convention section to the [schema fields page](https://entgo.io/docs/schema-fields/). The change can be added [here](https://entgo.io/docs/schema-fields/). 

Thanks 😃

PR #1144, opened 2021-01-06
Update documentation to include naming convention

adayNU:
This change relates to my question in #1107 around naming convention for fields / edges. As a bonus, I added documentation around the `StructTag` method for edges, as it seemed to be omitted.


doc/md/schema-edges.md
```diff
@@ -932,3 +949,8 @@ func (Pet) Edges() []ent.Edge {
 ```
 
 Read more about annotations and their usage in templates in the [template doc](templates.md#annotations).
+
+## Naming Convention
+
+By convention edge names should use `snake_case`. The corresponding struct fields generated by `ent` will follow the Go convention
+of using `camelCase`. In cases where `camelCase` is desired, you can do so with the `StorageKey` or `StructTag` methods.
```
a8m, line 956: I think it should be `PascalCase` because struct fields are exported. Right? 

```suggestion
of using `PascalCase`. In cases where `PascalCase` is desired, you can do so with the `StorageKey` or `StructTag` methods.
```

a8m (approved):
Thanks for adding this @adayNU 😃

doc/md/schema-edges.md
adayNU: 🤦  yep, good point. Done.

a8m (approved):
🎉

PR #1147, opened 2021-01-07
Fix ptr annotation

rubensayshi:
I ran into this issue when I was implementing https://github.com/facebook/ent/pull/1142

when an Annotation value is a pointer, like `*bool Incremental` it was being written to `ent/migate/schema.go` as such, with by `dumpFields` with the string written being the address, which of course doesn't work.

this PR fixes that, I extract it from https://github.com/facebook/ent/pull/1142 because I think it should be merged regardless if #1142 gets accepted.

I went for `boolPtr`, `strPtr`, `intPtr`, but open to any other naming for those functions or we could include the `pointer` library if you don't mind including another dependency.

### DIFF
--- a/entc/gen/func.go
+++ b/entc/gen/func.go
@@ -418,7 +418,12 @@ func dumpFields(v interface{}) string {
 		}
 		fv := rv.Field(i)
 		if !fv.IsZero() {
-			fields = append(fields, fmt.Sprintf("%s: %#v", f.Name, fv.Interface()))
+			if fv.Kind() == reflect.Ptr {
+				fv = reflect.Indirect(fv)
+				fields = append(fields, fmt.Sprintf("%s: &[]%s{%#v}[0]", f.Name, fv.Type(), fv.Interface()))
+			} else {
+				fields = append(fields, fmt.Sprintf("%s: %#v", f.Name, fv.Interface()))
+			}
 		}
 	}
 	return strings.Join(fields, ", ")
--- a/entc/gen/type_test.go
+++ b/entc/gen/type_test.go
@@ -273,6 +273,23 @@ func TestField_DefaultName(t *testing.T) {
 	}
 }
 
+func TestField_incremental(t *testing.T) {
+	tests := []struct {
+		annotations map[string]interface{}
+		def         bool
+		expected    bool
+	}{
+		{dict("EntSQL", nil), false, false},
+		{dict("EntSQL", nil), true, true},
+		{dict("EntSQL", dict("incremental", true)), false, true},
+		{dict("EntSQL", dict("incremental", false)), true, false},
+	}
+	for _, tt := range tests {
+		typ := &Field{Annotations: tt.annotations}
+		require.Equal(t, tt.expected, typ.incremental(tt.def))
+	}
+}
+
 func TestBuilderField(t *testing.T) {
 	tests := []struct {
 		name  string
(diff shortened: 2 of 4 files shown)


entc/gen/func.go
```diff
@@ -418,7 +418,24 @@ func dumpFields(v interface{}) string {
...
+					value = fmt.Sprintf("intPtr(%#v)", fv.Interface())
+				default:
+					value = fmt.Sprintf("%#v", fv.Interface())
+				}
+			} else {
+				value = fmt.Sprintf("%#v", fv.Interface())
+			}
+
+			fields = append(fields, fmt.Sprintf("%s: %s", f.Name, value))
 		}
```
a8m, line 439: This won't work if we call this function from package `migrate`.  I think it will be simpler to just copy the `entsql` annotation in the [init function](https://github.com/facebook/ent/blob/master/entc/gen/template/migrate/schema.tmpl#L113), or we can do something like `F: &[]bool{true}[0]`, which is a bit ugly.

rubensayshi: I like `&[]bool{true}[0]` more than the helpers I guess :P

rubensayshi: changed to `[]...{}[0]`

entc/gen/func.go
```diff
@@ -418,7 +418,24 @@ func dumpFields(v interface{}) string {
...
+				case reflect.Int:
+					value = fmt.Sprintf("&[]int{%#v}[0]", fv.Interface())
+				default:
+					value = fmt.Sprintf("%#v", fv.Interface())
+				}
+			} else {
+				value = fmt.Sprintf("%#v", fv.Interface())
+			}
+
+			fields = append(fields, fmt.Sprintf("%s: %s", f.Name, value))
```
a8m, line 438:
```suggestion
			if fv.Kind() == reflect.Ptr {
				fv = reflect.Indirect(fv)
				fields = append(fields, fmt.Sprintf("&[]%s{%#v}[0]", fv.Type(), fv.Interface()))
			} else {
				fields = append(fields, fmt.Sprintf("%s: %s", f.Name, value))
			}
```

a8m (approved):
I don't like the this approach (using `&[]T{..}[0]`), but I prefer to not block you at the moment. 
So, let's fix this small change and we can improve this later.

rubensayshi:
probably the cleanest solution would be to either use https://github.com/AlekSi/pointer or create the helpers I had before somewhere in the `github.com/facebook/ent/...?`

no need to rush it, let's do it right if you're not happy with this!

entc/gen/func.go
```diff
@@ -418,7 +418,12 @@ func dumpFields(v interface{}) string {
 		}
 		fv := rv.Field(i)
 		if !fv.IsZero() {
-			fields = append(fields, fmt.Sprintf("%s: %#v", f.Name, fv.Interface()))
+			if fv.Kind() == reflect.Ptr {
+				fv = reflect.Indirect(fv)
```
alexsn, line 422: No need to call Indirect, use .Elem() as we already know it's a pointer.
```go
fv = fv.Elem()
```

rubensayshi: fixed, thx, you learn something new every day :D

rubensayshi: actually that doesn't work because without doing `fv = reflect.Indirect(fv)` the `fv.Type()` will be `*bool` which results in `&[]*bool{false}[0]}`

a8m:
I'm going to test it locally on my template changes, and merge it to my branch if it works.

a8m:
Merged to master with 76c5e48ef817c16c8e1e1b2c844881658c243b7f. 

Thanks!

PR #1150, opened 2021-01-07
add hooks to Migrate

rubensayshi:
add hooks to Migrate to allow filtering and mutation of the tables being created.

as per https://github.com/facebook/ent/issues/1079 this will make it possible to change or remove tables before they're created for any custom modifications needed.

```go
	migrate, err := schema.NewMigrate(sql.OpenDB("mysql", db), schema.Hook(func(next CreateFunc) CreateFunc {
		return func(ctx context.Context, tables ...*Table) error {
			return next(ctx, tables[0], &Table{Name: "pets"})
		}
	}))

```

### DIFF
--- a/dialect/sql/schema/migrate.go
+++ b/dialect/sql/schema/migrate.go
@@ -66,6 +66,43 @@ func WithForeignKeys(b bool) MigrateOption {
 	}
 }
 
+type (
+	// Creator is the interface that wraps the Create method.
+	Creator interface {
+		// Create creates tables.
+		Create(context.Context, ...*Table) error
+	}
+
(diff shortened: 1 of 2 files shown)


a8m:
Wow! Thanks for adding this @rubensayshi. Minor style comment, please use the same patterns/naming uses in [ent.Hook](https://github.com/facebook/ent/blob/master/ent.go#L321-L342) and [entc/gen.Hook](https://github.com/facebook/ent/blob/master/entc/gen/graph.go#L70-L91).

rubensayshi:
yea, you're absolutely right, made it in the same style now.

a8m (approved):
Looks good @rubensayshi. Thanks a lot for your contribution!

PR #1145, opened 2021-01-07
Implement Other Field

cliedeman:
Resolves #1105

PR #1151, opened 2021-01-07
Added postgres specific integration tests

cliedeman:
Some features of postgres need to be tested in their own set of integration tests


a8m:
Hey @cliedeman, can you elaborate what kind of tests need this? I know that there are PostgreSQL-specific features or types  that users use, but I want to make it possible for users to run ent also with SQLite (mainly for testing).

cliedeman:
Hi @a8m it actually all boils down to the online migrations.

When the online migrations run they attempt to infer the types and return a different field type to other

```
		field.Other("link_other", &Link{}).
			SchemaType(map[string]string{
				dialect.MySQL:    "varchar",
				dialect.Postgres: "varchar",
				dialect.SQLite:   "text",
			}).
			Optional(),
```

field.TypeStringSelected
https://github.com/facebook/ent/blob/master/dialect/sql/schema/mysql.go#L421
Conversion then fails
https://github.com/facebook/ent/blob/master/dialect/sql/schema/schema.go#L174

You can see #1145 for a concret example. I can't get the new integration tests to pass on sqlite or mysql. Alternatively I could place the integration tests inside a test suite where the migrations are only run once. The issue occurs only on the second migrate

```
sql/schema: changing column type for "link_other" is invalid (text != varchar(255))
```

a8m:
Thanks for the explanation. I'll give it a deeper look. 

>  The issue occurs only on the second migrate

That's good in many cases (e.g. when running unit tests using SQLite).

cliedeman:
Thanks. Also TypeOther should not occur outside Postgres as sqlite and MySQL don’t support custom types. So you should never user field.Other with them

cliedeman:
Hi @a8m have you had a chance to look at this?

a8m:
> Hi @a8m have you had a chance to look at this?

Yes. I just checked it, and setting the "other" field as below fixed this issue:

```go
field.Other("link_other", &Link{}).
	SchemaType(map[string]string{
		dialect.Postgres: "varchar",
		dialect.MySQL:    "varchar(255)",
		dialect.SQLite:   "varchar(255)",
	}).
	Optional(),
```

I think the type-checking of the columns needs to be moved to be a dialect-specific thing. I'll work on it. It'll make the code much more flexible.

cliedeman:
Thanks @a8m. You can also let me know what needs to change and I can give it a shot

a8m:
> Thanks @a8m. You can also let me know what needs to change and I can give it a shot

Just copy my changes to #1145 and it'll solve the test issue.

PR #1153, opened 2021-01-08
schema: add DefaultFunc for numeric types and add support for it in entc

### LINKED ISSUE #1146: proposal: enable setting a DefaultFunc on a numeric type
I'm currently migrating a database from an existing structure where we have used a variation of [Snowflake](https://github.com/sony/sonyflake) to generate the IDs for our resources. The natural way to store such a value in the database would be an uint64/`BIGINT UNSIGNED NOT NULL`, thus a numeric type for ent.

In spite of there not being a straight `DefaultFunc` method on the `uint64Builder`, by getting the Descriptor I managed to set a custom function which is ran when the resource is created:

```go
	fd := field.Uint64("id").Immutable().Positive()
	fd.Descriptor().Default = func() uint64 {
		// generate...
	}
```

However, once we try to run entc, the following error shows up:

```
2021/01/07 10:08:32 running ent codegen:unmarshal schema [... json data ...]: unexpected default value type for field: "id"
```

Tracing back the error, it is from this method:

https://github.com/facebook/ent/blob/d4e04cd08f76389f0db76adec1c62d3a48c36d7b/entc/load/schema.go#L354-L369

Being a numeric type and having a default, the type conversion is attempted, but f.DefaultValue will actually be of type nil because only DefaultFunc is set.

Proposals to solve this:

1. add a `DefaultFunc` check on the `defaults()` method, thus skipping it when we have such a function
2. or, the former + add full support by also adding `DefaultFunc` on all numeric type builders - although this entails expanding the API

I'd be happy to work on a PR for this - but wanted to check first what's the best path to follow here.

thehowl:
Closes #1146.


doc/md/schema-fields.md
```diff
@@ -144,6 +144,19 @@ func (Pet) Fields() []ent.Field {
...
 
+If you need to set a custom function to generate IDs, you can use `DefaultFunc`
+to specify a function which will always be ran when the resource is created.
+
+```go
+// Fields of the User.
+func (User) Fields() []ent.Field {
+	return []ent.Field{
+		field.Uint64("id").
+			DefaultFunc(func() uint64 { return 89 }),
```
a8m, line 155: Can we use more realistic example here? BTW, if you want to use the [sonyflake](https://github.com/sony/sonyflake) package, you need [handle errors](https://pkg.go.dev/github.com/sony/sonyflake#Sonyflake.NextID), and [schema hooks](https://entgo.io/docs/hooks/#schema-hooks) is preferred here IMO. For example:

```go
// BaseMixin to be shared will all different schemas. 
type BaseMixin struct {
	mixin.Schema
}

func (BaseMixin) Fields() []ent.Field {
	return []ent.Field{
		field.Uint64("id"),
	}
}

func (BaseMixin) Hooks() []ent.Hook {
	return []ent.Hook{
		hook.On(IDHook(), ent.OpCreate),
	}
}

func IDHook() ent.Hook {
	type IDSetter interface {
		SetID(uint64)
	}
	return func(next ent.Mutator) ent.Mutator {
			is, ok := m.(IDSetter)
			if !ok {
				return nil, fmt.Errorf("unexpected mutation %T", m)
			}
			id, err := sf.NextID()
			if err != nil {
				return nil, err
			}
			is.SetID(id)
			return next.Mutate(ctx, m)
		})
	}
}

// User holds the schema definition for the User entity.
type User struct {
	ent.Schema
}

// Mixin of the User.
func (User) Mixin() []ent.Mixin {
	return []ent.Mixin{
		// Embed the BaseMixin in user schema.
		BaseMixin{},
	}
}

// ...
```

entc/load/schema_test.go
```diff
@@ -252,6 +252,10 @@ type WithDefaults struct {
 }
 
 func (WithDefaults) Fields() []ent.Field {
+	idf := field.Int("int_default_func")
+	idf.Descriptor().Default = func() int {
+		return 1e9
+	}
```
a8m, line 258: Why not using the `DefaultFunc` option?

schema/field/numeric.go
```diff
@@ -1474,6 +1537,13 @@ func (b *uint64Builder) Default(i uint64) *uint64Builder {
 	return b
 }
 
+// DefaultFunc sets the function that is applied to set the default value
+// of the field on creation.
+func (b *uint64Builder) DefaultFunc(fn interface{}) *uint64Builder {
+	b.desc.Default = fn
+	return b
+}
+
```
a8m, line 1546: Please add a check for the function signature in the `Descriptor` method as exists in other builders (e.g. see [`stringBuilder`](https://github.com/facebook/ent/blob/master/schema/field/field.go#L297-L302)).

entc/load/schema_test.go
thehowl: added this as a regression test for the actual issue before actually implementing `DefaultFunc` - fixing now

doc/md/schema-fields.md
thehowl: I think it's probably best if we keep the docs agnostic if possible - so if anything I would consider suggesting a dumb ID generator (ie. `time.Now().Unix() << 8 | atomic.AddInt64(&counter, 1) % 256`). Still, I'll point to hooks for users wishing to use sonyflake or any other ID generator which returns an error

a8m: > I think it's probably best if we keep the docs agnostic in the examples if possible - so if anything I would consider suggesting a dumb ID generator (ie. time.Now().Unix() << 8 | atomic.AddInt64(&counter, 1) % 256). 

I agree, just wanted to point it out because that was your use case. 

> Still, I'll point to hooks for users wishing to use sonyflake or any other ID generator which returns an error

Yes, let's add it to the [FAQ page](https://entgo.io/docs/faq/).

schema/field/field_test.go
```diff
@@ -90,6 +90,21 @@ func TestInt(t *testing.T) {
...
+	f1 := func() CustomInt { return 1000 }
+	fd := field.Int("id").DefaultFunc(f1).GoType(CustomInt(0)).Descriptor()
+	assert.NoError(t, fd.Err)
+
+	fd = field.String("id").DefaultFunc(f1).Descriptor()
+	assert.Error(t, fd.Err, "`var _ int = f1()` should fail")
+
+	f2 := func() int { return 1000 }
+	fd = field.String("dir").GoType(CustomInt(0)).DefaultFunc(f2).Descriptor()
+	assert.Error(t, fd.Err, "`var _ customInt = f2()` should fail")
```
a8m, line 105:
```suggestion
	fd = field.Int("id").DefaultFunc(f1).Descriptor()
	assert.Error(t, fd.Err, "`var _ int = f1()` should fail")

	f2 := func() int { return 1000 }
	fd = field.Int("dir").GoType(CustomInt(0)).DefaultFunc(f2).Descriptor()
	assert.Error(t, fd.Err, "`var _ customInt = f2()` should fail")
```

doc/md/schema-fields.md
```diff
@@ -144,6 +144,24 @@ func (Pet) Fields() []ent.Field {
...
+var globalIDCounter int64 = 0
+
+// Fields of the User.
+func (User) Fields() []ent.Field {
+	return []ent.Field{
+		field.Int64("id").
+			DefaultFunc(func() uint64 {
+				// example of a dumb ID generator - use a production-ready alternative instead
+				return time.Now().Unix() << 8 | atomic.AddInt64(&counter, 1) % 256
+			}),
```
a8m, line 160:
```suggestion
		field.Uint64("id").
			DefaultFunc(func() uint64 {
				// example of a dumb ID generator - use a production-ready alternative instead
				return time.Now().Unix() << 8 | atomic.AddInt64(&counter, 1) % 256
			}),
```

a8m (commented):
Minor comments. Please fix them and I'll merge it.

a8m (approved):
Wonderful! Thanks a lot for your contribution @thehowl 😃

PR #1154, opened 2021-01-08
Add locking reads

rubensayshi:
as per https://github.com/facebook/ent/issues/358

This adds `LockForUpdate()` and `LockForShare()` to both `Tx` and `{{Entity}}Query`.

I'm a bit on the fence about adding them to the `Tx`, a lot of sub queries etc. will also have the locking applied, personally I'd only want them on the `{{Entity}}Query`, but in the original feature request people expressed a desire to have it on the `Tx`.  

(still need to `go generate entc/integration/...` but haven't done so yet to keep the diff small for discussion)


dialect/sql/builder.go
```diff
@@ -1796,6 +1797,20 @@ func (s *Selector) Having(p *Predicate) *Selector {
 	return s
 }
 
+// ForUpdate appends the `FOR UPDATE` clause to the `SELECT` statement.
+func (s *Selector) ForUpdate() *Selector {
+	s.distinct = false // combination of DISTINCT and FOR UPDATE is not allowed
```
rubensayshi, line 1802: this is annoying, but you can't combine DISTINCT and FOR UPDATE. 

I'm not sure how to best fix this as we do distinct by default, I don't really like putting this here in `Selector` ...

I think I'd prefer handling this in `func (q *query) selector() (*sql.Selector, error)`? maybe even error if `Unique` is set ... or would it be okay that `LockForUpdate()` cancels out the Unique?

UnRyongPark:
Hi! Is there any additional update on this PR? I've been waiting for this feature 😇

rubensayshi:
hmm I'll rebase and maybe @a8m can take a peek at it :) ?

dialect/sql/builder.go
rubensayshi: @a8m curious what you think about this?

BrentChesny:
Also have a use case for this, so would be great if this could be merged in. 😁

rubensayshi:
@a8m if you can take a peek then I got time to take care of any feedback you may have!

dialect/sql/builder.go
a8m: Hey @rubensayshi, I agree with you here. I think the `*sql.Selector` should be simple. 

Regarding this API - I think it's clean, but we should make it possible to support all options (e.g. `FOR SHARE OF <T> NO WAIT`). 

Throwing here a few ideas:

```go
func (s *Selector) For(lockOpts ...sql.LockOptions) *Selector

func (s *Selector) For(strength sql.LockStrength, lockOpts ...sql.LockOptions) *Selector

func (s *Selector) ForShare(lockOpts ...sql.LockOptions) *Selector

func (s *Selector) ForShare(lockCfg sql.LockConfig) *Selector
```

```go
s.For(sql.Share(), sql.Of(t), sql.NoWait())

s.ForShare(sql.Of(t), sql.NoWait())

s.ForShare(sql.LockConfig{
	Tables: []string{t1, t2},
	NoWait: true,
})
```

rubensayshi:
merged in https://github.com/ent/ent/pull/1697

PR #1160, opened 2021-01-10
docs: fix link to Annotation interface

tmc:
These links were to the Annotation type in schema/field vs the interface in the schema package.


a8m (approved):
Thanks @tmc

PR #1163, opened 2021-01-12
entc/internal: catch syntax error due to conflict marker

BrentChesny:
A syntax error caused by a conflict marker doesn't get caught by the IsBuildError function, causing Ent not to try and recover using the stored schema snapshot. In this PR we add a check for the substring `found '<<'`.


a8m (approved):
Thanks for fixing this @BrentChesny

PR #1173, opened 2021-01-14
schema/field: add Comment value to Descriptor

kerbelp:
First commit for issue: https://github.com/facebook/ent/issues/944


a8m (approved):
Thanks for your contribution @kerbelp. Please let me know if you need any help with the codegen part.

PR #1181, opened 2021-01-17
Allow users to pass alternate schema names for models

marwan-at-work:
This PR exposes the already-merged inner functionality of passing a custom DB name (schema) for each model. 

It adds a new option that can be used as such: 

```golang
ent.Open("mysql", conn, ent.WithAlternateSchemas(dbschema.Schemas{User: "usersdb"})
```

The last todo in this PR is put each model that wants an alternate schema behind its own sql-annotation, but I'll wait on the first review pass before adding that functionality. And integration tests of course.


entc/gen/feature.go
```diff
@@ -43,11 +43,25 @@ var (
...
+	FeatureCustomSchemas = Feature{
+		Name:        "custom-schemas",
+		Stage:       Experimental,
+		Default:     false,
+		Description: "Allows alternate schema names for each ent model. Useful if SQL tables are spread out against multiple databases",
+		cleanup: func(c *Config) error {
+			return os.RemoveAll(filepath.Join(c.Target, "dbschema"))
+		},
+	}
+
```
a8m, line 58: Great, but let's move the name to be prefixed with `"sql"` (`"sql/<Name>"`), and maybe propose other names to it 😃  (e.g. `"sql/dynamicschema"`, `"sql/schemaconfig"`, `"sql/alternate-schema"`, etc).

I like the `"sql/schemaconfig"` one, because as I see it, this feautre-flag exposes a configuration functionality for the generated assets.

entc/gen/graph.go
```diff
@@ -216,6 +216,26 @@ func (g *Graph) addNode(schema *load.Schema) {
...
+				if _, ok := visited[model]; ok {
+					continue
+				}
+				visited[model] = struct{}{}
+				models = append(models, model)
+			}
+		}
+	}
+	return models
+}
```
a8m, line 237: Let's remove this functionality from here and keep it in the template as it has only one usage.

entc/gen/template.go
```diff
@@ -162,6 +162,13 @@ var (
 				return !g.featureEnabled(FeatureSnapshot)
 			},
 		},
+		{
+			Name:   "dbschema",
+			Format: "dbschema/dbschema.go",
```
a8m, line 167:
```suggestion
			Name:   "dbschema",
			Format: "internal/schemaconfig.go",
```

I prefer to not generate a new package if it's not really needed (especially, if the reason is technical).

We can store this info in the `internal` package, and just define `type SchemaConfig = internal.SchemaConfig` to expose it to the users (in client configuration).

a8m (commented):
Great progress on this huge feature @marwan-at-work. Let's talk over Slack how to break it to multiple PRs because there are some parts that I think we can simplify.

dialect/sql/builder.go
```diff
@@ -1531,6 +1535,24 @@ type Selector struct {
...
+
+// Context returns the Selector context or Background
+// if nil.
+func (s *Selector) Context() context.Context {
+	if s.ctx != nil {
+		return s.ctx
+	}
+	return context.Background()
+}
+
```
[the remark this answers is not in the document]
marwan-at-work: For sure, here you go: https://github.com/facebook/ent/pull/1185

dialect/sql/sqlgraph/graph.go
```diff
@@ -271,6 +271,7 @@ func HasNeighborsWith(q *sql.Selector, s *Step, pred func(*sql.Selector)) {
 			Join(to).
 			On(edge.C(pk1), to.C(s.To.Column))
 		matches := builder.Select().From(to)
+		matches.WithContext(q.Context())
```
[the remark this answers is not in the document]
marwan-at-work: @a8m this is needed because `matches` is what actually gets passed to `pred(matches)` right below which are the `Has` and `HasWith` functions that _need_ the Selector to have a context to extract the "schemaconfig" and apply it to edges.

entc/gen/feature.go
marwan-at-work: @a8m sounds good 👍🏼 should the `dbschema` package be instead named `schemaconfig` as well to stay consistent?

entc/gen/graph.go
marwan-at-work: @a8m by "keep it in the template", do you mean have this function be completely defined in the .tmpl files? I don't think Go's templates allow you to set values such as an empty struct to a map to ensure uniqueness. 

Or do you mean to just move it to another file or package?

entc/gen/feature.go
marwan-at-work: @a8m updated this PR by renaming the feature to `sql/schemaconfig`

entc/gen/template.go
marwan-at-work: @a8m updated this PR by removing `dbschema` and putting this struct inside `internal`

PR #1185, opened 2021-01-18
Add context.Context to sql.Selector

marwan-at-work:
This PR adds a context to the sql selector so that we can pass values from the generate code which can be extracted later as dynamic configuration for the query builder


a8m (approved):

a8m:
Wonderful. Thanks @marwan-at-work

PR #1186, opened 2021-01-18
Ensure sqlgraph passes the context to *sql.Selector

marwan-at-work:
This PR ensures that sqlgraph functions that receive a context and create a selector will make sure to pass the context to said selector so that future predicate functions can extract data relevant to them. 

### DIFF
--- a/dialect/sql/sqlgraph/graph_test.go
+++ b/dialect/sql/sqlgraph/graph_test.go
@@ -810,7 +810,7 @@ func TestHasNeighborsWithContext(t *testing.T) {
 	ctx := context.WithValue(context.Background(), key("mykey"), "myval")
 	for _, rel := range [...]Rel{M2M, O2M, O2O} {
 		t.Run(rel.String(), func(t *testing.T) {
-			sel := sql.Dialect("postgres").
+			sel := sql.Dialect(dialect.Postgres).
 				Select("*").
 				From(sql.Table("users")).
 				WithContext(ctx)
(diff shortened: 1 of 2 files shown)


dialect/sql/sqlgraph/graph_test.go
```diff
@@ -805,6 +805,35 @@ WHERE "s1"."users"."id" IN
 	}
 }
 
+func TestHasNeighborsWithContext(t *testing.T) {
+	type key string
+	ctx := context.WithValue(context.Background(), key("mykey"), "myval")
+	for _, rel := range [...]Rel{M2M, O2M, O2O} {
+		t.Run(rel.String(), func(t *testing.T) {
+			sel := sql.Dialect("postgres").
```
a8m, line 813:
```suggestion
			sel := sql.Dialect(dialect.Postgres).
```

dialect/sql/sqlgraph/graph_test.go
```diff
@@ -805,6 +805,35 @@ WHERE "s1"."users"."id" IN
...
+			pred := func(s *sql.Selector) {
+				called = true
+				if got := s.Context().Value(key("mykey")).(string); got != "myval" {
+					t.Fatalf("expected selector context to have %q but got %q", "myval", got)
+				}
+			}
+			HasNeighborsWith(sel, step, pred)
+			if !called {
+				t.Fatal("expected predicate function to be called")
+			}
```
a8m, line 832:
```suggestion
			pred := func(s *sql.Selector) {
				called = true
				got := s.Context().Value(key("mykey")).(string)
				require.Equal(t,  "myval" , got)
			}
			HasNeighborsWith(sel, step, pred)
			require.True(t, called, "expected predicate function to be called")
```

Minor comment - I'm not fan of the `testify/require` package, but we already use it, so let's stay consistent with the existing codebase. Thanks ❤️

a8m (approved):

Issue #1143, opened 2021-01-06
Where(ID) for Edge should be possible without subquery


rubensayshi:
(sorry for bad title, not sure how to better summarize)

- [x] I have searched the [issues](https://github.com/facebook/ent/issues) of this repository and believe that this is not a duplicate.

## Problem

When you want to find all `Pet`s by their `Owner`, giving an `ownerID` you currently have to write a custom predicate or do 2 queries;

```go

client.Pet.Query().Where(pet.HasOwnerWith(owner.UD(ownerID))

// or

client.Owner.Query().Where(owner.UD(ownerID)).QueryPets()

```

But the query could really just be a `SELECT * FROM pets WHERE owner_id = ?`

## Solution

I think a good solution would be to add simple where methods for edges, in this case`WhereOwner()` and `WhereOwnerID()` ?

a8m:
Hey @rubensayshi, yeah, I'm aware about this issue as it was raised before.
Currently, it's possible to solve this by using custom predicate as mentioned [in the doc](https://entgo.io/docs/predicates/#custom-predicates), and even add them to your project with custom templates. 

> I think a good solution would be to add simple where methods for edges, in this caseWhereOwner() and WhereOwnerID() ?

I prefer to solve this optimization in the framework without changing the current API, because there are many predicates on the ID field (In, NotIn, EQ, NEQ, etc). I'll give it another look this week and update.

rubensayshi:
hmm, seems neigh impossible to get it right when `HasOwnerWith(...)` is used, wouldn't it just make more sense to go for the simple solution of adding `pet.WhereOwnerID...()` methods? it will result is more readable code as well than `hasOwnerWith(owner.ID(ownerID))`

rubensayshi:
@a8m been looking at trying to determine if the predicate is a simple op on the foreign key (`owner_id`) but I can only think of 2 ways:

1) in `HasNeighborsWith` evaluate the `pred()` on a temporary `Selector` and try and determine if it's an simple `SELECT WHERE owner_id = ?` (or `IN(?`, etc), but this comes down to string comparing ...  
but this is a ugly and errorprone solution and probably not great for performance either ...

2) refactoring predicates so instead of `func(*sql.Selector)` they're of some interface type which we can then check if it's a foreign key by its type. but this breaks the current API.


I think the best solution is just to use code generating to create `WhereOwnerID...()` methods.

a8m:
[...]

PR #1192, opened 2021-01-19
Add WhereEdgeID... predicates for entities which own the FK

rubensayshi:
as discussed a bit in https://github.com/facebook/ent/issues/1143 already, this adds helpers for `WhereEdgeID...()` predicates to avoid doing a subquery that a simple `.HasEdgeWith(edge.ID(edgeID))` incurs.

as I expressed in #1143, I think dissecting `.HasEdgeWith(edge.ID(edgeID))` and automatically getting rid of the subquery is very hard because it requires either:
1) in `HasNeighborsWith` evaluate the `pred()` on a temporary `Selector` and try and determine if it's an simple `SELECT WHERE owner_id = ?` (or IN(?, etc), but this comes down to string comparing ...
but this is a ugly and errorprone solution and probably not great for performance either ...
2) refactoring predicates so instead of `func(*sql.Selector)` they're of some interface type which we can then check if it's a foreign key by its type. but this breaks the current API.

this PR is a simple alternative to be considered ;)


a8m:
Thanks @rubensayshi for working on this.

Closing, as this was addressed with #1213

PR #1203, opened 2021-01-23
Ensure SchemaConfig is set in sqlQuery's context

marwan-at-work:
Fixes an issue with the current open PR


a8m (approved):

a8m:
Wonderful @marwan-at-work :) Thanks for fixing this

PR #1205, opened 2021-01-24
load/schema: added comment to generated schema

kerbelp:


entc/integration/config/ent/schema/user.go
```diff
@@ -40,11 +40,13 @@ func (User) Fields() []ent.Field {
 			Annotations(entsql.Annotation{
 				Incremental: &incrementalDisabled,
 			}),
+		field.Int("age"),
```
a8m, line 43:
```suggestion
```

This one is not needed.

entc/integration/config/ent/user.go
```diff
@@ -19,7 +19,10 @@ type User struct {
 	config `json:"-"`
 	// ID of the ent.
 	ID int `json:"id,omitempty"`
-	// Name holds the value of the "name" field.
+	// Age holds the value of the "age" field.
+	Age int `json:"age,omitempty"`
+	// Comment line1
+	//  Comment line2
 	Name string `json:"name,omitempty"`
```
a8m, line 26: Make sure to test the generated comment in your test.

You can use the `ast`/`parser` packages as follows:

```go
f, err := parser.ParseFile(token.NewFileSet(), "ent/user.go", nil, parser.ParseComments)
ast.Inspect(f, func(n ast.Node) bool {
	if f, ok := n.(*ast.Field); ok && len(f.Names) > 0 && f.Names[0].Name == "Name" {
		require.Equal(t, descriptorComment, f.Doc.Text())
		return false
	}
	return true
})
``` 

It's a bit extreme. But this will make sure the behaviour won't change in the future.

entc/load/schema_test.go
```diff
@@ -73,7 +73,7 @@ func (User) Annotations() []schema.Annotation {
 
 func (User) Fields() []ent.Field {
 	return []ent.Field{
-		field.Int("age"),
+		field.Int("age").Comment("some comment"),
```
a8m, line 76:
```suggestion
		field.Int("age").
			Comment("some comment"),
```

a8m:
Thanks for adding this @kerbelp. Added a few minor comments.

entc/load/schema_test.go
```diff
@@ -210,6 +210,9 @@ func TestMarshalSchema(t *testing.T) {
 		require.Equal(t, []string{"parent"}, schema.Indexes[1].Edges)
 		require.Equal(t, "user_parent_name", schema.Indexes[1].StorageKey)
 		require.True(t, schema.Indexes[1].Unique)
+
+		require.Equal(t, "some comment", schema.Fields[0].Comment)
+		require.Equal(t, "", schema.Fields[1].Comment)
```
a8m, line 215:
```suggestion
		require.Empty(t, schema.Fields[1].Comment)
```

a8m (approved):
👍 

Looks good

PR #1207, opened 2021-01-25
Add flag to prefix subpackage name for entities

rubensayshi:
Alternative to https://github.com/facebook/ent/pull/1191

So personally I have 1 major gripe with ent and that is that the package names of the entities which hold the consts and the predicates conflicts a lot with variables names we'd ideally be using.  
ie. I'd ideally assign a `User` entity to `user` and not have to name it `usr` or `u` ...

I'd much prefer to prefix the ent subpackages with something, like `ent...` -> `entuser`, `db...` -> `dbuser`.

And I wouldn't be surprised if more people feel this way ... (my other team members feel the same for example)

I was trying to think of a solution, considering renaming the entities etc. but I think the cleanest solution is to either:

1) override the package name through a annotation
2) set a global prefix or suffix for the package names, probably as a cli arg

so this PR implements (2) as after implementing (1) in https://github.com/facebook/ent/pull/1191 I figured this would probably be nicer.


rubensayshi:
not sure why the generate test is failing, I can `go generate` all I want but it's not showing a diff :/

PR #1212, opened 2021-01-27
[bug] using .Total() in builder.Join results in wrong count

rubensayshi:
Opening this PR to demonstrate something which I think is a bug.  
It also includes a fix, but I can't say with 100% certainty this is the right fix.

I was using the `sql.Expr` to construct some custom query stuff and noticed that there was something wrong with the parameter numbering in postgres.

there's 2 tests to demonstrate the issue, both fail when the `b.total = st.Total()` is left in place.

the problem seems to be that without the `b.total = st.Total()` we loose the previously set `.SetTotal`, but with it it will reset it when we're doing nested predicates.  
I think my fix will work in both cases, and all tests pass ...

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -2221,10 +2221,7 @@ func (b *Builder) join(qs []Querier, sep string) *Builder {
 		query, args := q.Query()
 		b.WriteString(query)
 		b.args = append(b.args, args...)
-		b.total = len(b.args)
-		if ok {
-			b.total = st.Total()
-		}
+		b.total += len(args)
 	}
 	return b
 }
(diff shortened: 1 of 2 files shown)


rubensayshi:
I'm aware `b.Join(Expr("name = $1", "pedro"))` could be done as `b.Ident("name").WriteOp(OpEQ).Arg("pedro")` but for complex situations `Expr()` could be preferred and that's not working well atm.

in my real code I'm actually doing something like:

```go
total := b.Total() + 1
b.Join(Expr(fmt.Sprintf("name = $%d", total), "pedro"))
```

which is how I noticed the total was off

dialect/sql/builder.go
```diff
@@ -2221,10 +2221,7 @@ func (b *Builder) join(qs []Querier, sep string) *Builder {
 		query, args := q.Query()
 		b.WriteString(query)
 		b.args = append(b.args, args...)
-		b.total = len(b.args)
-		if ok {
-			b.total = st.Total()
-		}
```
a8m, line 2227: 👍 

I think this was added when the predicate implementation was different.

a8m (approved):
Thanks for your contribution @rubensayshi

PR #1215, opened 2021-01-29
Add documentation for the sql/schemaconfig feature

marwan-at-work:
Adds instructions on how to provide custom schemas to ent models.


a8m (approved):
🥳 

Thanks @marwan-at-work. Next step is the application-level sharding.

PR #1218, opened 2021-01-30
Otherfield

a8m:


schema/field/field.go
```diff
@@ -895,6 +916,131 @@ func (b *uuidBuilder) Descriptor() *Descriptor {
...
+
+// Annotations adds a list of annotations to the field object to be used by
+// codegen extensions.
+//
+//	field.Other("link", &Link{}).
+//		SchemaType(map[string]string{
+//			dialect.MySQL:    "text",
+//			dialect.Postgres: "varchar",
+//		}).
+//		Annotations(entgql.Config{
```
enjoylife, line 1027: I pulled down the latest ent-contrib and I'm not seeing the same annotation api as this doc comment. Instead I see something like:
```
Annotations(
				entgql.OrderField("LINK_FIELD"),
			).
```

a8m: You're right. This was change added by https://github.com/facebook/ent/pull/1145 and I missed this.

Thanks for adding your review, it's really appreciated 🙏

PR #1229, opened 2021-02-03
doc: fix github link

uta-mori:
I was happily reading ent documentation site and found that the link to github was broken

`https://entgo.io/ent/tree/master/examples/ `-> `https://github.com/facebook/ent/tree/master/examples/`


a8m (approved):
Thanks for fixing this @uta-mori

PR #1234, opened 2021-02-03
Make custom field comment render as additional line(s)

zzwx:
The intent is to keep originally generated comment that has the DB field name in it, and simply append the custom comments right below. Also ID field seemed to be ignoring the comment completely.

For those, who'd want to produce paragraph separation that simply means adding `\n` in front of the comment.


a8m (approved):
That makes sense. Thanks for addressing this @zzwx 

One small comment before I land this - you need to change [this test](https://github.com/facebook/ent/blob/master/entc/integration/config/config_test.go#L60-L66) to follow the new format (maybe use `Contains` instead of `Equals`).

zzwx:
My bad, missed that test, also integration tests fail...

a8m:
Thanks for your contribution @zzwx. It's appreciated!

zzwx:
Thank you for walking me through the process!

PR #1239, opened 2021-02-05
entc: add extra name check for private identifiers

### LINKED ISSUE #1225: The generated name is the same as the package name
I created a table using `ent init Config`,
after i generate the code use `ent generate ./ent/schema`.
The program was compiled with error

error code:
```
ent\client.go:13:2: config redeclared in this block
	previous declaration at ent\config.go:14:6
```

- [x ] The issue is present in the latest release.
- [x ] I have searched the [issues](https://github.com/facebook/ent/issues) of this repository and believe that this is not a duplicate.

## Steps to Reproduce 🕹

Steps:

1. ent init Config
2. ent generate ./ent/schema
3. coding a main function
4. build and run it

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.15.6  |
| Ent         | 0.5.4  |
| Database    | sqlite3|
| Driver      | https://github.com/mattn/go-sqlite3        |

napei:
Fixes #1225 

Supplementary to #1110 

Adds checks for schema names against private identifiers when running codegen and using `ent init`

image
image

### DIFF
--- a/entc/gen/type.go
+++ b/entc/gen/type.go
@@ -694,6 +694,9 @@ func ValidSchemaName(name string) error {
 	if _, ok := globalIdent[name]; ok {
 		return fmt.Errorf("schema name conflicts with ent predeclared identifier %q", name)
 	}
+	if _, ok := privateField[pkg]; ok {
+		return fmt.Errorf("schema name conflicts with ent builder fields %q", pkg)
+	}
 	return nil
 }
 


alkanna:
I was about to make a PR for this :)
While we're on this subject, I noticed we are able to create entities such as `Edge`, would that not also conflict with existing `ent` types ? @napei

napei:
> I noticed we are able to create entities such as `Edge`, wouldn't that also conflict with existing `ent` types ?

Will have to get @a8m 's opinion on this, but I would say it's probably not a conflict and more of a semantic challenge?

It's not a redeclaration and it still compiles without breaking generated code (I think). 

Because you have `ent.Edge` as an interface which is from `entgo.io/ent` and then you would also have `ent.Edge{}` from `path/to/local/ent` which is the most common import path when using ent in client code anyway.

alkanna:
Yes, you're right, it's just an ambiguity then. It compiles fine and does not conflict at runtime.

a8m (approved):
Thanks for fixing this @napei 🙏

a8m:
> > I noticed we are able to create entities such as `Edge`, wouldn't that also conflict with existing `ent` types ?
> 
> Will have to get @a8m 's opinion on this, but I would say it's probably not a conflict and more of a semantic ambiguity?
> 
> It's not a redeclaration and it still compiles without breaking generated code (I think).
> 
> Because you have `ent.Edge` as an interface which is from `entgo.io/ent` and then you would also have `ent.Edge{}` from `path/to/local/ent` which is the most common import path when using ent in client code anyway. By all means test it and see if having an entity called Edge breaks anything.

If it's compiled I wouldn't reject it, although  I don't really like it.

napei:
> > > I noticed we are able to create entities such as `Edge`, wouldn't that also conflict with existing `ent` types ?
> > 
> > 
> > Will have to get @a8m 's opinion on this, but I would say it's probably not a conflict and more of a semantic ambiguity?
> > It's not a redeclaration and it still compiles without breaking generated code (I think).
> > Because you have `ent.Edge` as an interface which is from `entgo.io/ent` and then you would also have `ent.Edge{}` from `path/to/local/ent` which is the most common import path when using ent in client code anyway. By all means test it and see if having an entity called Edge breaks anything.
> 
> If it's compiled I wouldn't reject it, although I don't really like it.

@a8m On this topic, is `Order` as a schema name technically a conflict with `order`. Just came across this in my own repo and there's no compiler issues, but just wondering why it's in the private fields array.

a8m:
`Order` is a valid schema name. I just added a patch to fix this (https://github.com/facebook/ent/pull/1244). Thanks

PR #1249, opened 2021-02-08
entc/gen: added ability to add comment to schema with external template

kerbelp:
This change will allow using external tools (such as swagger) to add comments to the schema.
For example, by adding a Swagger annotation to the schema we can add `// swagger:model` comment.


a8m (approved):
Looks good @kerbelp, thanks for adding this.

kerbelp:
This change will allow using external tools (such as swagger) to add comments to the schema.
For example, by adding a Swagger annotation to the schema we can add `// swagger:model` comment.

Issue #1242, opened 2021-02-06
fix entgo.io path resolution


napei:
Slight annoyance :)

`entgo.io/docs` should 301 to `entgo.io/docs/getting-started`. Currently it just errors out. 

```xml
<Error>
<Code>NoSuchKey</Code>
<Message>The specified key does not exist.</Message>
<Key>docs/index.html</Key>
<RequestId>9CEEF6127893C701</RequestId>
<HostId>tyiH3eHjb/QsYbNqSTAmSeYvR/4b1KftsvOrf2pslHLM1XZ6kdjO3AE1fPP3tvvNAE6vRkWmQuA=</HostId>
</Error>
```

PR #1257, opened 2021-02-16
doc: fix Github Link entgo.io -> github.com/ent

uta-mori:
#1242 

Fixed the link to github.


a8m (approved):
Thanks for fixing this @uta-mori 🙏

PR #1267, opened 2021-02-24
Remove whitespace from markdown files

justinpage:
Remove any additional trailing whitespaces from markdown files.


a8m (approved):
Thanks for fixing this @justinpage

PR #509, opened 2020-05-28
entc/gen/template: Wrap error returned by field validator

petersondmg:
Changes proposed in this pull request:

- Use verb `%w` on field validation call to `fmt.Errorf`, so that the error type can be checked using `errors.Is()` function.

fixes: #508

Issue #508, opened 2020-05-28
Error returned from field validator can't be easily checked


petersondmg:
As the error returned from validator is a result of a `fmt.Errorf("...: %v")`, it's hard to check whether the error is due to a certain validator.

**Example**:
```go
type Group struct {
    ent.Schema
}

var (
    LowerCaseErr = errors.New("group name must begin with uppercase")
)

func (Group) Fields() []ent.Field {
    return []ent.Field{
        field.String("name").
            Validate(func(s string) error {
                if strings.ToLower(s) == s {
                    return LowerCaseErr
                }
                return nil
            }),
    }
}

...

group, err := client.Group.Create().
    .SetName("cats")
    .Save(ctx)

// won't work
if errors.Is(err, LowerCaseErr) {
    // appropriate handle this kind of error
}

```

PR #1275, opened 2021-02-25
all: use %w instead of %v to wrap errors

mgabeler-lee-6rs:
... so that they can be unwrapped & the root causes inspected with `errors.Unwrap`, `errors.Is`, and `errors.As`.

Conceptually a follow-up to #508, #509, but trying to be systematic over the whole repo.

The particular use case that motivated this for me: I have code in my app that tries to detect & do Smart Things if database interactions failed because of context cancellation / deadline exceeded, as opposed to other (more worrisome) reasons. But if the errors happen in certain places within ent, that logic fails, as things like `errors.Is(err, context.Canceled)` return `false`, even when that _was_ the real underlying error.

Another use case (though one that is so far theoretical for me) is wanting to access the detailed underlying DB error type to get extended information out for logging/diagnostics/metrics. For example, `pgx` has a rich error type, but it's inaccessible if errors from ent don't consistently use wrapping.

----

Most of this was done with "supervised" search & replace. One place I did a bit of a more manual edit of note is the handling of errors during rollback recovering from another error. Several (all?) of these had comments _claiming_ the original error was wrapped, which is not what the implementation actually did, so I chose to make the implementation match the comment, since the documented behavior seems proper.

To hopefully ease review, I separated this into two commits, the first with the "edited" files, the second those updated by application of `go generate`

PS: I'm on gophers slack as `fastcat`, lurking in #ent, feel free to ping me there too for questions / follow-up

- [x] Tests pass locally
- [x] Lint passes locally

----

NB: I did get the tests to work locally, but I ran into some challenges first with just finding instructions on how to run tests & linting, and second that the instructions I did find for running the tests didn't quite work -- the docker compose file seems to have some features that, at best, only work with certain Docker-for-Windows configurations. They certainly don't seem to work on Linux, where I had to comment out the `platform:` directives.


doc/md/getting-started.md
```diff
@@ -141,12 +141,12 @@ import (
 func main() {
```
a8m, line 141: The change is not needed in this file, because we format and fail.

doc/md/privacy.md
```diff
@@ -102,7 +102,7 @@ func main() {
         Templates: entgql.AllTemplates,
```
a8m, line 102: ditto

a8m (approved):
Thanks @mgabeler-lee-6rs for contributing this chance.
Please fix the minor comments and I'll merge this.

doc/md/getting-started.md
mgabeler-lee-6rs: 👍 will fix this and update for changes on `master`

mgabeler-lee-6rs:
Addressed review items. Since things are a bit tricky to review after a rebase, here's the commit I made locally to address the review, before squashing & rebasing it: https://github.com/mgabeler-lee-6rs/ent/commit/c9722d34

Re-ran lint & test after the rebase, still passes locally

a8m:
Thanks @mgabeler-lee-6rs for addressing this. It's appreciated.

mgabeler-lee-6rs:
You're welcome, and thank you!

PR #1300, opened 2021-03-03
Fix //go:generate code generation to match go1.16

uta-mori:
Starting with go1.16, the current generate.go statement will result in an error.

To make it behave the same as before, we need to change the description to the following.

```diff
- //go:generate go run entgo.io/ent/cmd/ent generate ./schema
+ //go:generate go install entgo.io/ent/cmd/ent@latest
+ //go:generate ent generate ./schema
```


rotemtam:
Hey @uta-mori,
Can you please provide more details? What error are you seeing? How is it relevant to changes in go1.16? 
Thanks

uta-mori:
@rotemtam 
ok. I've added the error and the changes in go1.16.

rotemtam:
Thanks for updating the issue @uta-mori.
In general, it is less advised to use global installs, since some teams have multiple ent projects, and don't all use the same version. SInce the gen tool must be the same version as the ent library you are using, this can raise many day to day issues. 

I've tried to reproduce your issue with `go get` and it seems like codegen works:
```console
➜  go116 go mod init repro
go: creating new go.mod: module repro
➜  go116 go get entgo.io/ent/cmd/ent@latest
go get: added entgo.io/ent v0.6.0
➜  go116 go run entgo.io/ent/cmd/ent init User
➜  go116 go generate ./...
➜  go116 tree ./ent
./ent
├── client.go
├── config.go
├── context.go
├── ent.go
├── enttest
│   └── enttest.go
├── generate.go
├── hook
│   └── hook.go
├── migrate
│   ├── migrate.go
│   └── schema.go
├── mutation.go
├── predicate
│   └── predicate.go
├── runtime
│   └── runtime.go
├── runtime.go
├── schema
│   └── user.go
├── tx.go
├── user
│   ├── user.go
│   └── where.go
├── user.go
├── user_create.go
├── user_delete.go
├── user_query.go
└── user_update.go

7 directories, 22 files
```

WDYT?

rotemtam:
Re-reading your issue description, I'm able to reproduce  the issue, even without global installed binaries. Running `go mod tidy` will cause the error you described.

This is related to the discussion in https://github.com/golang/go/issues/43653. 

To fix the issue, we can add this import to the `generate.go` file:

```go
import _ "entgo.io/ent/cmd/ent"
```

This makes the package explicitly depend on the ent command line and its transitive dependencies, in which case `go mod tidy` will not remove them.

@uta-mori , would you like to contribute a fix to change the generated code in this spirit?

a8m:
Thanks for your contribution @uta-mori 🙏 

Any update on this?

uta-mori:
thank you

added `import _ "entgo.io/ent/cmd/ent"` to `generate.go`

---

`go get` will default to behavior with the `-d` flag in the future.
This is a behavior that changes only `go.mod`.

> https://golang.org/cmd/go/#hdr-Add_dependencies_to_current_module_and_install_them
>The -d flag instructs get not to build or install packages. get will only update go.mod and download source code needed to build packages.
>
>Building and installing packages with get is deprecated. In a future release, the -d flag will be enabled by default, and ‘go get’ will be only be used to adjust dependencies of the current module. To install a package using dependencies from the current module, use ‘go install’. To install a package ignoring the current module, use ‘go install’ with an @version suffix like “@latest” after each argument.

marwan-at-work:
@uta-mori @a8m why not instead of changing:

```
//go:generate go run entgo.io/ent/cmd/ent generate ./schema
```

To

```
+ //go:generate go install entgo.io/ent/cmd/ent@latest
+ //go:generate ent generate ./schema
```

Which would _only_ work in Go1.16 and fail for users who have not upgraded yet. Why don't we do this: 

```
//go:generate go run -mod=mod entgo.io/ent/cmd/ent generate ./schema
```

This would work for both 1.16 and previous Go versions _and_ more importantly: it would ensure the version compatibility between the library `ent` and the cmd line `ent`.

uta-mori:
@marwan-at-work

a8m (approved):
Wonderful! Thanks for fixing this @uta-mori.

PR #1304, opened 2021-03-04
Relax the Comparable result for TypeOther

### LINKED ISSUE #1303: Unable to use field.Other and entgql.OrderField annotation together
With the introduction of the `OtherField` type in #1218, it opens up custom types and works well, but doesn't yet compose well with the entgql annotations. 

## Current Behavior 😯
Unable to use `field.Other` and `entgql.OrderField` annotation, due to the `Comparable` method not respecting `TypeOther`.  Digging into the codebase shows that the annotation, [checks](https://github.com/ent/contrib/blob/6623819401500db45747a2419172963217cef619/entgql/template/pagination.tmpl#L217) if the type is comparable. This [method](https://github.com/ent/ent/blob/3ee6621194709db97e53f8ee6888593e156acc5e/schema/field/type.go#L126) doesn't have a case for handling these custom types. 

What happens is the codegen fails

```
running ent codegen: execute template "pagination": template: template/pagination.tmpl:218:6: executing "pagination" at
 <fail (printf "annotated field %s.%s must be comparable" $node.Name $f.Name)>: error calling fail: annotated field Thing.custom_created_time must be comparable
exit status 1
```
## Expected Behavior 🤔
The other codegen completes when using `field.Other(...).Annotations(entgql.OrderField(...))` . Perhaps it requires an interface the `otherType` neees to satisfy but nonetheless I would hope we could combine these two features, `field.Other` and `entgql.OrderField`.

## Steps to Reproduce 🕹

Steps:

1. Use an other field with the order field   annotation: `field.Other("custom_field", pkg.CustomType).  Annotations(entgql.OrderField("CUSTOM_FIELD"))`
2. run go generate

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.15.5  |
| Ent         | 0.6   |
| Database    | Postgres   |

enjoylife:
Fixes #1303

This method is currently only used in the contrib repo `entgql/template/pagination.tmpl`, so I'm pretty sure this wont affect much. I tried to figure out if there was a larger reason for this to exist, but the original PR, #576,  didn't contain much.


schema/field/type.go
```diff
@@ -127,6 +127,10 @@ func (t TypeInfo) Comparable() bool {
 	switch t.Type {
 	case TypeBool, TypeTime, TypeUUID, TypeEnum, TypeString:
 		return true
+	case TypeOther:
+		// Assumes the author of the type knows what they are doing and
+		// sets a reasonable schema type for the dialect used.
+		return true
```
a8m, line 133: Maybe we can check if the [`reflect.Type`](https://golang.org/pkg/reflect/#Type) is comparable and store this information on the [schema/field.TypeInfo](https://pkg.go.dev/entgo.io/ent@v0.6.0/schema/field#TypeInfo)? See the `Type.Comparable` doc. 

There are some cases where the Go type is not comparable, but the database type is comparable so maybe in this case we can check if the type implements an interface that we provide?

Can you share what type do you store on `field.Other`?

a8m (commented):
Thanks for your contribution @enjoylife. Please see my comment.

schema/field/type.go
enjoylife: I too thought briefly on using an interface, but creating another Go interface is independent to the output of `Value() (driver.Value, error)` which is what really matters at the dialect/db level.

As for the `field.Other`, I'm usually embedding/wrapping `time.Time`, or custom structs for holding tree-like data, marshaled out to the db as strings, e.g. [ltree](https://www.postgresql.org/docs/current/ltree.html#id-1.11.7.30.8) which when ordered, is returned as a tree traversal.

a8m: 👍 

Thanks and sorry for the delay.

a8m (approved):

a8m:
PR was added for now, but in the future, we should think about accepting an interface that allows "other types" to tell if they are comparable on the database side.

a8m:
Thanks @enjoylife 🙏

Issue #1310, opened 2021-03-07
Catch FK constraint errors using ent.IsConstraintError


rotemtam:
- [X] I have searched the [issues](https://github.com/facebook/ent/issues) of this repository and believe that this is not a duplicate.

## Summary 💡
The error returned when trying to save an entity with an invalid FK isn't identified by `ent.IsConstraintError`.

## Motivation 🔦
When writing entities with edges to others, it is possible to reference those entities using an ID setter (i.e `SetAuthorID`). If that ID points to a non-existing record, an error is returned, for example:
```
insert node to table "blog_posts": FOREIGN KEY constraint failed
```
It would be very useful for clients using this to be able to identify such cases using the existing `ent.IsConstraintError` helper function.

PR #1316, opened 2021-03-09
initial address for #1310 - allow the user to check if an error is a …

rotemtam:
…constraint error and if it is related to uniqueness or FK


dialect/sql/sqlgraph/errors.go
```diff
@@ -0,0 +1,44 @@
...
+func IsConstraintError(err error) bool {
+	return IsUniquenessConstraintError(err) || IsForeignKeyConstraintError(err)
+}
+
+// IsUniquenessConstraintError returns true if the error resulted from a DB uniqueness constraint error
+func IsUniquenessConstraintError(err error) bool {
+	uniquenessErrors := map[string]string{
+		"mysql":    "Error 1062",
+		"postgres": "violates unique constraint",
+		"sqlite":   "UNIQUE constraint failed",
```
a8m, line 21: Use `dialect.MySQL`, `dialect.PostgreSQL`, etc instead of these strings. Why do we need this, and not just `[3]string{...}`?

dialect/sql/sqlgraph/errors.go
```diff
@@ -0,0 +1,44 @@
...
+package sqlgraph
+
+import (
+	"strings"
+)
+
+// IsConstraintError returns true if the error resulted from a DB constraint error
+func IsConstraintError(err error) bool {
+	return IsUniquenessConstraintError(err) || IsForeignKeyConstraintError(err)
+}
```
a8m, line 14:
```suggestion
// IsConstraintError returns true if the error resulted from a DB constraint error.
func IsConstraintError(err error) bool {
	var e *ConstraintError
	return errors.As(err, &e) || IsUniquenessConstraintError(err) || IsForeignKeyConstraintError(err)
}
```

It does make sense to return true is the error is type `ConstraintError` as well. Also, please add more details regarding the possible constraint errors.

a8m (commented):
Looks good @rotemtam. Thanks for adding this. I've added minor suggestions.

Please use the following commit format when it's possible, it makes my life easier when I'm releasing a new version.

```
<package>: <message>
```

For example
```
dialect/sql/sqlgraph: add constraint error checks

Fixed #ISSUE-NUMBER
```

dialect/sql/sqlgraph/errors.go
rotemtam: we don't, i'm removing

entc/integration/integration_test.go
```diff
@@ -1352,6 +1354,20 @@ func CreateBulk(t *testing.T, client *ent.Client) {
 	require.False(t, pets[2].QueryOwner().ExistX(ctx))
 }
 
+func ConstraintChecks(t *testing.T, client *ent.Client) {
```
a8m, line 1357: Let's remove them after it's added to templates (it should be in sqlgraph).

a8m (approved):

PR #1319, opened 2021-03-09
entc/gen/template/dialect/sql: cleaning up isSQLConstraintError

### LINKED ISSUE #1310: Catch FK constraint errors using ent.IsConstraintError

rotemtam:
cleaning up isSQLConstraintError in generated ent.go to use new sqlgraph IsConstraintError checks.

Adding IsFK and IsUniqueness methods to the generated ent.ConstraintError (on SQL dialects) struct to allow users to glean more information about the specific constraint violation.

Fixes #1310


a8m (approved):

a8m (approved):
Thanks for adding this @rotemtam

PR #1325, opened 2021-03-12
Implement Upserts

### LINKED ISSUE #139: Feature Request: Upsert

ivanvanderbyl:
Resolves #139 

This PR implements upserts using native database support for upserts (`ON CONFLICT DO UPDATE …`) while supporting 3 modes of conflict handling in both single and bulk inserts.

**Example usage:**

```go
// user.email has a Unique() constraint

client.User.Create().
	SetEmail("roger-test@entgo.io").
	SetName("Test User").
	OnConflict(user.FieldEmail).
	SaveX(ctx)

client.User.Create().
	SetEmail("alex-test@entgo.io").
	SetName("Test User").
	OnConflict(user.FieldEmail).
	SaveX(ctx)

users, err := client.User.CreateBulk(
	client.User.Create().SetEmail("roger-test@entgo.io").SetName("Roger"),
	client.User.Create().SetEmail("alex-test@entgo.io").SetName("Alex"),
).
	OnConflict(user.FieldEmail).
	Save(ctx)

// alex-test will be updated to name=Alex
// roger-test will be updated to name=Roger
```

**Conflict handling:**

_Not currently defined in DSL, defaults to using new values_

1. **Do nothing:** This is supported by doing an update on itself in order to get the ID. This might not be ideal for your exact setup, but it seems like an acceptable tradeoff.
2. **Resolve with new values:** sets all fields to the new values provides in the INSERT clause.
3. **Resolve with alternate values:** Sets all fields to alternative values (probably can't support this for bulk updates)

**Things to note:**

- DSL isn't final, needs support for specifying which mode of conflict handling to use
- Bulk inserts that specify alt values will write these values to all records, as it's not possible in SQL to specify a per row upsert value other than the values used in the INSERT clause. Hence the default behaviour is to use these values.
- ~DSL needs an extra method on `CreateBulk` to specify conflict handling for ALL records, however, it's currently implemented to use the conflict handling of the FIRST record.~

Tested on:
- [x] Postgres ✅ 
- [x] MySQL ✅ 
- [x] SQLite ✅ 

TODO:

- [x] Validate that conflict fields have a unique constraint
- [ ] Define conflict resolution mode in DSL
- [ ] Implement builder for alternate column updates that set column=value
- [ ] Figure out how to handle Edge Foreign Key Constraints
- [ ] Show warning if OnConflict is used on individual inserts when doing a bulk insert. This has no effect.


ivanvanderbyl:
@a8m I'd love to get your thoughts on this approach?

ivanvanderbyl:
After thinking about this overnight, I think the DSL needs to support different types of conflict resolution:

```go
// Single inserts
client.User.Create().SetEmail("alex-test@entgo.io").SetUpdateCount(2).
	// Do a custom update
	OnConflict(user.FieldEmail).Update().SetEmail("another-alex@entgo.io").
	Save(ctx)
	
client.User.Create().SetEmail("alex-test@entgo.io").SetUpdateCount(2).
	// Writes each field to itself, forcing an update of the record and returning the ID
	OnConflict(user.FieldEmail).Ignore().
	Save(ctx) 
	
client.User.Create().SetEmail("alex-test@entgo.io").SetUpdateCount(2).
	// Default action, optional. Same as `OnConflict(user.FieldEmail).Save(ctx)`
	OnConflict(user.FieldEmail).UpdateWithNewValues().
	Save(ctx)
```

ivanvanderbyl:
... Another idea:

Support specifying the resolution action per column:

```go
users, err := client.User.CreateBulk(
	client.User.Create().SetEmail("roger-test@entgo.io").SetUpdateCount(2),
	client.User.Create().SetEmail("alex-test@entgo.io").SetUpdateCount(3),
).OnConflict(user.FieldEmail).SetUpdateCount(ent.IncomingValue).SetName("New Name Value")
```

It's also worth noting that specifying columns has no meaning for MySQL, but it is used internally to trigger the `ON DUPLICATE KEY..` logic.

a8m:
Hey @ivanvanderbyl and thanks a lot for starting this effort. It's really appreciated 🙏 

Are you on [Ent's Slack](https://app.slack.com/client/T029RQSE6/C01FMSQDT53)? If not, can you join and ping me there - we can discuss some of these ideas there.

Also, when we're aligned on the design/API, we'll need to break this into multiple PRs - it will be easier to review it this way.

sneakywombat:
hey thanks for doing this work. I was wondering about the syntax though - it would seem more natural to have a top level action like Upsert() than using Create() with additional methods chained later.  As a user, the code becomes much more natural to write.  If a new Upsert method was added, existing Create calls could continue to work as they do now.  Introducing a new one i think gives us a number of benefits.

* new upsert behavior is restricted to this new method, thus easier to review, iterate, land, etc.
* it provides a clear distinction between creating something first and then doing an updated vs trying to put these in one operation, which under the hood are actually a Query() and then Update().  Upsert then becomes a feature of the ent api that makes managing upserts more naturally.
* Create continues to behave as it does now without any additional side-effects or consideration to existing code.  New upsert code is then written to the behavior of upsert itself.  EG, upsert could fail if it needed to create an ent and didn't have enough required fields provided, just like create would.
* Upsert could use Save(ctx) and SaveOnly(ctx) or something like that.  The former, creating or updating all records it's able to following constraints and the latter returning an error if the where clause returned more than one ent to create or update.

I propose that perhaps this pattern may be more clear:

    client.User.Upsert().
        Where(user.NameEQ("Test User").
        SetAge(35).
        SetNickname("fooman").
        SaveOnly(ctx). // SaveOne(ctx) ??

Let's expand this to a full example:

```

type User struct {
    ent.Schema
}

    return []ent.Field{
        field.Int("age"),
        field.String("name"),
        field.String("nickname").
            Unique(),
    }
}

    return []ent.Edge{
        edge.To("groups", Group.Type),
        edge.To("friends", User.Type),
    }
}

func (User) Index() []ent.Index {
    return []ent.Index{
        index.Fields("age", "name").
            Unique(),
    }
}
```

This is taken from the tutorial, but consider a query such as one that uses the Name index field as a starting root...

```
    client.User.Upsert().
        Where(user.NameEQ("Test User").
        SetAge(35).
        SetNickname("fooman").
        Save(ctx). // SaveOnly, SaveOne ?? 

```

Using something like Upsert lets the programmer know, hey, i need to be able to query for something to Upsert.  If I don't have a way to do that sanely, then probably should do a Query() first to get that stuff.   I could add predicates in the where clause that match any field, but ideally, and realistically, we'd be starting with an indexed field or primary key, etc.  If multiple matches where there, Save would work, but, perhaps SaveOnly(ctx) (or SaveOne(ctx)) would return an error if more than one row was selected in from the where clause, further ensuring the desired behavior.  

I can appreciate the syntax that you have here already in the Create, providing required fields per the schema, but i personally feel that most upsert cases are indeed shortcuts to updating existing records where not all of the required data may be available without an additional query.  For really simple schemas, it's easy to provide the fields, but for complicated ones, it becomes increasingly difficult without bailing to a full create, or ensuring that all required fields are provided.

Another additional advantage to do this would be to provide a top level Set(e entType) type function.

```
    client.User.Upsert().
        Where(user.NameEQ("Test User").
        Set(func (u user.User) {
           // complicated work here, you could even use common table expressions if you had such complexity to run.
          if (externalVariable == SET_AGE) {
            u.SetAge(AGE_VALUE_FROM_SOMETHING)
          }
          ... repeat this as needed.  You could do RPCs here to get required data, etc.
              Also for bulk upserts, `u` would be each row selected from the where clause on each loop.
        }).
        Save(ctx)
```

I apologize if this is the current state of the api already, or at least proposed, but I didn't see a discussion like this so I wanted to ask.  thanks again!

timoha:
> Figure out how to handle Edge Foreign Key Constraints

Hello, curious if there's some workaround for this case today when doing CreateBulk with OnConlict columns in Posgtes driver?

PR #1331, opened 2021-03-14
doc/md/getting-started add twitter and newsletter links

rotemtam:


a8m:
This was added in docusarous v2. closing

image

PR #1345, opened 2021-03-17
always prefix receivers with _

evie404:
I have a model named `OS`, which generates package name `os`. Everything works ok until dealing with select query objects, which generates `os` as the receiver name, conflicting with the package name. I made the fix by always prefixing receivers with _, which might be too broad. Feedbacks welcomed.

### DIFF
--- a/entc/gen/func.go
+++ b/entc/gen/func.go
@@ -8,7 +8,6 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"go/token"
 	"path/filepath"
 	"reflect"
 	"sort"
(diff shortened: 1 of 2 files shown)


a8m (commented):
Hey @rickypai and thanks for your contribution, it's appreciated. 

I have another suggestion for this change (which I wanted to do for a long time) - what do you think about using constant receiver names for all query builders. For example:  `q` for `<T>Query` (e.g. `UserQuery`),  `c` for `<T>Create`, `u` for `<T>Update`, and `d` for `<T>Delete`.

It's a bigger change, but you can do it in multiple PRs. 


Let me know what do you think. Thanks!

a8m:
Hey @rickypai, any update on this PR?

rickypai:
I ended up renaming the model to get around the issue. I think your suggestion is great but I don't have the time to work on it right now. Maybe we can file an issue, tag me, and I can take a stab at a later time?

PR #1346, opened 2021-03-17
Add support for ON CONFLICT upserts in sql builder

ivanvanderbyl:


a8m:
Thanks for adding this @ivanvanderbyl, I'm going to review this tomorrow morning

ivanvanderbyl:
You're welcome :) 

One thing I thought of when putting the larger PR together was actually making the conflict resolution policy apply per column, instead of all columns. I posted an example of how this could look in the API on https://github.com/ent/ent/pull/1325#issuecomment-798808792

dialect/sql/builder.go
```diff
@@ -2171,6 +2271,16 @@ var ops = [...]string{
 	OpNotNull: "IS NOT NULL",
 }
 
+// A ConflictResolutionOp represents a possible action to take when an insert conflict occurrs.
+type ConflictResolutionOp int
```
a8m, line 2275: SQLite calls it ["conflict resolution"](https://sqlite.org/lang_conflict.html) and PostgreSQL calls it ["conflict actions"](https://www.postgresql.org/docs/11/sql-insert.html#SQL-ON-CONFLICT). Both are fine to me, but move these constants above the `InsertBuilder.OnConflict` method.

dialect/sql/builder.go
```diff
@@ -701,13 +726,88 @@ func (i *InsertBuilder) Query() (string, []interface{}) {
 			i.WriteByte('(').Args(v...).WriteByte(')')
 		}
 	}
+
+	if len(i.conflictColumns) > 0 {
+		// Update on conflict
+		i.buildConflictHandling()
+	}
```
a8m, line 733:
```suggestion
	if len(i.conflictColumns) > 0 {
		i.buildConflictHandling()
	}
```

This comment is redundant here.

dialect/sql/builder.go
```diff
@@ -701,13 +726,88 @@ func (i *InsertBuilder) Query() (string, []interface{}) {
...
+				if j > 0 {
+					i.Comma()
+				}
+				// Ignore conflict by setting column to itself e.g. "c" = "c"
+				i.Ident(c).WriteOp(OpEQ).Ident(c)
+			}
+		case OpResolveWithAlternateValues:
+			i.WriteString("DO UPDATE SET").Pad()
+			writeUpdateValues(i, i.updateColumns, i.updateValues)
+		}
```
a8m, line 774:
```suggestion
		i.WriteString("DO UPDATE SET ")
		switch i.onConflictOp {
		case OpResolveWithNewValues:
			for j, c := range i.columns {
				if j > 0 {
					i.Comma()
				}
				i.Ident(c).WriteOp(OpEQ).Ident("excluded").WriteByte('.').Ident(c)
			}
		case OpResolveWithIgnore:
			for j, c := range i.columns {
				if j > 0 {
					i.Comma()
				}
				// Ignore conflict by setting column to itself e.g. "c" = "c".
				i.Ident(c).WriteOp(OpEQ).Ident(c)
			}
		case OpResolveWithAlternateValues:
			writeUpdateValues(i, i.updateColumns, i.updateValues)
		}
```
Nice work!

dialect/sql/builder.go
```diff
@@ -701,13 +726,88 @@ func (i *InsertBuilder) Query() (string, []interface{}) {
...
+
+		switch i.onConflictOp {
+		case OpResolveWithIgnore:
+			for j, c := range i.columns {
+				if j > 0 {
+					i.Comma()
+				}
+				// Ignore conflict by setting column to itself e.g. `c` = `c`
+				i.Ident(c).WriteOp(OpEQ).Ident(c)
+			}
```
a8m, line 787: Move this to a helper function `writeIgnoreValues` or something else, and use it here and in the block above.

dialect/sql/builder_test.go
```diff
@@ -1395,6 +1395,42 @@ WHERE
...
+			input:     Dialect(dialect.MySQL).Insert("users").Set("email", "user@example.com").OnConflict(OpResolveWithIgnore).ConflictColumns("email"),
+			wantQuery: "INSERT INTO `users` (`email`) VALUES (?) ON DUPLICATE KEY UPDATE `email` = `email`",
+			wantArgs:  []interface{}{"user@example.com"},
+		},
+		{
+			input:     Dialect(dialect.MySQL).Insert("users").Set("email", "user@example.com").OnConflict(OpResolveWithNewValues).ConflictColumns("email"),
+			wantQuery: "INSERT INTO `users` (`email`) VALUES (?) ON DUPLICATE KEY UPDATE `email` = VALUES(`email`)",
+			wantArgs:  []interface{}{"user@example.com"},
+		},
 	}
```
a8m, line 1434: What about adding test-cases for SQLite?

a8m (commented):
Amazing work @ivanvanderbyl! Thanks a lot for pushing this effort.

dialect/sql/builder_test.go
ivanvanderbyl: Could do, but the SQLite output is identical to postgres. Fun fact, they designed their conflict resolution on the postgres SQL because there isn't a standard SQL approach to this.

dialect/sql/builder.go
ivanvanderbyl: Naming is hard, originally I called it `conflictOp` but that sounded too menacing. I'll clean up the ordering as suggested.

ivanvanderbyl:
I've updated everything as you suggested.

dialect/sql/builder_test.go
a8m: > Could do, but the SQLite output is identical to postgres. 

👍 

> Fun fact, they designed their conflict resolution on the postgres SQL because there isn't a standard SQL approach to this.

Cool 😃, thanks for the collaboration @ivanvanderbyl. It's appreciated 🙏

a8m:
Thanks for your contribution @ivanvanderbyl! I merged your PR to a [temporary branch](https://github.com/ent/ent/pull/1370) and then to master, due to git conflicts. I didn't see you in the [GitHub contributors](https://github.com/ent/ent/graphs/contributors), so I edited this commit - but still don't see you. Maybe you need to link your email (in Git) to GitHub?

Anyway, let me know when you want to continue to work on the other parts - I have some follow-up questions regarding this change.

ivanvanderbyl:
@a8m  Awesome! Happy to see this merged in and get started on the next part.

> I didn't see you in the GitHub contributors, so I edited this commit - but still don't see you. Maybe you need to link your email (in Git) to GitHub?

I think this might be cached and calculated once a day, will see if it recognized my commits tomorrow.

I'd like to hear your thoughts on how we could add an Op for upserts? I'm a little cloudy on how this would work. Also, thoughts on the API?

PR #1351, opened 2021-03-18
dialect/sql/schema: fix postgres index migration

### LINKED ISSUE #1344: Migration DROP INDEX, CREATE INDEX cycle
I've noticed with a (unique) index on multiple fields that the migration code generated by ent continually cycles between dropping the index and creating the index. After creating the table/index, the next migration includes a line to drop the index. After dropping the index, the next migration includes a line to create the index the same as before. And so on.

I've noticed that this only seems to happen when supplying a custom table name for the model and specifically when that table name is just the singular form of the model name (as opposed to the plural form that ent produces by default). Without a custom table name or with a custom table name that is not just the singular form, the issue is not encountered.

Of course there may be other cases where the same behavior occurs, but this is what I've narrowed down so far.

## Current Behavior 😯

Described above, but the generated migration code alternates between dropping and creating (unique) indices on multiple fields for a model when the model has a custom table name that is the singular form of the model name.

## Expected Behavior 🤔

Expect the migration code to contain a command to create the index and not contain commands to drop the index right after.

## Steps to Reproduce 🕹

Reproduction repo: https://github.com/henrywoody/ent-index-migration-test

Steps:

1. Create a model with a few fields and a unique index on those columns.
2. Give the model a custom table name that is the singular form of the model name (using `Config` method).
3. Generate migration (write to file).

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.15.6  |
| Ent         | 0.7.0   |
| Database    | Postgres   |
| Driver      | github.com/lib/pq |

a8m:
Fix PostgreSQL index migration when table_name = type_name.

Closed #1344

### DIFF
--- a/dialect/sql/schema/migrate.go
+++ b/dialect/sql/schema/migrate.go
@@ -390,9 +390,7 @@ func (m *Migrate) changeSet(curr, new *Table) (*changes, error) {
 
 	// Drop indexes.
 	for _, idx := range curr.Indexes {
-		_, ok1 := new.fk(idx.Name)
-		_, ok2 := new.index(idx.Name)
-		if !ok1 && !ok2 {
+		if _, isFK := new.fk(idx.Name); !isFK && !new.hasIndex(idx.Name, idx.realname) {
 			change.index.drop.append(idx)
 		}
 	}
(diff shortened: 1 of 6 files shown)


a8m:
Skipping the migration CI test, because this PR renames tables.

PR #1358, opened 2021-03-19
add helpers for basic predicates for comparing 2 columns.

rubensayshi:
this adds basic predicate helpers do compare 2 columns without having to build them yourself every time.

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -946,6 +946,14 @@ func (p *Predicate) Not() *Predicate {
 	})
 }
 
+func (p *Predicate) columnsOp(col1, col2 string, op Op) *Predicate {
+	return p.Append(func(b *Builder) {
+		b.Ident(col1)
+		b.WriteOp(op)
+		b.Ident(col2)
+	})
+}
+
 // And combines all given predicates with AND between them.
 func And(preds ...*Predicate) *Predicate {
 	p := P()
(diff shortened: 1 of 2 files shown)


dialect/sql/builder.go
```diff
@@ -968,6 +968,19 @@ func (p *Predicate) EQ(col string, arg interface{}) *Predicate {
...
+
+// EQCol appends a "=" predicate between 2 columns.
+func (p *Predicate) ColEQ(col1 string, col2 string) *Predicate {
+	return p.Append(func(b *Builder) {
+		b.Ident(col1)
+		b.WriteOp(OpEQ)
+		b.Ident(col2)
+	})
+}
+
```
a8m, line 983: Maybe define a private method `columnsOp` and use it in all other methods: 

```go

func (p *Predicate) ColumnsEQ(col1 string, col2 string) *Predicate {
	return p.columnsOp(col1, col2, OpEQ)
}

func (p *Predicate) ColumnsNEQ(col1 string, col2 string) *Predicate {
	return p.columnsOp(col1, col2, OpNEQ)
}


func (p *Predicate) columnsOp(col1, col2 string, op Op) *Predicate {
		b.Ident(col1)
		b.WriteOp(op)
		b.Ident(col2)
	})
}
```

dialect/sql/builder.go
```diff
@@ -968,6 +968,19 @@ func (p *Predicate) EQ(col string, arg interface{}) *Predicate {
 	})
 }
 
+func ColEQ(col1 string, col2 string) *Predicate {
```
a8m, line 971: What do you think about naming these predicates as `Columns<Op>` (e.g. `ColumnsEQ`) instead of `Col<Op>`? It's a little bit longer, but more clear IMO. 

BTW, I just want to mention that this query-builder API is going to be changed soon into a new implementation (I don't like this one at all).

a8m (approved):
Thanks for your contribution @rubensayshi 😃 🎉

PR #1391, opened 2021-03-25
enable crowdin translation 

rotemtam:
Enabling i18n. Currently not adding the language picking to the top navbar until we have some content in. 
Adding crowdin sync to the CD flow. 
Changing `<a>` tags that ref internal document in the index page to `<Link>` components so b/c docusaurus will account for i18n that way. 

Before merging we need to set the Crowdin env var for the auth token.


a8m:
shipitttt :shipit:

a8m (approved):

PR #1394, opened 2021-03-25
doc/md: update predicates docs about how to use predicates in subquery

chrisguox:


a8m (approved):
Thanks for your contribution @chrisguox

PR #1407, opened 2021-03-29
dialect/sql/schema: json column migration for mariadb10.3.13

### LINKED ISSUE #1395: json in mariadb migration
create a column as `JSON` type in schema.go

and do migration in mariadb 10.3 twice

it will throw error like:
`failed creating schema resources: sql/schema: changing column type for "xxxxxxx" is invalid (json != longtext)`

I think it's a mariadb issue but ent framework should deal with it.

ref:
https://mariadb.com/kb/en/json-data-type/

And I think we should change those lines to fix this issue:
https://github.com/ent/ent/blob/b4ad29f7f3970687d4e583eb643356ea23a83f03/dialect/sql/schema/schema.go#L193

a8m:
See https://github.com/ent/ent/issues/1395\#issuecomment-809420881
Fixed #1395

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,6 +141,19 @@ jobs:
           --health-start-period 10s
           --health-timeout 5s
           --health-retries 10
+      maria103:
+        image: mariadb:10.3.13
+        env:
+          MYSQL_DATABASE: test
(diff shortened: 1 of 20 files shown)


rotemtam (approved):
💯

PR #1431, opened 2021-04-07
Feat/document pb column

rotemtam:


doc/md/faq.md
```diff
@@ -532,4 +533,90 @@ func main() {
...
+	}
+	if b, ok := src.([]byte); ok {
+		if err := proto.Unmarshal(b, x); err != nil {
+			return err
+		}
+		return nil
+	}
+	return fmt.Errorf("unexpected type %T", src)
+}
+
```
a8m, line 579:
```suggestion
```

doc/md/faq.md
```diff
@@ -532,4 +533,90 @@ func main() {
...
+func (x *Hi) Scan(src interface{}) error {
+	if src == nil {
+		return nil
+	}
+	if b, ok := src.([]byte); ok {
+		if err := proto.Unmarshal(b, x); err != nil {
+			return err
+		}
+		return nil
```
a8m, line 575:
```suggestion
		return proto.Unmarshal(b, x)
```

doc/md/faq.md
```diff
@@ -532,4 +533,90 @@ func main() {
...
+
+```
+
+We add a new `field.Bytes` to our schema, setting the generated protobuf struct as its underlying `GoType`:
+
+```go
+// Fields of the Message.
+func (Message) Fields() []ent.Field {
+	return []ent.Field{
+		field.Bytes("hi").GoType(&pb.Hi{}),
```
a8m, line 588:
```suggestion
		field.Bytes("hi").
			GoType(&pb.Hi{}),
```

a8m (approved):
Thanks for adding this @rotemtam. Looks good

Please accept these changes and I'll merge this.

doc/md/faq.md
a8m: Not sure if it's preferred, because it's less explicit. You can choose to accept it or reject it - both are fine.

rotemtam: Interesting suggestion, but  I think I prefer the more explicit version

Issue #1187, opened 2021-01-19
DebugDriver should be friendlier to structured logging


mgabeler-lee-6rs:
- [x] I have searched the [issues](https://github.com/facebook/ent/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

The log calls issued by `DebugDriver` and `DebugTx` have a structure to them. Each has an action, usually some SQL, args for the SQL, and a transaction id when appropriate. When working with structured logging (e.g. JSON via zerolog), it would be useful to retain this structure in the output, instead of all of those being concatenated into a string.

For example, instead of logging something like this:
```json
{"msg":"Tx(f210ebe1-83a6-4aca-b8ed-78466b59987e).Query: query=SELECT `table`.`id` FROM `deliveries` WHERE `table`.`field` <= ? LIMIT 100 args=[xyzzy]"}
```
it would be nice to instead be able to log something like this:
```json
{
  "txId":"f210ebe1-83a6-4aca-b8ed-78466b59987e",
  "args":["xyzzy"],
  "action":"Query",
  "msg":"SELECT `table`.`id` FROM `deliveries` WHERE `table`.`field` <= ? LIMIT 100"
}
```

## Motivation 🔦

Being able to search the structured logs structurally (e.g. find all the debug logs for a given transaction) is easier when this structure is maintained.

Keeping `args` as an array avoids ambiguity when there is more than one value, and also improves processing of structured logs.

----

One possible way to achieve this:

The `DebugDriver.log` signature might look like:
```go
func (ctx context.Context, txId, action, sql string, args ...interface{}
```
The existing `Debug` and `DebugWithContext` methods could then be updated to wrap the provided log function with the current `fmt.Sprintf` templates, and new `DebugStructured` and `DebugStructuredWithContext` added (though that last function name is getting long winded...). Similar updates to the generated `config` struct and its `log` method might make sense as well, though those would need some thought as to how many of the logging options should be exposed.

In the case of non-transactional actions, `id` would be `""`. In the case of begin/commit/rollback logging, `sql` would similarly be `""`. For SQL actions, `args` naturally contains the query args. In the case of `BeginTx`, `args` could contain the `*sql.TxOptions` so that those can be logged (they cannot right now). `action` would be one of `Exec`, `Query`, `Begin`, `Rollback`, `Commit`, loosely following the current logging pattern.

Minor downside: implementing the existing `Sprintf` patterns in an adapter function would require some conditional logic to determine which format string to use.

[...]

PR #1437, opened 2021-04-08
first pass at support for structured debug logging

imle:
I took a first pass at #1187 since it looks like no one else has taken that on.

Added some basic tests to ensure the changes I put in return the same log statements as before.

I put in a `panic` as a placeholder since I wasn't sure what a good plan was for a situation where a `logEntry` was checked in without being properly filled out. Some guidance on this would be appreciated.


imle:
Wanted to note that the error for the linter was not obvious when for the missing copyright header.

efectn:
Are you planning to look at this PR? Using Zerolog/Zap with Ent would be great for me. @a8m @yonidavidson

dialect/dialect.go
```diff
@@ -76,26 +76,34 @@ func Debug(d Driver, logger ...func(...interface{})) Driver {
...
 	}
-	drv := &DebugDriver{d, func(_ context.Context, v ...interface{}) { logf(v...) }}
+	drv := &DebugDriver{d, func(_ context.Context, v LogEntry) { logf(v) }}
 	return drv
 }
 
 // DebugWithContext gets a driver and a logging function, and returns
 // a new debugged-driver that prints all outgoing operations with context.
-func DebugWithContext(d Driver, logger func(context.Context, ...interface{})) Driver {
+func DebugWithContext(d Driver, logger func(context.Context, LogEntry)) Driver {
```
a8m, line 85: This introduces a breaking change. I'll try to think of an alternative how to add it without breaking users' code and still keep the API lean.

a8m:
Thanks for submitting this btw, and thanks @giautm for reminding me to give it a look.

imle:
Sorry for not following up after putting up this PR. I think that the

imle:
Accidentally closed this while writing a comment

PR #1441, opened 2021-04-08
entc: add global annotations option

a8m:


examples/entcpkg/ent/entc.go
```diff
@@ -22,6 +23,7 @@ func main() {
 	// rest are provided with the `Templates` option.
 	opts := []entc.Option{
 		entc.TemplateFiles("template/stringer.tmpl"),
+		entc.Annotations(Annotation{StructTag: "rql"}),
```
a8m, line 26: @alexsn - here's how we pass global annotations.

entc/gen/graph.go
```diff
@@ -55,6 +55,24 @@ type (
...
+		//
+		// For type templates, we access the Config field to access the global
+		// annotations, and not the type-specific annotation.
+		//
+		//	{{- with $.Config.Annotations.GQL }}
+		//		{{/* Annotation usage goes here. */}}
+		//	{{- end }}
+		//
+		// Note that the mapping is from the annotation-name (e.g. "GQL") to a JSON decoded object.
+		Annotations Annotations
```
a8m, line 75: @alexsn - see the docs on how to access these annotations in templates.

I think the next step will be to add a `gen.Extension` (or a `gen.Plugin`), that will wrap the codegen hooks, annotations, and templates in one option.

alexsn:
LGTM

PR #1442, opened 2021-04-08
Correct wrong error message ("mysql" hard coded in Inspector method)

bshihr:


a8m (approved):
Thanks for your contribution @bshihr 🙏🏻

PR #1444, opened 2021-04-09
entc/gen/template: add GoLand annotation for template data-types

a8m:
🤯 🤯 🤯 🤯 🤯 🤯 🤯 🤯 🤯 🤯 🤯 🤯 🤯 🤯 
This annotation enables the auto-completion of templates in GoLand.  

image


a8m:
cc @alexsn

entc/gen/template/predicate.tmpl
```diff
@@ -17,7 +19,7 @@ import (
 	"entgo.io/ent/dialect/sql"
 )
 
-{{ range $_, $n := $.Nodes -}}
+{{ range $n := $.Nodes -}}
```
giautm, line 22: hey @a8m, I believe this fix is wrong. See: https://play.golang.org/p/TXAOkWN43pp

Screen Shot 2021-04-09 at 17 57 47

a8m: The range clause works a bit different in templates (see the `text/template` [doc](https://golang.org/pkg/text/template/#hdr-Variables)):

> Note that if there is only one variable, it is assigned the element; this is opposite to the convention in Go range clauses.

But I still appreciate the review. Thanks 🙏

giautm: Yeah, I have checked the doc but missed the note. :)))

PR #1453, opened 2021-04-13
Templating: Allow user to alter Query struct init

cliedeman:
After adding fields to the Query struct I found I had no way to initialize them.

Added an new template and example of use


entc/integration/template/ent/template/query.tmpl
```diff
@@ -0,0 +1,21 @@
...
+
+{{ define "dialect/sql/query/fields" }}
+	{{- with $.UnexportedForeignKeys }}
+		withFKs bool
+	{{- end }}
+	// additional query fields.
+	extra string
+{{ end }}
+
+{{ define "client/query/init" }}
```
cliedeman, line 18: not a fan of init, maybe "client/query/new"

a8m:
Hey @cliedeman! It's great to see you here again 👋 

One little question, can you elaborate on the use case? I'm just interested.

Also, did you know that you can extend all builders (including the `ent.Client`) and inject them custom dependencies? See https://entgo.io/docs/faq#how-to-extend-the-generated-builders

cliedeman:
Hi @a8m it is good to be back

I am trying to achieve this

user_query.go (this part can already be done with templates)
```go
// UserQuery is the builder for querying User entities.
type UserQuery struct {
	config
	limit      *int
	offset     *int
	order      []OrderFunc
	fields     []string
	predicates []predicate.User
	// start extension
	globalPredicate predicate.User
        // end extension
	// intermediate query (i.e. traversal path).
	sql  *sql.Selector
	path func(context.Context) (*sql.Selector, error)
}
```

And then I would like to populate the field during the `Query()` call


client.go
```go
// Query returns a query builder for User.
func (c *UserClient) Query() *UserQuery {
	return &UserQuery{config: c.config, globalPredicate: user.Active(true)}
}
```

a8m:
Got it. What about mutations? 

I think this type of issue should be addressed with hooks and interceptors. However, interceptors are not supported atm (we can start work on it), but you can implement this using a [privacy filter rule](https://entgo.io/docs/privacy) on the schema.

```go
// FilterTenantRule is a query rule that filters out entities that are not in the tenant.
func FilterTenantRule() privacy.QueryMutationRule {
    type ActiveFilter interface {
        WhereActive(...entql.BoolP)
    }
    return privacy.FilterFunc(func(ctx context.Context, f privacy.Filter) error {
        af, ok := f.(ActiveFilter)
        if !ok {
            return privacy.Denyf("unexpected filter type %T", f)
        }
        // Make sure that a tenant reads only entities that has an edge to it.
        af.WhereActive(entql.BoolEQ(true))
        // Skip to the next privacy rule (equivalent to return nil).
        return privacy.Skip
    })
}
```

entc/gen/template/client.tmpl
```diff
@@ -251,6 +254,7 @@ func (c *{{ $client }}) Hooks() []Hook {
 {{ end }}
 {{ end }}
 
-{{/* A template that can be overrided in order to add additional fields to the client.*/}}
+{{/* A template that can be overridden in order to add additional fields to the client.*/}}
 {{ define "client/fields/additional" }}{{ end }}
 
+{{ define "client/query/init" }}{{ end }}
```
a8m, line 260: Add a template comment?

entc/gen/template/dialect/sql/query.tmpl
```diff
@@ -11,6 +11,11 @@ in the LICENSE file in the root directory of this source tree.
 	{{- with $.UnexportedForeignKeys }}
 		withFKs bool
 	{{- end }}
+	{{- with $tmpls := matchTemplate "dialect/sql/query/fields/additional/*" }}
+    	{{- range $tmpl := $tmpls }}
+    		{{ xtemplate $tmpl $ }}
+    	{{- end }}
+    {{- end }}
```
a8m, line 18:
```suggestion
	{{- with $tmpls := matchTemplate "dialect/sql/query/fields/additional/*" }}
		{{- range $tmpl := $tmpls }}
			{{ xtemplate $tmpl $ }}
		{{- end }}
	{{- end }}
```

entc/gen/template/client.tmpl
```diff
@@ -204,7 +204,10 @@ func (c *{{ $client }}) DeleteOneID(id {{ $n.ID.Type }}) *{{ $n.DeleteOneName }}
 
 // Query returns a query builder for {{ $n.Name }}.
 func (c *{{ $client }}) Query() *{{ $n.QueryName }} {
-	return &{{ $n.QueryName }}{config: c.config}
+	return &{{ $n.QueryName }}{
+	    config: c.config,
+	    {{- template "client/query/init" $n }}
```
a8m, line 209: Maybe `"dialect/{{ .Storage }}/query/fields/init/*"`?

entc/integration/template/ent/template/query.tmpl
a8m: See my comment above, I think it will look better when you define the fields in one template name `"dialect/sql/query/fields/additional/example"`, and initialize them in another template named `"dialect/sql/query/fields/init/example"`.

a8m (approved):
Looks fine to me, except for these small comments.

entc/gen/template/client.tmpl
cliedeman: Will add comment

I actually dont like the name init because i find it a bit mis leading. Because the actual template is adding fields to struct init. What about `define "client/query/init/fields"`

entc/gen/template/client.tmpl
cliedeman: will do

entc/integration/template/ent/template/query.tmpl
```diff
@@ -0,0 +1,18 @@
+{{/*
+Copyright 2019-present Facebook Inc. All rights reserved.
+This source code is licensed under the Apache 2.0 license found
+in the LICENSE file in the root directory of this source tree.
+*/}}
+
+{{/* The line below tells Intellij/GoLand to enable the autocompletion based *gen.Graph type. */}}
+{{/* gotype: entgo.io/ent/entc/gen.Graph */}}
+
+{{ define "dialect/sql/query/fields/additional/example" }}
```
a8m, line 10:
```suggestion
{{ define "dialect/sql/query/fields/additional/extra" }}
```

Keept the same name in this example.

entc/gen/template/client.tmpl
cliedeman: yes

a8m:
Appreciate the contribution Ciaran. Thanks!

Issue #967, opened 2020-11-20
Question: Generate Schema Fields


cliedeman:
Is there any code available to generate a schema from an existing database?

alexsn:
No, we never had a need for such a tool even though I can see how it might be useful.

a8m:
Hey @cliedeman, I actually have a POC I created a long time ago, and can open source it in a different repo (not part of ent) as it was just an experimental project - there were multiple edge cases that I didn’t cover.

cliedeman:
Thanks @alexsn and @a8m. 

@a8m Would you mind sharing the code? I did a little experiment [here](https://github.com/cliedeman/ent/commit/808d7d8ac64c4af4e96df67161b26b8cc105780e)
and we actually have almost all we need with the migration code.

I think if we have an interface like this

```go
SqlImporter {
  AllTables(ctx context.Context, tx dialect.Tx, ignoredTables []string) (*Table, error)
}
```

We would actually have everything we need to build an importer in a standalone library.

Background:
I have some python and nodejs projects I would like to start adding some go to but for the now the database schema would be managed externally. I am looking for a tool that I could use to keep the go models in sync with the upstream models and integrate it into CI.

I previously tried hooking directly into sqlalchemy but this was proving more difficult than just inspecting the db

cliedeman:
Scratch that. after looking deeper into it a lot more queries need to happen to extract the required information from the database.

It should be completely standalone to migrations.

Sytten:
I also support that, our database schema/migrations are separate from the system using them. In other services I use https://github.com/go-jet/jet which generates golang structs and helpers based on the DB schema. I think it could serve as inspiration.

mattn:
Currently, this does only support postgresql. But I have plan to support MySQL or SQLite3.

https://github.com/mattn/entgen

masseelch:
There is now: [entimport](https://github.com/ariga/entimport)

PR #1493, opened 2021-04-22
Added foreign keys to postgres inspect

cliedeman:
Building blocks for https://github.com/ent/ent/issues/967

Allow the inspector to retrieve foreign keys using the `information_schema`.

This should be enough to create a module in ent/contrib that can generate an ent schema.

We already have all the columns, Now with foreign keys and unique indices we should be able to guess which (ent supported) relationships exist in a given database


a8m:
Thanks for the contribution @cliedeman. Let me know when it's ready for review

dialect/sql/schema/postgres.go
```diff
@@ -526,3 +526,94 @@ func arrayType(t string) bool {
...
 	return true
 }
+
+func (d *Postgres) foreignKeys(ctx context.Context, tx dialect.Tx, tables []*Table) (map[*Table][]*ForeignKey, error) {
+	var fks = make(map[*Table][]*ForeignKey)
+
+	var tableLookup = make(map[string]*Table)
+
+	// TODO: include schema
+	for _, t := range tables {
```
cliedeman, line 536: I feel like I am not properly accounting for tables in different schemas but I don't believe this is current supported anyway

cliedeman:
@a8m ready for review

a8m (commented):
Appreciate the PR @cliedeman. Please address these comments and I'll re-review it. Thanks

dialect/sql/schema/postgres.go
```diff
@@ -526,3 +526,94 @@ func arrayType(t string) bool {
...
+		tableLookup[t.Name] = t
+	}
+
+	for _, t := range tables {
```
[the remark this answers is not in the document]
cliedeman: I don't follow. The query is already run once per table

entc/integration/integration_test.go
```diff
@@ -101,6 +102,11 @@ func TestPostgres(t *testing.T) {
 					tt(t, client)
 				})
 			}
+
+			t.Run("inspect", func(t *testing.T) {
```
cliedeman, line 106: temporary measure until this is implemented for all dialects

entc/integration/inspect_test.go
```diff
@@ -0,0 +1,186 @@
+// Copyright 2019-present Facebook Inc. All rights reserved.
+// This source code is licensed under the Apache 2.0 license found
+// in the LICENSE file in the root directory of this source tree.
+
+package integration
```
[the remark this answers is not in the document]
cliedeman: I amended the unit test. Should I rather expand on that test or did you have something else in mind? I think only a M2M relation should be added

dialect/sql/schema/postgres.go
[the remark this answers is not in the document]
cliedeman: I think I understand now. I pass through all tables because I need to lookup the related columns and table to build the foreign key object

```go
&ForeignKey{
	Symbol:     constraintName,
	Columns:    []*Column{column},
	RefTable:   refTable,
	RefColumns: []*Column{refColumn}, // <--
}
```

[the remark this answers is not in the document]
a8m: 👍

dialect/sql/schema/postgres.go
```diff
@@ -526,3 +526,94 @@ func arrayType(t string) bool {
 	}
 	return true
 }
+
+func (d *Postgres) foreignKeys(ctx context.Context, tx dialect.Tx, tables []*Table) (map[*Table][]*ForeignKey, error) {
+	var fks = make(map[*Table][]*ForeignKey)
+
+	var tableLookup = make(map[string]*Table)
+
+	// TODO: include schema
```
a8m, line 535:
```suggestion
	tableLookup := make(map[string]*Table)
```

Why do we return `map[*Table][]*ForeignKey` instead of adding the FKs to the tables?

dialect/sql/schema/postgres.go
```diff
@@ -526,3 +526,94 @@ func arrayType(t string) bool {
...
+		query := fmt.Sprintf(fkQuery, t.Name)
+		if err := tx.Query(ctx, query, []interface{}{}, rows); err != nil {
+			return nil, fmt.Errorf("querying foreign keys for table %s: %w", t.Name, err)
+		}
+		defer rows.Close()
+		var (
+			tableFksLookup = make(map[string]*ForeignKey)
+			// Maintain stable order
+			tableFks []*ForeignKey
+		)
```
a8m, line 551:
```suggestion
		fksLookup := make(map[string]*ForeignKey)
```

Append directly to the table.

entc/integration/inspect_test.go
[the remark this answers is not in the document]
a8m: The inspection test should be closed to the `sql/schema`. Let's add the M2M relation test to it and remove this one.

dialect/sql/schema/inspect.go
```diff
@@ -62,6 +62,24 @@ func (i *Inspector) Tables(ctx context.Context) ([]*Table, error) {
...
+					t.AddForeignKey(fk)
+				}
+			}
+		}
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	return tables, nil
```
a8m, line 83:
```suggestion
	if ok {
		if err := fki.foreignKeys(ctx, tx, tables); err != nil {
			return nil, err
		}
	}
	return tables, nil
```

See my comment above.

a8m (approved):
Looks good! Thanks for your contribution @cliedeman.

a8m:
> This should be enough to create a module in ent/contrib that can generate an ent schema.

cc @rotemtam

dialect/sql/schema/postgres.go
cliedeman: will change it. On my. first pass I wanted the new method to not mutate any of the tables but I will change it

cliedeman:
@a8m feedback addressed

a8m (approved):
Looks good. Thanks for adding this @cliedeman

PR #1510, opened 2021-04-25
Run integration tests for inspect

cliedeman:
Created this to repo a failure I encountered on inspect for sqlite

Tracked down the failure to strange behaviour on sqlite for this query

```sql
SELECT `name`, `type`, `notnull`, `dflt_value`, `pk` FROM pragma_table_info('sqlite_sequence') ORDER BY `pk`;
```

Output:

| name | type | notnull | dflt\_value | pk |
| :--- | :--- | :--- | :--- | :--- |
| name |  | 0 | NULL | 0 |
| seq |  | 0 | NULL | 0 |

The empty type breaks table parsing

### DIFF
--- a/dialect/sql/schema/inspect.go
+++ b/dialect/sql/schema/inspect.go
@@ -56,6 +56,9 @@ func (i *Inspector) Tables(ctx context.Context) ([]*Table, error) {
 	tx := dialect.NopTx(i.sqlDialect)
 	tables := make([]*Table, 0, len(names))
 	for _, name := range names {
+		if i.skipTable(ctx, name) {
+			continue
+		}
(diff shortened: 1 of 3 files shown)


dialect/sql/schema/inspect.go
```diff
@@ -65,6 +68,15 @@ func (i *Inspector) Tables(ctx context.Context) ([]*Table, error) {
 	return tables, nil
 }
 
+// skipTable skip system tables
+func (i *Inspector) skipTable(ctx context.Context, name string) bool {
```
cliedeman, line 72: alternative fix would be to allow `.scanColumn` to handle the "" type but this is such an edge case its probably not worth it

PR #1523, opened 2021-04-30
Fix typo in "O2O Same Type" section

DuGlaser:
Double quotation was missing, so it has been added.


a8m (approved):
Thanks for your contribution @DuGlaser

a8m:
@all-contributors please add @DuGlaser for doc

PR #1533, opened 2021-05-04
sql.Builder Join include Querier's error

### LINKED ISSUE #1529: Selector should gather all errors or propagate inner error

## Current Behavior 😯

```go
func ApplyQuery(node *sqlgraph.Node, query string) func(s *sql.Selector) {
  return func(s *sql.Selector) {
    s.Where(sql.P(func(b *sql.Builder) {
            // this error is ignored - should just panic ?
	    b.AddError(errors.New("ERR"))
  }))
}
```

The error in Builder will just ignored.

In `func (b *Builder) join(qs []Querier, sep string) *Builder` error is ignored,
so in `func (s *Selector) Query() (string, []interface{})` all error is ignored.

Maybe `Builder.Join` also collect Querier's error ?

## Expected Behavior 🤔

select error out

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.16  |
| Ent         | 0.7   |
| Database    | -   |
| Driver      | -     |

wenerme:
fix #1529

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -29,6 +29,11 @@ type Querier interface {
 	Query() (string, []interface{})
 }
 
+// querierErr allowed propagate Querier's inner error
+type querierErr interface {
+	Err() error
+}
+
 // ColumnBuilder is a builder for column definition in table creation.
 type ColumnBuilder struct {
 	Builder
@@ -2091,6 +2096,7 @@ func (s *Selector) Query() (string, []interface{}) {
 		b.WriteString(strconv.Itoa(*s.offset))
 	}
 	s.total = b.total
+	s.AddError(b.Err())
 	return b.String(), b.args
 }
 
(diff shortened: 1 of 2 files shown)


dialect/sql/builder.go
```diff
@@ -2488,6 +2497,9 @@ func (b *Builder) join(qs []Querier, sep string) *Builder {
 		b.WriteString(query)
 		b.args = append(b.args, args...)
 		b.total += len(args)
+		if qe, ok := q.(querierErr); ok {
+			b.AddError(qe.Err())
+		}
```
a8m, line 2502: No need to call this method if `Err()` is nil. 
```suggestion
		if qe, ok := q.(querierErr); ok {
			if err := qe.Err(); err != nil {
				b.AddError(err)
			}
		}
```

a8m (approved):
Looks good. Thanks for adding this.

dialect/sql/builder.go
wenerme: this pr makes AddError ignored nil error, makes code more clear, or do you thing all AddError should be checked ?

a8m: We can keep the logic in `AddError` that ignores nil errors, but there's no need to call `AddError` on every `join` call if it's unnecessary. Sounds good?

wenerme: Sounds good.

PR #1537, opened 2021-05-07
schema.Column ScanDefault allowed uuid default to function

wenerme:
also add more info in error, the error before `unsupported default type: [16]byte` is too hard to find the error reason


a8m (approved):
Looks good. Thanks for adding this @wenerme

PR #1542, opened 2021-05-08
Add github.com/rs/xid usage to FAQ documentation.

shanna:
Package xid is a globally unique ID generator library that uses the
Mongo Object ID algorithm to generate 12 byte, 20 character IDs
with no configuration or external coordination service.

PR adds XID examples to FAQ as requested.


a8m (approved):
Thanks for adding this @shanna.

PR #1548, opened 2021-05-08
specific field collation support

### LINKED ISSUE #1540: schema/field: support setting field collation
Is there any way to set `Collation` for all tables at once rather than creating a `Mixin` and include it? I need to set the default `Collation` to `utf8mb4_unicode_ci` instead of `utf8mb4_bin`.

mahmud2011:
**Summary**
Specific field collation will override default and table level collation.

**Issue**
Fixes #1540


a8m (approved):
Looks good! Thanks for adding this @mahmud2011

This change doesn't support PostgreSQL, but I can add it in a separate PR.

Issue #1513, opened 2021-04-27
Keep order of Fields and Edges columns on indexes


petersondmg:
- [x] I have searched the [issues](https://github.com/facebook/ent/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

```go
index.Edges("myEdge").Fields("myField")
```

should generate a different index than

```go
index.Fields("myField").Edges("myEdge")
```

## Motivation 🔦

The order of columns on index is relevant.
Currently `Edges` columns are always added after `Fields` columns and this prevents more specific optimizations.

a8m:
Hey @petersondmg and thanks for suggesting this.

You can achieve this today by using [edge fields](https://entgo.io/docs/schema-edges#edge-field). I added [an example](https://github.com/ent/ent/pull/1514) to our integration tests. Would you like to add it to our documentation?

PR #1566, opened 2021-05-17
docs/md: add index edge fields example to docs

BrentChesny:
#1513


a8m:
Thanks for adding this @BrentChesny, it's appreciated 🙏 

@all-contributors please add @BrentChesny for docs

Issue #1484, opened 2021-04-20
Case insensitive search


michaelkub:
Hi, 
I needed to perform case insensitive search, I didn’t find generated predicate for this, but because I am using mariadb, I used the following approach :
`field.String("name").SchemaType(
			map[string]string{
				dialect.MySQL: "varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL",
			}),`
and it worked great. 
So, is this approach will continue to work and won't be remove in next versions or it maybe it will be replaced ?

a8m:
Hey @michaelkub and it's great to see you here again :)

Ent generates case-insensitive predicates for string fields named `<F>ContainsFold` and `<F>EqualFold`. Don't you see thse predicates in your project?

a8m:
Closing, but please feel free to reopen if you didn't find those methods or you have some trouble generating them. Thanks 🙏

a8m:
Hey @ppphp,
Functional indexes are not supported atm, but they can be added like #1545. However, I think the current code can be replaced with something as follows: `where <column> COLLATE utf8mb4_general_ci = <value>`. Pretty sure it's more efficient, but didn't test it tbh. Do you want to contribute this?

This code:
https://github.com/ent/ent/blob/6d3df1469c831c26fc09f39c3417b0254911e2c7/dialect/sql/builder.go#L1350-L1359

Should be replaced with something as follows:

https://github.com/ent/ent/blob/6d3df1469c831c26fc09f39c3417b0254911e2c7/dialect/sql/builder.go#L1378-L1381

PR #1572, opened 2021-05-18
Dialect: improve equalfold

sywesk:
Hello ! 

I recently found a ticket mentioning the performance problem of EqualFold because it used "lower", and the impact it had on the different indexes. (#1484 )

I took some inspiration from the ContainsFold method and updated EqualFold to follow the same pattern, except for SQLite that needed the old way of doing (because if we have to lower on both sides, let's just use equal)

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -1351,9 +1351,18 @@ func (p *Predicate) EqualFold(col, sub string) *Predicate {
 	return p.Append(func(b *Builder) {
 		f := &Func{}
 		f.SetDialect(b.dialect)
-		f.Lower(col)
-		b.WriteString(f.String())
-		b.WriteOp(OpEQ)
+		switch b.dialect {
+		case dialect.MySQL:
+			// We assume the CHARACTER SET is configured to utf8mb4,
+			// because this how it is defined in dialect/sql/schema.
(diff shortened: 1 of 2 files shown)


dialect/sql/builder.go
```diff
@@ -1351,9 +1351,18 @@ func (p *Predicate) EqualFold(col, sub string) *Predicate {
...
 		f := &Func{}
 		f.SetDialect(b.dialect)
-		f.Lower(col)
-		b.WriteString(f.String())
-		b.WriteOp(OpEQ)
+		switch b.dialect {
+		case dialect.MySQL:
+			// We assume the CHARACTER SET is configured to utf8mb4,
+			// because this how it is defined in dialect/sql/schema.
+			b.Ident(col).WriteString(" COLLATE utf8mb4_general_ci LIKE ")
```
a8m, line 1358: I think this works too, right?

```suggestion
			b.Ident(col).WriteString(" COLLATE utf8mb4_general_ci = ")
```

a8m (approved):
Thanks for adding this @sywesk 😃

dialect/sql/builder.go
sywesk: Thanks for those :eagle: :eyes: ! I'll correct this

PR #1577, opened 2021-05-19
Make non-string ValueScanner types work with enum fields

### LINKED ISSUE #1575: Enum with GoType that are not strings doesn't work
Let's say I define a type that is backed by int and fulfils `ValueScanner` interface:
```golang
type Animal int

const (
	Dog Animal = iota
	Cat
)

var (
	animals = [...]string{"Dog", "Cat"}
)

func (a Animal) String() string {
	return animals[p]
}

// Value implements database/sql/driver.Valuer
func (a Animal) Value() (driver.Value, error) {
	return animals[p], nil
}

// Scan implements database/sql.Scanner
func (a *Animal) Scan(src interface{}) error {
	srcStr, ok := src.(string)
	if !ok {
		return errors.New("wrong src type")
	}
	i, ok := utils.Find(animals[:], srcStr)
	if !ok {
		return fmt.Errorf("animals is invalid: %s", srcStr)
	}
	*a = Animal(i)
	return nil
}

// Values is part of `field.EnumValues` interface in ent package
func (Animal) Values() []string {
	return animals[:]
}
```

And then I define enum field based on it:
```golang
	return []ent.Field{
		field.Enum("type").
			GoType(animal.Dog),
         }
}
```

The code generation works but the code doesn't compile due the following function that is created:
```golang
func AnimalValidator(a animal.Animal) error {
	switch a {
	case "Dog", "Cat":
		return nil
	default:
		return fmt.Errorf("pet: invalid enum value for type field: %q", a)
	}
}
```

According to documentation `GoType` can be either string type or any type that implements `ValueScanner` like I did. Is there any way to support such enums ?

- [ x] The issue is present in the latest release.
- [ x] I have searched the [issues](https://github.com/facebook/ent/issues) of this repository and believe that this is not a duplicate.

adayNU:
This change fixes #1575 which now makes it possible to specify a GoType
for an enum that is not a string alias. It requires that if the specified
type is not a string alias, the type must satisfy the Stringer interface.


adayNU:
So it looks like this works assuming you don't use `Default` (since that takes a string value).

Seems _okay_ to merge this incrementally w/o fully fixing that? Alternatively we could modify that to accept an interface (which I _think_ would be a non-breaking change) but am less confident doing that without a 👍 .

entc/gen/template/meta.tmpl
```diff
@@ -130,7 +130,7 @@ const (
 	{{ $name := $f.Validator }}
 	// {{ $name }} is a validator for the "{{ $f.Name }}" field enum values. It is called by the builders before save.
 	func {{ $name }}({{ $receiver }} {{ $enum }}) error {
-		switch {{ $receiver }} {
+		switch {{ $receiver }}{{ if $f.IsNonStringEnum }}.String(){{ end }} {
```
a8m, line 133:
```suggestion
		switch {{ $receiver }}{{ if $f.Type.Stringer }}.String(){{ end }} {
```

This is also OK because the validation already happened on the `schema/field` package.

entc/gen/type.go
```diff
@@ -991,6 +991,15 @@ func (f Field) IsInt() bool { return f.Type != nil && f.Type.Type == field.TypeI
...
 
+// IsNonStringEnum returns true if the field is an enum field w/ a non-string GoType.
+func (f Field) IsNonStringEnum() bool {
+	if f.Type != nil && f.Type.Type != field.TypeEnum {
+		return false
+	}
+
+	return f.HasGoType() && f.Type.RType.Kind != reflect.String
+}
+
```
a8m, line 1002:
```suggestion
```

This code is not needed. See the code suggestion above.

a8m (approved):
Thanks, @adayNU for adding this. It's appreciated 🙏

entc/gen/template/meta.tmpl
adayNU: I think this is problematic in cases where you just provide string values (i.e. `Values("foo", "bar")`) due to:

https://github.com/adayNU/ent/blob/draft/non-string-enums/entc/gen/template/meta.tmpl#L118-L126

Because below typed constant values are used in the case statement: 

https://github.com/adayNU/ent/blob/draft/non-string-enums/entc/gen/template/meta.tmpl#L134

entc/integration/ent/schema/fieldtype.go
```diff
@@ -212,6 +212,9 @@ func (FieldType) Fields() []ent.Field { //nolint:funlen
 		field.Enum("role").
 			Default(string(role.Read)).
 			GoType(role.Role("role")),
+		field.Enum("priority").
+			// There is still an issue with Default (since it accepts string).
+			GoType(role.Priority(0)),
```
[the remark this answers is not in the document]
adayNU: Agreed. I can remove the comment for now.

entc/gen/template/meta.tmpl
a8m: > I think this is problematic in cases where you just provide string values (i.e. Values("foo", "bar")) due to:

Not sure why it's problematic. Can you provide a field example that will cause this template to produce invalid code?

a8m:
Thanks for adding this @adayNU! 🎉

PR #1588, opened 2021-05-22
docs: change `github.com/facebook/ent` to `github.com/ent/ent`

y-yagi:
This reduces an unneeded redirect.


a8m (approved):
Thanks @y-yagi 🙏

PR #1602, opened 2021-05-26
Remove rollback

cliedeman:
Rollback is not used anywhere.


a8m:
Thanks @cliedeman for adding this ❤️

PR #1608, opened 2021-05-28
field.JSON support Nillable

wenerme:


a8m:
Hey @wenerme and thanks for opening this PR 🙏 

Can you explain what problem this PR solves, and why this option is not enough (`field.JSON("t", T{})` -> `field.JSON("t" &T{})`)?

wenerme:
@a8m Hi, using ent's model to generate other stuff, need to check is it real nillable, ent consider value nillable  and field nillable is nillable, but outside of ent need a more proper way to tell is it nillable.

a8m:
Do you mean to check if the value is a pointer type? If not, can you elaborate on what exactly you try to do?

You can use the `gen.Field.Info` to get the field type information.  See https://github.com/ent/ent/pull/1611 for example.

PR #1629, opened 2021-06-08
fix nil dereference (SIGSEGV) on db errors during Save with Hooks

mgabeler-lee-6rs:
For entities that have hooks, the generated `Save` code includes a bit that looks like this:
```go
			sc.mutation = mutation
			node, err = sc.sqlSave(ctx)
			mutation.id = &node.ID
```
If `sqlSave` returns `nil, someError`, this crashes.

Fix: add an `err` check here, and move the existing `err` check in the bulk save up so that it protects against this.

Verified via go.mod replace directive that this fixes the real world crash in my app.

----

NB: PR is broken into two commits to simplify review -- first templates update, then all the generated code. Happy to squash if it looks good


giautm:
hey @a8m, I just re-generate based on master branch and I found similar code in git diff. Can we fix it soon?

a8m:
Thanks @mgabeler-lee-6rs for fixing this! 🙏

a8m (approved):

PR #1639, opened 2021-06-12
Fix case insensitive match on non ASCII data in SQLite

Qluxzz:
The current implementation for SQLite used the LOWER function for transforming the string, which according to SQLite's documentation only works for ASCII characters
> The lower(X) function returns a copy of string X with all ASCII characters converted to lower case. The default built-in lower() function works for ASCII characters only. To do case conversions on non-ASCII characters, load the ICU extension. 

https://sqlite.org/lang_corefunc.html#lower

I replaced the query with `COLLATE NOCASE`:

> NOCASE - Similar to binary, except that it uses sqlite3_strnicmp() for the comparison. Hence the 26 upper case characters of ASCII are folded to their lower case equivalents before the comparison is performed. Note that only ASCII characters are case folded. SQLite does not attempt to do full UTF case folding due to the size of the tables required. Also note that any U+0000 characters in the string are considered string terminators for comparison purposes.

https://sqlite.org/datatype3.html#collating_sequences

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -1386,14 +1386,21 @@ func (p *Predicate) EqualFold(col, sub string) *Predicate {
 			// We assume the CHARACTER SET is configured to utf8mb4,
 			// because this how it is defined in dialect/sql/schema.
 			b.Ident(col).WriteString(" COLLATE utf8mb4_general_ci = ")
+			b.Arg(sub)
 		case dialect.Postgres:
 			b.Ident(col).WriteString(" ILIKE ")
-		default: // SQLite.
+			b.Arg(sub)
+		case dialect.SQLite:
+			b.Ident(col)
+			b.WriteOp(OpEQ)
+			b.Arg(sub)
+			b.WriteString(" COLLATE NOCASE ")
+		default:
 			f.Lower(col)
 			b.WriteString(f.String())
 			b.WriteOp(OpEQ)
+			b.Arg(strings.ToLower(sub))
 		}
-		b.Arg(strings.ToLower(sub))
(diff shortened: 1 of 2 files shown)


Qluxzz:
Seems like there's no easy solution for this problem. The correct solution would be to enable the ICU extension for SQLite, but as noted in the SQLite FAQ, this would cause the SQLite library to double in size https://www.sqlite.org/faq.html#q18

The SQLite driver used in Ent supports enabling the ICU extension via build tags.
https://github.com/mattn/go-sqlite3#feature--extension-list
Extension | Build Tag | Description
-- | -- | --
International Components for Unicode | sqlite_icu | This option causes the International Components for Unicode or "ICU" extension to SQLite to be added to the build

a8m:
Hey @Qluxzz and thanks for your contribution.

> Seems like there's no easy solution for this problem. The correct solution would be to enable the ICU extension for SQLite, but as noted in the SQLite FAQ, this would cause the SQLite library to double in size https://www.sqlite.org/faq.html#q18

I'm not sure I have a suggestion on how to tackle this problem atm, but if you're blocked on this, you can use this PR as a custom predicate and build your Go project with `sqlite_icu` build flag. 

Something a bit ugly, but worth thinking about is to choose the SQLite predicate based on the Go build flag.

Qluxzz:
After enabling the ICU extension it works as expected, since LOWER now works for non ASCII-characters, so I'll close this PR.

I don't know if we'd want to mention this in the documentation somewhere that you'd have to build it with the sqlite_icu flag to enable full unicode support. I know it's not really ent's concern.

PR #1666, opened 2021-06-28
Add IndexBuilder.IfNotExists and use for Create() case.

bshihr:
Append-only mode creates tables with the `IF NOT EXISTS` annotation.
Make indices case symmetric with this.

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -524,6 +524,7 @@ type IndexBuilder struct {
 	Builder
 	name    string
 	unique  bool
+	exists  bool
 	table   string
 	columns []string
 }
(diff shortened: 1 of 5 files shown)


dialect/sql/schema/migrate.go
```diff
@@ -208,7 +208,7 @@ func (m *Migrate) txCreate(ctx context.Context, tx dialect.Tx, tables ...*Table)
 			}
 			// indexes.
 			for _, idx := range t.Indexes {
-				query, args := m.addIndex(idx, t.Name).Query()
+				query, args := m.addIndex(idx, t.Name).IfNotExists().Query()
```
a8m, line 211: Can you please move this configuration to `sqlite.go` and `postgres.go` as it's not supported by MySQL/MariaDB.

a8m (approved):
Thanks for your contribution @bshihr. Looks good except for a minor comment.

dialect/sql/schema/migrate.go
bshihr: Done.

a8m (approved):
🎉 🎉 🎉 

Thanks for your contribution @bshihr!

PR #1675, opened 2021-07-01
website/blog: automatic graphql filters generation

a8m:


doc/website/blog/2021-07-01-automatic-graphql-filters-generation.md
```diff
@@ -0,0 +1,328 @@
...
+      node {
+        id
+      }
+    }
+  }
+}
+```
+
+### Background
+Many libraries that deal with data in Go choose the path of passing around empty interace instances
```
arielitovsky, line 63: Typo `interace` => interface

Issue #1374, opened 2021-03-22
Why adding "ON DELETE SET NULL" on foreign keys?


AlexanderMatveev:

AlexanderMatveev:
Accidentally posted without description.
Here are my models:

```golang
func (User) Edges() []ent.Edge {
	return []ent.Edge{
		//...
		edge.To("streams", Stream.Type),
		/...
	}
}

func (Stream) Edges() []ent.Edge {
	return []ent.Edge{
		// ...
		edge.From("owner", User.Type).
			Ref("streams").
			Unique().
			Required(),
		// ...
	}
}
```

Migrations provided:
```sql
-- ...
ALTER TABLE "streams" ADD CONSTRAINT "streams_users_streams" FOREIGN KEY("user_streams") REFERENCES "users"("id") ON DELETE SET NULL;
-- ...
```

Why `ON DELETE SET NULL` is added `Required()` in edge?

a8m:
Hello @AlexanderMatveev and thanks for opening this issue.

> Why ON DELETE SET NULL is added `Required()` in edge?

The `ON DELETE SET NULL` clause is added by default to all non-M2M relations because their columns are not configured as `NOT NULL` columns in the database (edge checks happen only on the application level). However, you can configure the foreign-key reference action differently to one of [these actions](https://github.com/ent/ent/blob/master/dialect/entsql/annotation.go#L130-L136) using [edge annotations](https://entgo.io/docs/schema-annotations/#foreign-keys-configuration). 

Does it solve your problem?

AlexanderMatveev:
Hi @a8m, thanks for response.

My point is that default PostgreSQL action is `NO ACTION`, which same as `RESTRICT`, see `referential_action` here https://www.postgresql.org/docs/current/sql-createtable.html. And I think it must be a strong argument of making `SET NULL` as default, because this creates an unexpected integrity violation.

And, about `OnDelete` annotation. There is no `OnDelete` field in 0.7.0 release: https://github.com/ent/ent/blob/v0.7.0/dialect/entsql/annotation.go

a8m:
> And, about OnDelete annotation. There is no OnDelete field in 0.7.0 release: https://github.com/ent/ent/blob/v0.7.0/dialect/entsql/annotation.go

You can pull the version in master. It's stable.

[...]

PR #1703, opened 2021-07-09
entc/gen: set foreign-key columns non-nullable for required edges

### LINKED ISSUE #1688: Edge field makes required field nullable in database
Before adding edge field, the required field was defined as `NOT NULL`
However once I added a required edge field for this field (https://entgo.io/docs/schema-edges#edge-field) the field was changed to be of type `NULL` which is incorrect

## Current Behavior 😯

Define the following schema:
```
// Fields of the Post.
func (Post) Fields() []ent.Field {
    return []ent.Field{
        field.Int("author_id"),
    }
}

// Edges of the Post.
func (Post) Edges() []ent.Edge {
    return []ent.Edge{
        edge.To("author", User.Type).
            // Bind the "author_id" field to this edge.
            Field("author_id").
            Required(),
    }
}
```

Database migration code is:
```
CREATE TABLE IF NOT EXISTS "post"(
  "author_id" BIGINT NULL,
  PRIMARY KEY("id")
);
```

## Expected Behavior 🤔

The database migration code should be:
```
CREATE TABLE IF NOT EXISTS "post"(
  "author_id" BIGINT NOT NULL,
  PRIMARY KEY("id")
);
```

## Your Environment 🌎

I also ran the migration with `WithForeignKeys(false)`

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.16  |
| Ent         | 0.8   |
| Database    | Postgres   |
| Driver      | https://github.com/jackc/pgx        |

a8m:
Note, this only applies to edges without circular references.
Fixed #1688 and #1374


naorsec:
Hi @a8m can I help somehow with merging this PR ?
It's really needed for my use case

naormatania:
Hi @a8m can I help somehow with merging this PR ?
It's really needed for my use case

edigaryev:
We need this feature at @cirruslabs to ensure consistency and I've tried reproducing the failed migration test locally and got this:

```
=== RUN   TestMySQL/56
    enttest.go:67: sql/schema: alter table "groups": Error 1830: Column 'group_info' cannot be NOT NULL: needed in a foreign key constraint 'test/groups_group_infos_info' SET NULL
    --- FAIL: TestMySQL/56 (0.05s)
```

The previous migration created by `master` had a `ON DELETE SET NULL` constraint on a `group_info` column, but the new migration tries to change the `group_info` column type to `NOT NULL`, which makes no sense without changing the constraint too and triggers [error 1830](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_fk_column_not_null).

It seems that the current migration logic is somewhat append-only and won't change any constraints:

https://github.com/ent/ent/blob/14cba0abb63025518bed2f77a95b6db6c5b5c86f/dialect/sql/schema/migrate.go#L136-L144

@a8m do you think adding a separate migration option that will enable schema modifications such as constraint updates would be a reasonable way solve this?

I've also read in some issues that a work on a new migration engine is being done, do you have any plans to de-compose this task into smaller chunks and make this effort more transparent overall, so that the community could help?

iwata:
@a8m Any progress on this pull request?

a8m:
Hey all, this PR actually waited a long time for the new migration engine, due to migration failures, and thanks to [github.com/ariga/atlas](https://github.com/ariga/atlas), this issue is resolved now and I plan to merge it on Sunday morning. CI is green, and the only thing left is to update the documentation. 

If you didn't have a chance to see our v0.10 yesterday, here's the [announcement post](https://entgo.io/blog/2022/01/20/announcing-new-migration-engine/). Thanks 🙏 ❤️

PR #1720, opened 2021-07-15
schema/field: support unique bytes

tarrencev:
Allows `bytea` fields to be indexes in postgres.

I see it is possible to make `Other` types unique, but in that case it is not possible to use the column as a foreign key.


a8m:
Thanks for your contribution @tarrencev ❤️

schema/field/field.go
```diff
@@ -549,6 +549,13 @@ func (b *bytesBuilder) Optional() *bytesBuilder {
 	return b
 }
 
+// Unique makes the field unique within all vertices of this type.
+// Only supported in PostgresSQL.
```
a8m, line 553:
```suggestion
// Only supported in PostgreSQL.
```

Found a typo after merging it if you want to send a patch @tarrencev

PR #1721, opened 2021-07-15
Add support for postgres and mysql numeric and decimal types in migrations

rubensayshi:
Currently if you have a `field.Float("price").SchemaType(map[string]string{dialect.Postgres: "numeric(6, 4)"})` field then the migration will forever think that it needs to change the type because when scanning the existing table structure it will see it as just `numeric` without the precision and scale.

This PR fixes that by looking at the `numeric_precision` and `numeric_scale` fields which are present in the `information_schema.columns` for both engines.

I chose to use the `c.SchemaType` instead of adding the metadata to the `Column{}` since currently that's also how it's being created anyway.
I suppose in the future we could maybe add a `field.Decimal` field type and use `big.Float` or https://github.com/shopspring/decimal as the GoType for it ... 

I added 2 test cases, one to check it no longer continuously keeps trying to migrate the column and one which checks that you can actually migrate the column.

Also added a better error message to make it a little easier to figure out which entity is causing issues.

### DIFF
--- a/dialect/sql/schema/inspect_test.go
+++ b/dialect/sql/schema/inspect_test.go
@@ -60,41 +60,42 @@ func TestInspector_Tables(t *testing.T) {
 							AddRow("pets").
 							AddRow("groups").
 							AddRow("user_groups"))
-					mock.ExpectQuery(escape("SELECT `column_name`, `column_type`, `is_nullable`, `column_key`, `column_default`, `extra`, `character_set_name`, `collation_name` FROM `INFORMATION_SCHEMA`.`COLUMNS` WHERE `TABLE_SCHEMA` = ? AND `TABLE_NAME` = ?")).
+					mock.ExpectQuery(escape("SELECT `column_name`, `column_type`, `is_nullable`, `column_key`, `column_default`, `extra`, `character_set_name`, `collation_name`, `numeric_precision`, `numeric_scale` FROM `INFORMATION_SCHEMA`.`COLUMNS` WHERE `TABLE_SCHEMA` = ? AND `TABLE_NAME` = ?")).
 						WithArgs("public", "users").
-						WillReturnRows(sqlmock.NewRows([]string{"column_name", "column_type", "is_nullable", "column_key", "column_default", "extra", "character_set_name", "collation_name"}).
-							AddRow("id", "bigint(20)", "NO", "PRI", "NULL", "auto_increment", "", "").
-							AddRow("name", "varchar(255)", "YES", "YES", "NULL", "", "", "").
-							AddRow("text", "longtext", "YES", "YES", "NULL", "", "", "").
-							AddRow("uuid", "char(36)", "YES", "YES", "NULL", "", "", "utf8mb4_bin"))
+						WillReturnRows(sqlmock.NewRows([]string{"column_name", "column_type", "is_nullable", "column_key", "column_default", "extra", "character_set_name", "collation_name", "numeric_precision", "numeric_scale"}).
+							AddRow("id", "bigint(20)", "NO", "PRI", "NULL", "auto_increment", "", "", nil, nil).
+							AddRow("name", "varchar(255)", "YES", "YES", "NULL", "", "", "", nil, nil).
+							AddRow("text", "longtext", "YES", "YES", "NULL", "", "", "", nil, nil).
+							AddRow("uuid", "char(36)", "YES", "YES", "NULL", "", "", "utf8mb4_bin", nil, nil).
+							AddRow("price", "decimal(6, 4)", "NO", "YES", "NULL", "", "", "", "6", "4"))
 					mock.ExpectQuery(escape("SELECT `index_name`, `column_name`, `sub_part`, `non_unique`, `seq_in_index` FROM `INFORMATION_SCHEMA`.`STATISTICS` WHERE `TABLE_SCHEMA` = ? AND `TABLE_NAME` = ? ORDER BY `index_name`, `seq_in_index`")).
 						WithArgs("public", "users").
 						WillReturnRows(sqlmock.NewRows([]string{"index_name", "column_name", "sub_part", "non_unique", "seq_in_index"}).
 							AddRow("PRIMARY", "id", nil, "0", "1"))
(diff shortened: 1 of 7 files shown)


dialect/sql/schema/inspect_test.go
```diff
@@ -280,38 +287,41 @@ func TestInspector_Tables(t *testing.T) {
...
 	}
 }
 
-func columnsMatch(t *testing.T, got, expected []*Column) {
+func columnsMatch(t *testing.T, drv string, got, expected []*Column) {
 	require.Equal(t, len(expected), len(got))
 	for i := range got {
 		c1, c2 := got[i], expected[i]
-		require.Equal(t, c1.Name, c2.Name)
-		require.Equal(t, c1.Nullable, c2.Nullable)
```
rubensayshi, line 303: sorry, couldn't resist fixing the order of `actual, expected` here .. the error message being wrong is really confusing.

dialect/sql/schema/postgres.go
```diff
@@ -229,8 +232,11 @@ func (d *Postgres) scanColumn(c *Column, rows *sql.Rows) error {
 		nullable sql.NullString
 		defaults sql.NullString
 		udt      sql.NullString
+		numericPrecision sql.NullInt64
+		numericScale sql.NullInt64
```
liptakmatyas, line 236: NIT: Formatting is off.

liptakmatyas (changes requested):
I think the code itself is OK.
However, there are some formatting problems.

dialect/sql/schema/mysql.go
```diff
@@ -377,12 +380,15 @@ func (d *MySQL) prepare(ctx context.Context, tx dialect.Tx, change *changes, tab
 }
 
 // scanColumn scans the column information from MySQL column description.
+// nolint:funlen
```
rubensayshi, line 383: `scanColumn` went over the 80 max with my additions, though I think it's hard an unnecessary to break this up

tarrencev: Looks like @a8m has a change here to bump this limit: https://github.com/ent/ent/pull/1703/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R11

tarrencev:
I've seen this behavior too. Thanks for proposing a fix!

dialect/sql/schema/mysql.go
a8m: >  I think it's hard an unnecessary to break this up

Agree. Please change `.golangci.yml` as mentioned by @tarrencev

dialect/sql/schema/migrate.go
```diff
@@ -14,6 +14,7 @@ import (
 	"entgo.io/ent/dialect"
 	"entgo.io/ent/dialect/sql"
 	"entgo.io/ent/schema/field"
+	"github.com/pkg/errors"
```
a8m, line 17:
```suggestion
```

See change below.

dialect/sql/schema/migrate.go
```diff
@@ -189,7 +190,7 @@ func (m *Migrate) txCreate(ctx context.Context, tx dialect.Tx, tables ...*Table)
 			}
 			change, err := m.changeSet(curr, t)
 			if err != nil {
-				return err
+				return errors.Wrapf(err, "creating changeset for %s", t.Name)
```
a8m, line 193:
```suggestion
				return fmt.Errorf("creating changeset for %q: %w", t.Name, err)
```

dialect/sql/schema/mysql.go
```diff
@@ -421,8 +427,15 @@ func (d *MySQL) scanColumn(c *Column, rows *sql.Rows) error {
 		default:
 			c.Type = field.TypeInt8
 		}
-	case "numeric", "decimal", "double":
+	case "double":
 		c.Type = field.TypeFloat64
+	case "numeric", "decimal":
+		c.Type = field.TypeFloat64
+		// if precision is specified then we should take that into account
```
a8m, line 434:
```suggestion
		// If precision is specified then we should take that into account.
```

dialect/sql/schema/postgres.go
```diff
@@ -247,8 +252,15 @@ func (d *Postgres) scanColumn(c *Column, rows *sql.Rows) error {
 		c.Type = field.TypeInt64
 	case "real":
 		c.Type = field.TypeFloat32
-	case "numeric", "decimal", "double precision":
+	case "double precision":
 		c.Type = field.TypeFloat64
+	case "numeric", "decimal":
+		c.Type = field.TypeFloat64
+		// if precision is specified then we should take that into account
```
a8m, line 259:
```suggestion
		// If precision is specified then we should take that into account.
```

dialect/sql/schema/postgres_test.go
```diff
@@ -820,6 +820,65 @@ func TestPostgres_Create(t *testing.T) {
 				mock.ExpectCommit()
 			},
 		},
+		{
+			// there shouldn't be a migration if the numeric type didn't change (bugfix)
```
a8m, line 824:
```suggestion
```

dialect/sql/schema/mysql_test.go
```diff
@@ -1149,6 +1149,67 @@ func TestMySQL_Create(t *testing.T) {
 				mock.ExpectCommit()
 			},
 		},
+		{
+			// there shouldn't be a migration if the numeric type didn't change (bugfix)
```
a8m, line 1153:
```suggestion
```

a8m (approved):
Looks good @rubensayshi! Thanks for adding this 🙏 

There are a few small style fixes. Please accept these changes and I'll merge this.

rubensayshi:
thanks for the suggestions, squashed them all into my commit.

and as always: thanks for the time and effort you put into maintaining this <3

a8m:
> thanks for the suggestions, squashed them all into my commit.
> 
> and as always: thanks for the time and effort you put into maintaining this <3

Thanks for the kind words and for your contribution ❤️

Issue #1718, opened 2021-07-14
UpdateDefault func was not supported without Time


godcong:
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

Due to time zone issues, I use `Int64` to save the time.
But I found that Int64 has no `UpdateDefault` function.

## Motivation 🔦

I hope this function can be used in scenarios other than Time.
After looking at the implementation, it doesn't seem to be complicated, but I don't know if it will affect other functions.

otherwise, `DefaultFunc` now used interface{} as a parameter
Whether to change to use the corresponding type is more convenient to use
like:
```go
func (b *int64Builder) DefaultFunc(fn interface{}) *int64Builder {
	b.desc.Default = fn
	return b
}
```
=>
```go
func (b *int64Builder) DefaultFunc(fn func()int64) *int64Builder {
	b.desc.Default = fn
	return b
}
```

a8m:
Hey @godcong, and thanks for proposing this.

I don't mind adding it, but if you want to contribute this change, I'll be happy to support you if you need any help. 

BTW, if you're blocked on this atm, you can use [ent/hooks](https://entgo.io/docs/hooks/#schema-hooks) for setting this field for you until the `DefaultFunc` option will be supported.

godcong:
The last version implements this.
i think this can be closed?

a8m:
Yes!  This is actually a really useful feature. I already use it in a few places.

Thanks again for your contribution @godcong.

PR #1725, opened 2021-07-16
schema/field: add UpdateDefault to numeric (#1718)

godcong:
support set default value on update with numeric

### DIFF
--- a/schema/field/numeric.go
+++ b/schema/field/numeric.go
@@ -547,6 +547,18 @@ func (b *int8Builder) Validate(fn func(int8) error) *int8Builder {
 	return b
 }
 
+// UpdateDefault sets the function that is applied to set default value
+// of the field on update. For example:
+//
+//	field.Int8("int8").
+//		Default(0).
+//		UpdateDefault(0),
+//
+func (b *int8Builder) UpdateDefault(f func() interface{}) *int8Builder {
+	b.desc.UpdateDefault = f
+	return b
+}
+
 // StorageKey sets the storage key of the field.
 // In SQL dialects is the column name and Gremlin is the property.
 func (b *int8Builder) StorageKey(key string) *int8Builder {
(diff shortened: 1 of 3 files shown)


schema/field/numeric.go
```diff
@@ -547,6 +547,18 @@ func (b *int8Builder) Validate(fn func(int8) error) *int8Builder {
...
+//
+//	field.Int8("int8").
+//		Default(0).
+//		UpdateDefault(0),
+//
+func (b *int8Builder) UpdateDefault(f func() interface{}) *int8Builder {
+	b.desc.UpdateDefault = f
+	return b
+}
+
```
a8m, line 561: Hey @godcong, the return type should match the field type, right?

```suggestion
// UpdateDefault sets the function that is applied to set default value
// of the field on update. For example:
//
//	field.Int8("int8").
//		Default(0).
//		UpdateDefault(GenNumber()),
//
func (b *int8Builder) UpdateDefault(f func() int8) *int8Builder {
	b.desc.UpdateDefault = f
	return b
}
```

Also, this file is generated and you should change it by updating the template in `schema/field/internal/numeric.tmpl `. Thanks 🙏

godcong: I look forward to such a change, because it is easier to understand.
But I found that this change will affect the original design.
I don’t know if this destruction is acceptable.
relate test code：
```
	type CustomInt int

	f1 := func() CustomInt { return 1000 }
	fd := field.Int("id").DefaultFunc(f1).GoType(CustomInt(0)).Descriptor()
	assert.NoError(t, fd.Err)

	fd = field.Int("id").DefaultFunc(f1).Descriptor()
	assert.Error(t, fd.Err, "`var _ int = f1()` should fail")

	f2 := func() int { return 1000 }
	fd = field.Int("dir").GoType(CustomInt(0)).DefaultFunc(f2).Descriptor()
	assert.Error(t, fd.Err, "`var _ CustomInt = f2()` should fail")
```
If this can be done, I will fix the relevant changes.

a8m: Hey @godcong, I merged your changes to https://github.com/ent/ent/pull/1770 and finished the implementation. Would you mind giving it a look before I merge it? Thanks 🙏

a8m:
PR addressed with #1770

PR #1727, opened 2021-07-16
fix BC break in schema by falling back to the original method.

rubensayshi:
hmm, in https://github.com/ent/ent/pull/1428/files a BC break in the schema was introduced that is a bit of a pain to migrate.

the main issue is that `RType.Ident` is blank in old schema, I think a very simple fix is to fallback to using the `Type.String()` as `Type.Ident` isn't blank and can be used.  

ugh this is harder to fix than it looks ...  
I *think* this doesn't cause any harm and will allow for smooth upgrading...

it isn't very pretty though as someone could accidently start using `Type.RType.String()` again somewhere in the future instead of `Type.RTypeString()` ...

### DIFF
--- a/entc/gen/type.go
+++ b/entc/gen/type.go
@@ -1044,7 +1044,7 @@ func (f Field) ScanType() string {
 		if f.Nillable && !f.standardNullType() {
 			return "sql.NullScanner"
 		}
-		return f.Type.RType.String()
+		return f.Type.RTypeString()
 	}
 	switch f.Type.Type {
 	case field.TypeJSON, field.TypeBytes:
@@ -1069,7 +1069,7 @@ func (f Field) ScanType() string {
 // nillable-type supported by the SQL driver (e.g. []byte).
 func (f Field) NewScanType() string {
 	if f.Type.ValueScanner() {
-		expr := fmt.Sprintf("new(%s)", f.Type.RType.String())
+		expr := fmt.Sprintf("new(%s)", f.Type.RTypeString())
 		if f.Nillable && !f.standardNullType() {
 			expr = fmt.Sprintf("&sql.NullScanner{S: %s}", expr)
 		}
@@ -1103,7 +1103,7 @@ func (f Field) ScanTypeField(rec string) string {
 			expr = "*" + expr
 		}
 		if f.Nillable && !f.standardNullType() {
-			return fmt.Sprintf("%s.S.(*%s)", expr, f.Type.RType.String())
+			return fmt.Sprintf("%s.S.(*%s)", expr, f.Type.RTypeString())
 		}
 		return expr
 	}
--- a/schema/field/type.go
+++ b/schema/field/type.go
@@ -102,6 +102,15 @@ func (t TypeInfo) String() string {
 	}
 }
 
+// String returns the string representation of a type.
+func (t TypeInfo) RTypeString() string {
+	if t.RType != nil && t.RType.String() != "" {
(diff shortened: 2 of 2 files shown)


a8m:
Hey @rubensayshi, can you please share an example that break the compatibility and why the current solution is not enough for migration (from https://github.com/ent/ent/pull/1428):

>  After this change, users that use a `GoType(&T{})` will get `*T` in all generated interfaces (instead of `T`). However, if users still want to use values (instead of pointers), they can pass `T{}` to `GoType` (e.g. `GoType(T{})`), even if `T` doesn't implement the `sql.Scanner` interface (but `*T` does).

rubensayshi:
@a8m you always ask good questions <3

if you're reliant on `schema/snapshot` and had `GoType(&T)` and upgraded ent then you can run in to this issue.

if necessary I can make a demo project for you if you'd like, but maybe just showing the json in `internal/schema.go` and the resulting messed up code will do ;)

I have this field:
```
		field.String("rate").
			SchemaType(map[string]string{dialect.Postgres: rateNumeric}).
			GoType(&decimal.Decimal{}),
```


This is what is in my schema json for this field from an earlier version of ent:

```
        {
          "name": "rate",
          "type": {
            "Type": 7,
            "Ident": "decimal.Decimal",
            "PkgPath": "github.com/shopspring/decimal",
            "Nillable": true,
            "RType": {
              "Name": "Decimal",
              "Ident": "",
              "Kind": 25,
              "PkgPath": "github.com/shopspring/decimal",
              "Methods": {}
            }
          },
          "position": {
            "Index": 1,
            "MixedIn": false,
            "MixinIndex": 0
          },
          "schema_type": {
            "postgres": "numeric(12, 6)"
          }
        },
```
 
but atm when I want to re generate the code I get the following output for `scanValues` 

```
// scanValues returns the types for scanning values from sql.Rows.
func (*AuctionOrder) scanValues(columns []string) ([]interface{}, error) {
	values := make([]interface{}, len(columns))
	for i := range columns {
		switch columns[i] {
				case entauctionorder.FieldRate:
					values[i] = new()

```

I think this is simply because for some reason the older version of ent we were using was creating a schema json with the `RType.Ident` left blank (newer version actually sets it properly).  
Tbh I haven't looked too deep into why it was leaving `RType.Ident` blank, maybe there aren't too many people with this issue ...



I'm not sure how others do this, but to avoid a myriad of issues we had in the past, we actually put most ent files on gitignore and regenerate when we switch branches / add new fields or entities. thus always forcing the usage of the schema json.  
(we only keep the `ent/generate.go`, `ent/schema` and `ent/internal`). 

Because of this we hit this problem with the schema json BC break.

A temporary fix for me was to regenerate the ent files with the old version of ent, and then without first removing the gitignored files regenerate it with the new version. then it doesn't need to use the schema json and will create a fresh and proper schema json.

I think anyone who uses the same approach could have this issue upgrading, or someone who upgraded and was doing schema changes at the same time (which is a bad idea, but you know ... people do the weirdest things :P)


Maybe just updating the title and description of this issue to hopefully direct anyone else running into this issue and reading that they should just do things step by step will actually suffice for now?  
Or if they read this and it doesn't solve their issue we can reopen this?

rubensayshi:
to leave a breadcrumb for anyone else who got the same issue, you actually get this error while doing `go generate`:

```
format file ent/auctionorder.go: ent/auctionorder.go:158:34: expected type, found ')'
```

Issue #1460, opened 2021-04-15
Is it possible to generate readonly (SELECTs only) code?


h7kanna:
Hi, I want to develop an application that has read-only access to Database, It is just used for SELECT's. In the default configuration, the code generates CRUD API. How can we restrict an entity to be only read-only, that is only generate the READ part of the CRUD and the predicates?

a8m:
Hey @h7kanna, I actually wanted to add this functionality using a [feature-flag](https://entgo.io/docs/feature-flags) a long time ago.  We can add this to the next release because I expect the change is not too big. 

Since it's not supported at the moment, you can block mutations using [hooks](https://entgo.io/docs/hooks) or defining a read-only driver.

```go
// Reject all mutations using hooks.
client.Use(hook.Reject(ent.OpCreate|ent.OpUpdate|ent.OpUpdateOne|ent.OpDelete|ent.OpDeleteOne))
```

```go
// Reject all exec operations using a custom driver. 
drv, err := sql.Open("<dialect>", "<mysql-dsn>")
if err != nil {
	// ...
}
drv = ReadOnly(drv)
ent.NewClient(ent.Driver(drv))

// ...

// ReadOnly returns a new driver that rejects all exec operations.
func ReadOnly(drv dialect.Driver) dialect.Driver {
	return &readonly{drv}
}

type readonly struct { dialect.Driver }
func (r *readonly) Exec(context.Context, string, interface{}, interface{}) error { return ErrReadOnly } 
func (r *readonly) Tx(context.Context) (dialect.Tx, error) { return nil, ErrReadOnly }
```

h7kanna:
Hi Thanks for your quick response. I am actually looking to be able to generate less code for the readonly app.
I will wait for the feature-flag. Ent is awesome!

a8m:
Sure, as I mentioned above, I wanted to add this functionality a long time ago because I have a generated ent-client that uses [osquery](https://osquery.io/) actually, so I'm going to address this in the upcoming weeks. Thanks!

PR #1733, opened 2021-07-17
Add readonly client

cliedeman:
Allow the user to create a readonly version of the client that can only perform `Query` and `Get` operations

Transactions still need to be handled

Motviation:
I would to convery extra intent in differnet parts of the application by using a readonly client. E.g. in graphql resolvers there should only be reads.


a8m:
Hey @cliedeman and thanks for creating this PR. For a read-only mode, I suggest users https://github.com/ent/ent/issues/1460#issuecomment-820614860, or to use a custom template, same as I used in my projects:

```go
{{ define "readonly" }}
{{ template "header" $ }}

// ReadOnly returns a new readonly client.
//
//	client := client.ReadOnly()
//
func (c *Client) ReadOnly() *Client {
	cfg := config{driver: &readonly{Driver: c.driver}, log: c.log}
	return &Client{
		config: cfg,
		{{- if $.SupportMigrate }}
			Schema: migrate.NewSchema(cfg.driver),
		{{- end }}
		{{ range $_, $n := $.Nodes -}}
			{{ $n.Name }}: New{{ $n.Name }}Client(cfg),
		{{ end -}}
	}
}

// ErrReadOnly returns when a read-only user tries to execute a write operation.
var ErrReadOnly = &PermissionError{cause: "permission denied: read-only user"}

// PermissionError represents a permission denied error.
type PermissionError struct {
	cause string
}

func (e PermissionError) Error() string { return e.cause }

type readonly struct {
	dialect.Driver
}

func (r *readonly) Exec(context.Context, string, interface{}, interface{}) error {
	return ErrReadOnly
}

func (r *readonly) Tx(context.Context) (dialect.Tx, error) {
	return nil, ErrReadOnly
}
{{ end }}
```

a8m:
To be honest, I'd rather add it to the FAQ/examples as a referenced implementation, than add it to the core of the framework ❤️

cliedeman:
@a8m no problem. I will try the external template. 

The example above won't work for me because I want a compile time failure instead of runtime. 

Can this be added to contrib or a similar repo? FAQ and examples are ok but they won't have CI or stay up to date. 

Ciaran

cliedeman:
@a8m thanks for the pointers. I managed to combine the 2 approaches and I am quite happy with the result.

PR #1738, opened 2021-07-19
entc/gen: return an error instead of panic when import forgotten

a8m:
Return an error instead of panic in case of `ent/runtime` import was forgotten. 

cc @cliedeman


cliedeman:
Nice

PR #1742, opened 2021-07-19
entc/gen: add support for helper templates

### LINKED ISSUE #1655: Extend External Template Support

## Summary 💡

1. Currently every templates one adds to extend ent is run against the `Graph`. I'd like to be able to run templates against the `Type`s. 
2. Another feature I'd like to see is to be able to define templates that do not result in another file generated.

## Motivation 🔦
I am trying to extend `ent` in the way its done by `entql`. I am implementing crud http handlers. There are two issues, that came up while doing so:

1. If one creates a lot of code per `Type` the generated file gets huge, since there is one created for the whole `Graph`.
2. If I am to define helper templates, `ent` tries to generate a go file out of them.

```go
{{ define "http/create" }}
  {{- with extend $ "Package" "http" -}}
    {{ template "header" . }}
  {{ end }}

  {{ range $node := $.Nodes }}
    // This is {{ $node.Name }} create handler
    {{ template "helper" }}
  {{ end }}
{{ end }}

{{ define "helper" }}
  // some stuff
{{ end }}
```
The template above will create one file `http/create.go`, that has every single `Type` in it. I'd like to be able to have several files `http/type1_create.go`, `http/type1_read.go`, `http/type2_create.go`, ...
Having defined the template `helper` results in ent creating a file `helper.go` with `// some stuff` as content, which is not a valid go file.

a8m:
Closed #1655

cc @masseelch

### DIFF
--- a/doc/md/templates.md
+++ b/doc/md/templates.md
@@ -71,6 +71,52 @@ In order to override an existing template, use its name. For example:
 {{ end }}
 ```
 
+## Helper Templates
+
+As mentioned above, `ent` writes each template's execution output to a file named the same as the template.
+For example, the output from a template defined as `{{ define "stringer" }}` will be written to a file named
+`ent/stringer.go`.
+
+By default, `ent` writes each template declared with `{{ define "<name>" }}` to a file. However, it is sometimes
+desirable to define helper templates - templates that will not be invoked directly but rather be executed by other
+templates. To facilitate this use case, `ent` supports two naming formats that designate a template as a helper.
+The formats are:
+
+1\. `{{ define "helper/.+" }}` for global helper templates. For example:
+
+```gotemplate
+{{ define "helper/foo" }}
+    {{/* Logic goes here. */}}
+{{ end }}
+
+{{ define "helper/bar/baz" }}
+    {{/* Logic goes here. */}}
+{{ end }}
+```
+
+2\. `{{ define "<root-template>/helper/.+" }}` for local helper templates. A template is considered as "root" if
+its execution output is written to a file. For example:
+
+```gotemplate
+{{/* A root template that is executed on the `gen.Graph` and will be written to a file named: `ent/http.go`.*/}}
+{{ define "http" }}
+    {{ range $n := $.Nodes }}
+        {{ template "http/helper/get" $n }}
+        {{ template "http/helper/post" $n }}
+    {{ end }}
+{{ end }}
+
+{{/* A helper template that is executed on `gen.Type` */}}
+{{ define "http/helper/get" }}
+    {{/* Logic goes here. */}}
+{{ end }}
+
+{{/* A helper template that is executed on `gen.Type` */}}
+{{ define "http/helper/post" }}
+    {{/* Logic goes here. */}}
+{{ end }}
+```
+
 ## Annotations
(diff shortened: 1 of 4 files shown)


entc/gen/graph.go
```diff
@@ -598,29 +599,52 @@ func (g *Graph) typ(name string) (*Type, bool) {
 	return nil, false
 }
 
-// templates returns the template.Template for the code and external templates
-// to execute on the Graph object if provided.
+// templates returns the Template to execute on the Graph,
```
masseelch, line 602: previosly it was `templates.Template`. Removed on purpose?

entc/gen/graph.go
```diff
@@ -598,29 +599,52 @@ func (g *Graph) typ(name string) (*Type, bool) {
...
 			if parse.IsEmptyTree(tmpl.Root) {
 				continue
 			}
 			name := tmpl.Name()
-			// If the template does not override or extend one of
-			// the builtin templates, generate it in a new file.
-			if templates.Lookup(name) == nil && !extendExisting(name) {
+			switch {
+			// Helper templates can be either global (prefixed with "helper/"),
+			// or local, where their names follow the format: <root-tmpl>/helper/.+).
```
masseelch, line 620: Surround with `"`: `format: "<root-tmpl>/helper/.+")`

entc/gen/graph.go
a8m: Yes.

entc/gen/graph.go
a8m: Right, thanks

doc/md/templates.md
```diff
@@ -71,6 +71,50 @@ In order to override an existing template, use its name. For example:
 {{ end }}
 ```
 
+## Helper Templates
+
+As mentioned above, `ent` writes the template execution output to a file named as the template.
+For example, a template defined as `{{ define "stringer" }}` will be written to a file named `ent/stringer.go`.
```
rotemtam, line 77:
```suggestion
As mentioned above, `ent` writes each template's execution output to a file named the same as the template.
For example, the output from a template defined as `{{ define "stringer" }}` will be written to a file named `ent/stringer.go`.
```

doc/md/templates.md
```diff
@@ -71,6 +71,50 @@ In order to override an existing template, use its name. For example:
...
+
+By default, `ent` write all templates declared with `{{ define "<name>" }}` to files. However, sometimes
+you want to define templates that will function as helpers (i.e. executed by other templates). For this,
+`ent` allows you to name your templates in 2 formats to mark them as helpers. The formats are:
```
rotemtam, line 81:
```suggestion
By default, `ent` writes each template declared with `{{ define "<name>" }}` to a file. However,
it is sometimes desirable to define helper templates - templates that will not be invoked directly but rather be executed by other templates. To facilitate this use case, `ent` supports two naming formats that designate a template as a helper. The formats are:
```

rotemtam (commented):
Looks great! Some minor comments on the documentation text

doc/md/templates.md
masseelch: I think it has to be `...named the same like the template.`. I could be wrong though, not my native language.

PR #1745, opened 2021-07-19
Return selector's alias as a table name

timoha:
In case selector's from clause has *Selector instead of *SelectTable,
TableName() function (used for column checking) will panic.

Instead, return selector's alias which is useful if the query returns
a table with the same schema as some entity. Essentially, this allows
materialized views to reuse the same entity code as long as columns
are matching.


a8m (approved):
Thanks for your contribution @timoha 🙏

Issue #1750, opened 2021-07-20
Improve Rejection Error


cliedeman:
```golang
client := ...
client.Use(hook.Reject(ent.OpCreate | ent.OpUpdate | ent.OpUpdateOne | ent.OpDelete | ent.OpDeleteOne))
client.Card.Create().SetNumber("1234").SaveX(ctx)
// err: Op(31) operation is not allowed
```

- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

The hook rejection error is hard to decipher. I expected the error to be

```
Create operation is not allowed
```

## Motivation 🔦

Trying to get better feedback for readonly driver

cliedeman:
@a8m do you know of a good pattern or example of deconstructing a unioned enum in go?

Something to convert 
```
ent.OpCreate | ent.OpUpdate
```
to Create, Update

PR #1751, opened 2021-07-20
Improved hook rejection error

cliedeman:
Fix for #1750


cliedeman:
Looks like there is some test interaction failures. Running the TestDeletion in isolation passes but it failes when run together

cliedeman:
> Each connection to ":memory:" opens a brand new in-memory sql database, so if the stdlib's sql engine happens to open another connection and you've only specified ":memory:", that connection will see a brand new database. A workaround is to use "file::memory:?cache=shared" (or "file:foobar?mode=memory&cache=shared"). Every connection to this string will point to the same in-memory database.

From go-sqlite3. This is probably unintentional.
```
file:ent?mode=memory&cache=shared&_fk=1
```
So all the tests share the same database

PR #1753, opened 2021-07-20
entc/gen: add Exec and ExecX to Create

tarrencev:
Exposes `Exec` and `ExecX` on Creator structs. This aligns the interface with `Update` and `Delete` and is useful when the client wants to construct operations and does not care about the return value.

For example, the operations can be combined in a slice
```
type Execer interface {
    Exec(ctx context.Context) error
}

var execers []Execer
execers = append(execers, client.User.Create()...)
execers = append(execers, client.User.Update()...)
execers = append(execers, client.User.Delete()...)

for _, e := range execers {
    if err := e.Exec(ctx); err != nil {
        ...
    }
}
```


a8m (approved):
I like it tbh.

Would you mind migrating places that use `_ = client...SaveX()`, or `_, err = client...Save()` to this new API? See examples:

https://github.com/ent/ent/blob/master/entc/integration/integration_test.go#L649
https://github.com/ent/ent/blob/master/entc/integration/integration_test.go#L685
https://github.com/ent/ent/blob/master/entc/integration/integration_test.go#L687
https://github.com/ent/ent/blob/master/entc/integration/integration_test.go#L1065

PR #1769, opened 2021-07-25
fix panics during initTemplates

cliedeman:
When trying to debug code generation I came across a lot of panics.
These don't affect output but maybe debugging very difficult (during templating)

The error is caused because an empty config is used to invoke `import.tmpl`

https://github.com/ent/ent/blob/0751c342c65ee155fc7a76bdb775793cf0ca1b7f/entc/gen/template.go#L198

and Storage is referenced in the template

https://github.com/ent/ent/blob/0751c342c65ee155fc7a76bdb775793cf0ca1b7f/entc/gen/template/import.tmpl#L28


entc/gen/template/import.tmpl
```diff
@@ -52,9 +52,11 @@ import (
 			{{- xtemplate $tmpl $ }}
 		{{- end }}
 	{{- end }}
+	{{- if $.Storage }}
 	{{- with $tmpls := matchTemplate (printf "dialect/%s/import/additional/*" $.Storage) }}
 		{{- range $tmpl := $tmpls }}
 			{{- xtemplate $tmpl $ }}
 		{{- end }}
 	{{- end }}
+	{{- end }}
```
a8m, line 61:
```suggestion
	{{- if $.Storage }}
		{{- with $tmpls := matchTemplate (printf "dialect/%s/import/additional/*" $.Storage) }}
			{{- range $tmpl := $tmpls }}
				{{- xtemplate $tmpl $ }}
			{{- end }}
		{{- end }}
	{{- end }}
```

Fix indentation?

a8m (approved):
Thanks for the fix ❤️

entc/gen/template/import.tmpl
cliedeman: Sorted. Is there some sort of template formatter perhaps?

PR #1775, opened 2021-07-27
entc/integration: use exec when not interested in entity return value

tarrencev:


a8m (approved):
🚀 

Thanks @tarrencev!

tarrencev:
@a8m do you mind retrying ci? seems it is flakey due to timeout?

PR #1792, opened 2021-08-02
WIP: 

seiichi1101:


a8m:
Hey @seiichi1101, can you explain why [the edge field option](https://entgo.io/docs/schema-edges#edge-field) does not this?

a8m:
Hey @seiichi1101, can you please explain why [the edge field option](https://entgo.io/docs/schema-edges#edge-field) does not solve this?

seiichi1101:
Hey @a8m, sorry for lack of the information.

When I add [edge field](https://entgo.io/docs/schema-edges/#edge-field) `ent` automatically generate `QueryXXX()` method in like below.

- schema

```
// Fields of the User.
	return []ent.Field{
		field.String("name").
			NotEmpty().Unique(),
	}
}

// Edges of the User.
	return []ent.Edge{
		edge.To("friend", User.Type),
	}
}
```

- client.go

```
// QueryFriend queries the friend edge of a User.
func (c *UserClient) QueryFriend(u *User) *UserQuery {
	query := &UserQuery{config: c.config}
	query.path = func(ctx context.Context) (fromV *dsl.Traversal, _ error) {

		fromV = g.V(u.ID).Both(user.FriendLabel)
		return fromV, nil
	}
	return query
}
```

If I'd like to get all `friend` of user `alice`, I have two way to achieve this.

1. `client.User.QueryFriend(alice).AllX(ctx)`
2. `client.User.Query().Where(user.IDEQ(alice.ID)).QueryFriend().AllX(ctx)`

In my opinion, 1st does not fit when I only know the user's `id` because it's requiring `User` object. 
About 2nd, it can query by only user's `id` but it looks redundant. 

So I proposed the method `func (c *UserClient) QueryFriendId(Id string) *UserQuery`.

seiichi1101:
Hi @a8m, 

I know this PR includes breaking changes, so that I kept it in Draft PR. 
Do you think this feature is not necessary? Or is there a problem with my implementation?
Anyway, I need your advice.

PR #1793, opened 2021-08-02
entc/gen: add support for upsert/on-conflict feature-flag

a8m:
Implement the `Upsert` and `UpsertBulk` APIs (i.e. `ON CONFLICT`/`ON DUPLICATE KEY UPDATE`) using the `sql/upsert` feature flag.

```go
// Simple case.
client.User.Create().
	SetPhone("0000").
	OnConflict().
	DoNothing().
	ExecX(ctx)

// Simple case.
client.User.Create().
	SetName("Mashraki").
	SetPhone("0000").
	OnConflictColumns(
		 // This option is required only on PostgreSQL.
		sql.ConflictColumns(user.FieldPhone),
	).
	// Tell ent to use the new values on UPDATE.
	UpdateNewValues().
	ExecX(ctx)

// Use a custom `UPDATE` function and get back
// the ID fields without executing additional query.
id, err := client.User.Create().
	SetName("Boring").
	SetPhone("0000").
	OnConflict().
	// Name was set to the new value ("Boring").
	UpdateName().
	SetAddress("localhost").
	ID(ctx)

// Use a custom `UPDATE` function and get back
// the ID fields without executing additional query.
id, err := client.User.Create().
	SetName("Boring").
	SetPhone("0000").
	OnConflict().
	// Override some of the fields with a custom update.
	Update(func(u *ent.UserUpsert) {
		// Name was set to the new value ("Boring").
		u.SetNewName()
		u.SetAddress("localhost")
	}).
	ID(ctx)

// Use a custom `UPDATE` function and get back
// the ID fields without executing additional query.
id, err := client.User.Create().
	SetName("Boring").
	SetPhone("0000").
	OnConflict(
		sql.ConflictColumns(user.FieldPhone),
		// Ignore changes for all columns except for the ones below.
		sql.ResolveWithIgnore(),
	).
	// Override some of the fields with a custom update.
	Update(func(u *ent.UserUpsert) {
		// Name was set to the new value ("Boring").
		u.SetNewName()
		u.SetAddress("localhost")
	}).
	ID(ctx)

// Bulk upsert.
client.User.CreateBulk(builders...).
	OnConflictColumns(user.FieldPhone).
	UpdateNewValues().
	Exec(ctx)
```

The `Save` option will be added in the future as the `RETURNING` clause is not supported by all databases/versions.


sneakywombat:
For this block, what would you think about swapping the update func behavior to not require repeating oneself?  EG, here we say SetName("Boring") and then call u.SetNewName() in update.  It'd be more natural to me to negate the behavior.  EG, if in a conflict, fields i said previously to set already would by default be updated, but i have the option to prevent it and do something else.  In this case, if I didn't want what I said previously in the Create call to happen, i'd just say u.SkipName(), or something. The wording is also strange to me.  Whats the difference between SetName("Boring") and SetNewName?  Does the latter use the formers string argument?  What do you think?

```
id, err := client.User.Create().
	SetName("Boring").
	SetPhone("0000").
	OnConflict(
		sql.ConflictColumns(user.FieldPhone),
		// Ignore changes for all columns except for the ones below.
		sql.ResolveWithIgnore(),
	).
	// Override some of the fields with a custom update.
	Update(func(u *ent.UserUpsert) {
		// Name was set to the new value ("Boring").
		u.SetNewName()
		u.SetAddress("localhost")
	}).
	ID(ctx)
```

masseelch:
To add my two cents to that: If I wanted to have an update instead of an insert I'd want to tell.ent what fields to update, so the behaviour is intuitive, at least for me. 

The wording though is strange, I agree. Since I have an update now, why can't I simply call `UpdateName`?

a8m:
Hey @sneakywombat and @masseelch, and thanks for the feedback. 

The `OnConflict(...sql.ConflictOption)` method takes a list of SQL conflict options, that allow configuring the `INSERT` statement in the SQL level. For example, `sql.ResolveWithNewValues()` tells the builder to update the row with the new values proposed for insert (i.e. `VALUES(...)`), the `sql.ResolveWith(func(*sql.UpdateSet))` allows customizing the `UPDATE` statement, and there are additional options like, `sql.DoNothing()`, [`sql.ResolveWithIgnore()`](https://github.com/ent/ent/blob/gen/onconflict/dialect/sql/builder.go#L817-L840), and [more](https://github.com/ent/ent/blob/gen/onconflict/dialect/sql/builder.go#L767-L798). 

That means, the following code is valid:

```go
client.User.Create()
	SetName("...").
	OnConflict(sql.DoNothing()).
	Exec(ctx)

// INSERT INTO "users" ("name") VALUES(?) ON CONFLICT DO NOTHING

client.User.Create()
	SetName("...").
	OnConflict(sql.ResolveWithNewValues()).
	Exec(ctx)

// INSERT INTO "users" ("name") VALUES(?) ON CONFLICT DO UPDATE SET "name" = "excluded"."name"
//
// "excluded" means the new value exists in "VALUES". See:
// https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT

client.User.Create()
	SetRank(1).
	SetName("...").
	OnConflict(
		// Multiple options override each other.
		sql.ResolveWithNewValues(),
		sql.ResolveWith(func(s *UpdateSet) {
			// Override some of the columns with a different value.
			s.Set("rank", 2)
		}),
	).
	Exec(ctx)

// INSERT INTO "users" ("rank", "name") VALUES(1, "..") ON CONFLICT DO UPDATE SET "name" = "excluded"."name", "rank" = 2
```

More examples exist [here](https://github.com/ent/ent/blob/gen/onconflict/dialect/sql/builder_test.go#L1704-L1716).

I also wanted to add a type-safe API for updating the columns when a customized UPDATE is needed. Therefore, I added the `Update(*<T>Upsert)` method. I'll try to explain this by example:

```go
client.User.Create()
	SetRank(1).
	SetName("...").
	SetCompany("...").
	OnConflict(
		// Update all fields with the new values proposed for insertion
		// (in the create builder), except for the ones below.
		sql.ResolveWithNewValues(),
	).
	Update(func(u *UserUpsert) {
		u.SetRank(2)
	}).
	Exec(ctx)

// INSERT INTO "users" ("rank", "name", "company") VALUES(1, "..", "..") ON CONFLICT DO UPDATE SET "name" = "excluded"."name", "company" = "excluded"."company", "rank" = 2


client.User.Create()
	SetRank(1).
	SetName("...").
	SetCompany("...").
	OnConflict().
	Update(func(u *UserUpsert) {
		// Update the "rank" field with a custom value (different from the one that was set on create).
		u.SetRank(2)
		// Update the "name" field with the value that was set on create (i.e. the "new value").
		u.SetNewName()
	}).
	Exec(ctx)

// INSERT INTO "users" ("rank", "name", "company") VALUES(1, "..", "..") ON CONFLICT DO UPDATE SET "name" = "excluded"."name", "rank" = 2
```

> The wording though is strange, I agree. Since I have an update now, why can't I simply call UpdateName?

@masseelch, I'm more than happy to get suggestions for the new naming. Additional ideas I have in minds are:

```go
client.User.Create().
	Set..().
	OnConflict().
	// Avoid using the SQL package.
	DoNothing().
	Exec(ctx)

client.User.Create().
	Set..().
	OnConflict().
	// Avoid using the SQL package.
	UpdateWithNewValues().
	Exec(ctx)

client.User.Create().
	Set..().
	OnConflict().
	// Update the "name" field to the value that was set on create.
	UpdateName().
	// Provide a custom value to the "rank" field (different from the one that was set on create).
	SetRank(2).
	Exec(ctx)
```

cc @rotemtam

masseelch:
There is one more concern I'd like to raise:

The current solution is kind of a mixed up API. There is the `OnConflict(opts ...sql.ConflictOption)`-method integrated into the fluent builders. I can tell ent how to handle the conflict by giving options like with `sql.ResolveWithNewValues()`. But I can also change the conflict-handling behaviour by the use of more methods on the fluent builder like with `.Update(func(u *UserUpsert) {}).` This feels weird to me, cause I can need to configure the conflict handling in two separate places. 

I say hide the dialect and expose high level methods on the builder, like in this idea of yours:

```go
client.User.Create().
	Set..().
	OnConflict().
	// Update the "name" field to the value that was set on create.
	UpdateName().
	// Provide a custom value to the "rank" field (different from the one that was set on create).
	SetRank(2).
	Exec(ctx)
```

a8m:
Thanks for the feedback @masseelch ❤️. In general, I try to push to the fluent API logic that is compatible across all different databases/dialects, but still let users the freedom to fallback to the SQL builder if they have database-specific logic that is not exported by the framework (e.g. like you can do with custom predicates). 

I think, in most cases, users will use something like this:

```go
client.User.Create().
	Set..().
	OnConflict().
	Ignore().
	Exec(ctx)

client.User.Create().
	Set..().
	OnConflict().
	UpdateNewValues().
	Exec(ctx)

client.User.Create().
	Set..().
	OnConflict().
	Set...().
	Exec(ctx)
```

But we still want to unblock them from executing queries like:

```go
client.User.Create().
	Set..().
	OnConflict(
		sql.ConflictWhere(...),
		sql.ResolveWithNewValues(...),
		sql.UpdateWhere(...),
	).
	Exec(ctx)

// INSERT INTO "users" (...) VALUES(...) ON CONFLICT WHERE ... DO UPDATE SET ... WHERE ...
```

rotemtam:
I think having very intuitive support for the most common cases is great. If I think from the novice dev perspective, in terms of auto-completion, it's much easier to discover the conflict action options if they are presented this way:
```go
client.User.Create().
	Set..().
	OnConflict().
	// Avoid using the SQL package.
	DoNothing().
	Exec(ctx)

client.User.Create().
	Set..().
	OnConflict().
	// Avoid using the SQL package.
	UpdateNewValues().
	Exec(ctx)
```
than to need to explore the `sql` package which can be quite eclectic to explore.

GreenHatHG:
> Implement the `Upsert` and `UpsertBulk` APIs (i.e. `ON CONFLICT`/`ON DUPLICATE KEY UPDATE`) using the `sql/upsert` feature flag.
> 
> ```go
> // Simple case.
> client.User.Create().
> 	SetPhone("0000").
> 	OnConflict().
> 	DoNothing().
> 	ExecX(ctx)
> 
> // Simple case.
> client.User.Create().
> 	SetName("Mashraki").
> 	SetPhone("0000").
> 	OnConflictColumns(
> 		 // This option is required only on PostgreSQL.
> 		sql.ConflictColumns(user.FieldPhone),
> 	).
> 	// Tell ent to use the new values on UPDATE.
> 	UpdateNewValues().
> 	ExecX(ctx)
> 
> // Use a custom `UPDATE` function and get back
> // the ID fields without executing additional query.
> id, err := client.User.Create().
> 	SetName("Boring").
> 	SetPhone("0000").
> 	OnConflict().
> 	// Name was set to the new value ("Boring").
> 	UpdateName().
> 	SetAddress("localhost").
> 	ID(ctx)
> 
> // Use a custom `UPDATE` function and get back
> // the ID fields without executing additional query.
> id, err := client.User.Create().
> 	SetName("Boring").
> 	SetPhone("0000").
> 	OnConflict().
> 	// Override some of the fields with a custom update.
> 	Update(func(u *ent.UserUpsert) {
> 		// Name was set to the new value ("Boring").
> 		u.SetNewName()
> 		u.SetAddress("localhost")
> 	}).
> 	ID(ctx)
> 
> // Use a custom `UPDATE` function and get back
> // the ID fields without executing additional query.
> id, err := client.User.Create().
> 	SetName("Boring").
> 	SetPhone("0000").
> 	OnConflict(
> 		sql.ConflictColumns(user.FieldPhone),
> 		// Ignore changes for all columns except for the ones below.
> 		sql.ResolveWithIgnore(),
> 	).
> 	// Override some of the fields with a custom update.
> 	Update(func(u *ent.UserUpsert) {
> 		// Name was set to the new value ("Boring").
> 		u.UpdateName()
> 		u.SetAddress("localhost")
> 	}).
> 	ID(ctx)
> 
> // Bulk upsert.
> client.User.CreateBulk(builders...).
> 	OnConflictColumns(user.FieldPhone).
> 	UpdateNewValues().
> 	Exec(ctx)
> ```
> 
> The `Save` option will be added in the future as the `RETURNING` clause is not supported by all databases/versions.

Hi. Is there any way to implement `INSERT ... ON CONFLICT DO NOTHING/UPDATE` by myself?

PR #1810, opened 2021-08-05
Fix typos in documentation

arielitovsky:
I found these while reading the documentation.

Screen Shot 2021-08-05 at 9 17 29 AM


a8m (approved):
Thanks @arielitovsky ❤️

PR #1814, opened 2021-08-06
update index comments

msal4:
Comments do not reflect an up-to-date usage of indexes.
I'm new to ent. Let me know if i missed something.

### DIFF
--- a/schema/index/index.go
+++ b/schema/index/index.go
@@ -106,9 +106,7 @@ func (b *Builder) StorageKey(key string) *Builder {
 //
 //		// Partial index on name where the entity is not deleted.
 //		index.Fields("name").
-//			Annotations(index.Annotation{
-//				WhereClause: "deleted_at is not null",
-//			})
+//			Annotations(entsql.Prefix(100))
 //
 //	}
 //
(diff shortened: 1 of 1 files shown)


schema/index/index.go
```diff
@@ -105,7 +105,7 @@ func (b *Builder) StorageKey(key string) *Builder {
 //	func (T) Indexes() []ent.Index {
 //
 //		// Partial index on name where the entity is not deleted.
-//		index.Field("name").
+//		index.Fields("name").
 //			Annotations(index.Annotation{
 //				WhereClause: "deleted_at is not null",
```
msal4, line 110: Should change these too but I'm not sure what to replace them with.

msal4: maybe `field.Annotation{}`?

a8m: I'll comment a suggestion.

schema/index/index.go
```diff
@@ -105,7 +105,7 @@ func (b *Builder) StorageKey(key string) *Builder {
...
 //			})
```
a8m, line 111:
```suggestion
//			Annotations(entsql.Prefix(100))
```

PR #1816, opened 2021-08-08
moved setup go envirnoment before installation

yonidavidson:
Change order in documentation to first create a go env' and then install ent.
This is meant to avoid having ent on your global pkgs.


a8m:
Thanks @yonidavidson 😃

PR #1818, opened 2021-08-09
doc/website/blog: Updated instructions to meet updated ent and elk.

masseelch:
As discussed in slack: an updated version of the blog. Please have a look at the sections about dependency issues. Some users reported about those and I added some text about it. @a8m 

cc @rotemtam

### DIFF
--- a/doc/website/blog/2021-07-29-generate-a-fully-working-go-crud-http-api-with-ent.md
+++ b/doc/website/blog/2021-07-29-generate-a-fully-working-go-crud-http-api-with-ent.md
@@ -66,7 +66,7 @@ go get -u github.com/masseelch/elk
 `elk` uses the
 Ent [extension API](https://github.com/ent/ent/blob/a19a89a141cf1a5e1b38c93d7898f218a1f86c94/entc/entc.go#L197) to
 integrate with Ent’s code-generation. This requires that we use the `entc` (ent codegen) package as
-described [here](https://entgo.io/docs/code-gen#use-entc-as-a-package). Follow the next three steps to enable it and to
+described [here](https://entgo.io/docs/code-gen#use-entc-as-a-package). Follow the next two steps to enable it and to
 configure Ent to work with the `elk` extension:
 
 1\. Create a new Go file named `ent/entc.go` and paste the following content:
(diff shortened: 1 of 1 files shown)


masseelch:
I was thinking about adding some words about easyjson and current state of ent-validation integration, what is your opinion on this?

rotemtam:
hey
the blog post should serve as an appetizer for getting into the project. it should not be much further maintained. i think any such commentary as it evolves would best fit in the `elk` README file. 
if we go through the path of adding parts/all of elk as an "official" ent extension (a-la gRPC/GraphQL) such discussions/caveats/information will belong in the relevant documentation website section.

doc/website/blog/2021-07-29-generate-a-fully-working-go-crud-http-api-with-ent.md
```diff
@@ -147,7 +139,7 @@ func (Pet) Fields() []ent.Field {
 ```
 
 We added two fields to our `Pet` entity: `name` and `age`. The `ent.Schema` just defines the fields of our entity. To
-generate runnable code from our schema, run:
+generate runnable code from our schema, run (and ignore the dependency issues for now):
```
rotemtam, line 142: this isn't clear. what dependency issue?

rotemtam (changes requested):
I'm not sure I understand what dependency issue you are referring to. 
if there is such an issue we should resolve it in a way that it doesn't happen and not tell the user to ignore it. If there is some serious issue that forces us to go through an "erroneous" state , it should be documented explicitly with the expected error message so it is searchable (ie people hitting this will find the result in Google easily).

doc/website/blog/2021-07-29-generate-a-fully-working-go-crud-http-api-with-ent.md
masseelch: If you have a look [here](https://github.com/masseelch/elk/blob/41b7573b4b2e21bcb53bd27cb540e6a27778e18f/easyjson.go#L33) after the ent and elk templates have been generated the hook executes the easyjson generator. After the generation of ent / elk the new packages (like `masseelch/render` or `mailru/easyjson` are not yet reflected in the `go.mod`. This happens in the `go run -mod=mod step`. The easyjson generator however collects information about the go ast and this requires compilation and therefore the `go generate ./...` command reports missing dependencies in the `go.mod`/`go.sum`. 

To circumvent this the user has to run the generator without easyjson, run create a main function, `go get` the dependencies, run the generator again and then start the server. 

I am clueless, how to prevent this. If you have any idea I am very happy to try it out 😊

masseelch:
:bow:

doc/website/blog/2021-07-29-generate-a-fully-working-go-crud-http-api-with-ent.md
masseelch: I changed it back and added a third step when setting up the extension: Pull in dependecies manually.

rotemtam (approved):

PR #1820, opened 2021-08-10
ent: Fixes error in CONTRIBUTING.md

masseelch:
If one would change a template in `entc/gen/template` and not run regenerate `go-bindata/bindata` contents, code generation would not take changes into account.


a8m (approved):

a8m:
Thanks @masseelch 🙏🏻

PR #1822, opened 2021-08-10
return an error when combining `DoNothing()` and `ID()` with PostgreSQL

### LINKED ISSUE #1821: DoNothing() + Upsert.ID() returns low-level SQL error in Postgres
## Current Behavior 😯

```go
id, err := entc.User.Create().
  SetEmail(email).
  OnConflict(
    sql.ConflictConstraint("user_email_key"),
  ).
  DoNothing().
  ID(ctx)
```

The above code results in an error if there's a conflict. Postgres's `RETURNING` combined with `ON CONFLICT DO NOTHING` doesn't return any rows, and the internal attempt at scanning the expected returned rows causes a low-level SQL error to leak from the ent API and be returned here.

Using `Ignore()` instead of `DoNothing()` works as expected. [Here's an example of the difference in behavior.](https://www.db-fiddle.com/f/61kemDmQ9AkqyNXbKyt1f/0)

## Expected Behavior 🤔

One of two things should happen: Either usage of Postgres as a driver should cause a call to `ID()` combined with `DoNothing()` to immediately return an error, or the error returned from the SQL should be wrapped nicely into something that can be dealt with using ent's API.

This odd edge case should also probably be documented in the `DoNothing()` and/or `ID()` godoc comments, along with a recommendation of taking a look at `Ignore()` as a potential workaround.

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          |1.16.7|
| Ent         |v0.9.0|
| Database    |Postgres 12, Postgres 13|

DeedleFake:
Fixes #1821.

### DIFF
--- a/entc/gen/template/dialect/sql/feature/upsert.tmpl
+++ b/entc/gen/template/dialect/sql/feature/upsert.tmpl
@@ -160,6 +160,10 @@ func (u *{{ $upsertOne }}) Ignore() *{{ $upsertOne }} {
 
 // DoNothing configures the conflict_action to `DO NOTHING`.
 // Supported only by SQLite and PostgreSQL.
+//
+// In PostgreSQL, this behavior prevents returning data from the same
+// query. Using DoNothing with ID will therefore result in an
+// immediate error. For a similar alternative, see Ignore.
 func (u *{{ $upsertOne }}) DoNothing() *{{ $upsertOne }} {
 	u.create.conflict = append(u.create.conflict, sql.DoNothing())
 	return u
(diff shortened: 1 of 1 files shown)


edigaryev:
Note, that it not just the `ID()` that returns [`sql.ErrNoRows`](https://pkg.go.dev/database/sql#pkg-variables), but `Exec()` too, it just discards the value but is similar otherwise.

entc/gen/template/dialect/sql/feature/upsert.tmpl
```diff
@@ -228,7 +243,16 @@ func (u *{{ $upsertOne }}) ID(ctx context.Context) (id {{ $.ID.Type }}, err erro
...
+//
+// In PostgreSQL, this method is incompatible with DoNothing, and
+// combining the two will result in an error.
 func (u *{{ $upsertOne }}) IDX(ctx context.Context) {{ $.ID.Type }} {
+	switch u.create.config.driver.Dialect() {
+	case dialect.Postgres:
+	if u.create.donothing {
+		panic(/* TODO: Return an error. */)
+	}
+
```
giautm, line 255: I think this code is unnecessary, because you really checked it in `u.ID(ctx)`

PR #1825, opened 2021-08-11
doc/md: Hide my clear name

masseelch:
:smile:


rotemtam:
its generated automatically from
https://github.com/ent/ent/blob/master/.all-contributorsrc

Issue #1830, opened 2021-08-12
OnConflict cannot get correct ID and ForUpdate return error on sqlite


godcong:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

Because sqlite3 often has lock table problems.
So I want to streamline my Insert code as much as possible.
I don't know if this is a database problem or ent needs to be fixed

## Steps to Reproduce 🕹

this is the two failed case:
1:    ForUpdate error: near "UPDATE": syntax error
```
	all, err := m.Pin.Query().ForUpdate().All(ctx)
	if err != nil {
		return nil, err
	}
```

2:    OnConflictColumns get 0 with id，
I checked the db and confirmed that the value `rid` related value has been successfully updated:
```
create := m.Pin.Create().SetPin(p)
	create.OnConflictColumns(pin.FieldRid).Update(func(upsert *ent.
		PinUpsert) {
                //do dosomthing
	})
	first, err := create.Save(ctx)
	if err != nil {
		return nil, err
	}
        //first.ID is 0

      or

	create := m.Pin.Create().SetPin(p)
	id, err := create.OnConflictColumns(pin.FieldRid).Update(func(upsert *ent.
	PinUpsert) {
		upsert.SetRid(p.Rid)
		upsert.SetStatus(p.Status)
		upsert.SetStep(p.Step)
		upsert.SetPriority(p.Priority)
		upsert.SetRelate(p.Relate)
		upsert.SetUpdatedUnix(p.UpdatedUnix)
		upsert.SetComment(p.Comment)
	}).ID(ctx)
        //id is 0
```
## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.16.6  |
| Ent         | 0.9.0   |
| Database    | sqlite3 |
| Driver      | https://github.com/mattn/go-sqlite3 1.14.8 |

rotemtam:
Hey @godcong ,

Thanks for reporting this. I'm looking into it

rotemtam:
Hi @godcong,

I am able to reproduce the first issue. Indeed, "FOR UPDATE" isn't supported in SQLite and the generated SQL is invalid in this case. I'm thinking out loud, should a special error type for unsupported featured be returned here? (cc @a8m )

Regrading your second case, I'm tried to reproduce it with a test, but I haven't been able to do so:

```go
func TestSqlLite(t *testing.T) {
[...]

PR #1847, opened 2021-08-17
dialect/sql: report explicit error when trying to SELECT .. FOR UPDAT…

rotemtam:
…E with sqlite

Related to the first issue mentioned in https://github.com/ent/ent/issues/1830


a8m (approved):
Looks good @rotemtam. Thanks for fixing this ❤️

PR #1859, opened 2021-08-20
go: upgrade to 1.17

a8m:


examples/version/ent/internal/schema.go
```diff
@@ -4,6 +4,7 @@
 
 // Code generated by entc, DO NOT EDIT.
 
+//go:build tools
```
a8m, line 7: gofmt adds an equivalent `//go:build` constraint when encountering the older syntax (`// +build ...`).
The old syntax should be removed on Go 1.18.

PR #1860, opened 2021-08-21
Goodbye bindata 😃 

a8m:


entc/gen/template.go
```diff
@@ -191,17 +189,17 @@ var (
...
+	templateDir embed.FS
 )
 
 func initTemplates() {
-	templates = NewTemplate("templates")
-	for _, asset := range internal.AssetNames() {
-		templates = MustParse(templates.Parse(string(internal.MustAsset(asset))))
-	}
+	templates = MustParse(NewTemplate("templates").
+		ParseFS(templateDir, "template/*.tmpl", "template/*/*.tmpl", "template/*/*/*.tmpl", "template/*/*/*/*.tmpl"))
```
a8m, line 202: Not a fan of it, but I prefer these static strings than to `Walk` over the fs.

PR #1861, opened 2021-08-22
fix: error message for the missing fk constraints

dakimura:
fix: a typo in the error message


a8m:
Thanks for your contribution @dakimura 🙏 


@all-contributors please add @dakimura as a contributor for code

PR #1885, opened 2021-08-29
dialect/sqlcommenter: initial work, still WIP

hedwigz:
Created a new Driver "CommenterDriver" which adds sqlcomments to every query


rotemtam (commented):
This is REALLY cool. 
But shouldn't it be published under contrib, or at the very least under a feature flag?

hedwigz:
Moved to contrib https://github.com/ent/contrib/pull/134

PR #1908, opened 2021-09-02
doc/website/blog: post about extending ent with the new Extension API

rotemtam:


doc/website/blog/2021-09-02-ent-extension-api.md
```diff
@@ -0,0 +1,280 @@
...
+name `Greet` that returns a greeting with the type's name when invoked. We're aiming for something 
+like:
+
+```go
+func (u *User) Greet() string {
+    return "Greetings, User"
+}
+```
+
+To do this, let's add a new external template file and place it in `ent/templates/greet.tmpl`:
```
yonidavidson, line 115: Maybe worth letting the reader understand where he/she can understand template language from.

yonidavidson (approved):
Excellent!

PR #1911, opened 2021-09-03
Custom predicate example for LIKE operator

sadmansakib:


a8m (approved):
Thanks for adding this @sadmansakib

PR #1924, opened 2021-09-07
entc/gen: add schema type to generated validation errors

### LINKED ISSUE #1913: Add label for validator error

## Summary 💡

I have multiple entities with field "name". On one of the entity I got this error (MaxLen is 255): `ent: validator failed for field \"name\": value is greater than the required length`. There is no way I can understand which entity this error refers to. I guess I can do better error handling on my side, but there should be no problem to add label to this errors.

## Proposed solution

Example of generated code now:
```golang
// check runs all checks and user-defined validators on the builder.
func (wc *WorkCreate) check() error {
	if v, ok := wc.mutation.Name(); ok {
		if err := work.NameValidator(v); err != nil {
			return &ValidationError{Name: "name", err: fmt.Errorf(`ent: validator failed for field "name": %w`, err)}
		}
	}
	return nil
}
```

My proposal:
```golang
// check runs all checks and user-defined validators on the builder.
func (wc *WorkCreate) check() error {
	if v, ok := wc.mutation.Name(); ok {
		if err := work.NameValidator(v); err != nil {
			return &ValidationError{Name: "name", err: fmt.Errorf(`ent: validator failed for field "name": %w`, err), Label: work.Label} // work is an entity and already imported in all of the work_create work_update and etc files.
		}
	}
	return nil
}
```
So we can add additional label field to ValidatorError and pass it to its Error method or we could inline label inside fmt.errorf

a8m:
Fixed #1913

cc @masseelch, @giautm, @maxifom

### DIFF
--- a/entc/gen/template/base.tmpl
+++ b/entc/gen/template/base.tmpl
@@ -73,7 +73,7 @@ func As(fn AggregateFunc, end string) AggregateFunc {
 	}
 {{ end }}
 
-// ValidationError returns when validating a field fails.
+// ValidationError returns when validating a field or edge fails.
 type ValidationError struct {
 	Name string // Field or edge name.
 	err error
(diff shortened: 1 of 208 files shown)


maxifom (approved):

a8m:
Thanks for the review @maxifom 🙏

Issue #1624, opened 2021-06-03
Add support for creating indexes concurrently in Postgres


BrentChesny:
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

Postgres supports [concurrent index builds](https://www.postgresql.org/docs/13/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) to avoid locking the entire table while building an index. It would be nice if there was a way to use this `CONCURRENTLY` option when migrating indexes. 

## Motivation 🔦

We recently added an index on a fairly large table, which we auto-migrated at runtime when deploying a new version. This locked the table for further writes, which was not desirable.

a8m:
Hey @BrentChesny and thanks for proposing this feature. We can add this using SQL annotation. 

I can add this in the upcoming days/week, but if you're interested in adding it, we can discuss it on Slack, and I'll be more than happy to help 🙏

a8m:
Hey, @BrentChesny and @ivanstanev 👋 
The concurrent option [was added](https://github.com/ariga/atlas/pull/1153.) by @ivanstanev a few weeks ago. You can now use [migration diff hooks](https://entgo.io/docs/migrate/#atlas-diff-and-apply-hooks) to attach this option to create new indexes concurrently. 

Closing, but please feel free to reopen or join our [Discord community](https://discord.gg/qZmPgTE6RX) and ping me if you still need help with this. Thanks 🙏

PR #1939, opened 2021-09-12
Add create index concurrently option

s-takehana:
Related to #1624.

https://www.postgresql.org/docs/14/sql-createindex.html
> Another difference is that a regular CREATE INDEX command can be performed within a transaction block, but CREATE INDEX CONCURRENTLY cannot.

https://thoughtbot.com/blog/how-to-create-postgres-indexes-concurrently-in
> ActiveRecord 4’s new disable_ddl_transaction! method must therefore be used in combination with algorithm: :concurrently migrations (as seen above).


s-takehana:
Please review.

s-takehana:
@a8m Do you want me to separate this PR?

ivanstanev:
Hey @a8m is it possible to merge this change in? I also have a good use-case for using concurrent index creation!

PR #1947, opened 2021-09-14
entc/gen: fix fk handling for bytes

tarrencev:


a8m (approved):
Thanks for fixing this @tarrencev 🙏

PR #1957, opened 2021-09-17
doc/website: fix security concern transitive depedency in website docs

RiskyFeryansyahP:
I have made some changes to the version of the transitive dependency in `doc/website`, due to security issues.

Here's a list of upgraded version of transitive dependency:

1. set-value
2. immer

Please check it, if you have time, Thanks!

cc @a8m


a8m:
Thanks for the contribution @RiskyFeryansyahP 🙏🏻

PR #1959, opened 2021-09-18
entc/gen: allow distinct in sqlQuery to enable distinct work in Select()

EndlessIdea:
We find using xxx.Query().Unique(true).Select("field1").Strings(ctx) doesn't return the expected distinct values we want.
For example, there are two records:

> id, username, city
> 1, Gary, Singapore
> 2, Jack, Singapore

The value of xxx.Query().Unique(true).Select("city").Strings(ctx) is [Singapore, Singapore], but [Singapore] is the expected return value.


a8m:
Thanks for your contribution @EndlessIdea 🙏 

This looks good. I'll wait for the CI to pass and review this.

a8m:
Thanks for your contribution @EndlessIdea and @Lining-Grab.

@all-contributors please add @EndlessIdea and @Lining-Grab for code.

EndlessIdea:
Thank you @a8m !

PR #1968, opened 2021-09-21
Unique Field Selection - Merge #1959

a8m:


a8m:
I usually don't use merge-commits, but since there are 3 authors involved in this feature, I'll merge all commits to master.

PR #1982, opened 2021-09-25
Add Sensitive() option to bytesBuilder

evanlurvey:
I have a use case where I want to store bytes but I want to use the sensitive option.


a8m (approved):
Thanks for adding this @evanlurvey. 

Would you mind adding an example field to our integration tests and run the codegen? See: https://github.com/ent/ent/blob/master/entc/integration/ent/schema/fieldtype.go#L198

evanlurvey:
@a8m I think I did what you asked. Not sure if I did the generate correctly, I ran `go generate ./...` from the `entc/intergration/` dir.

a8m:
Thanks for your contribution @evanlurvey 🙏 

@all-contributors please add @evanlurvey for code

Issue #1985, opened 2021-09-25
Upserting a row with UUID as ID results in an error


726b:
First of all let me just say THANK YOU to everyone's involved in this project.

I extremely appreciate all of your work.

- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

I use UUIDs ([Google's package](https://github.com/google/uuid)) as primary keys, for example:
````go
func (Test) Fields() []ent.Field {
	return []ent.Field{
		field.UUID(`id`, uuid.UUID{}).Unique(),
		field.Int8(`count`).Min(0),
	}
}
````

If I insert new rows, everything works as expected without errors:
````go
err = r.client.Test.
	Create().
	SetID(id).
	SetCount(1).
	Exec(r.context)
````

But, if I try to upsert rows, even new ones:
````go
err = r.client.Test.
	Create().
	SetID(id).
	SetCount(currentCount + 1).
	OnConflictColumns(test.FieldID).
	UpdateCount().
	Exec(r.context)
````

I get the following error: `interface conversion: driver.Value is string, not uuid.UUID`

## Expected Behavior 🤔

I expect a new row to be updated if the ID exists or inserted if the ID does not exist, without any error.

## Steps to Reproduce 🕹

I think the above code should suffice to reproduce the issue.

## Some Notes

I tracked the issue to a method called `sqlSave` in the generated file `test_create.go`, to the following lines:
````go
if _spec.ID.Value != nil {
	_node.ID = _spec.ID.Value.(uuid.UUID)
}
````

I _believe_ the issue is related to the combination of Ent's upsert and Google's UUID Value method.

When just inserting a new row (without `OnConflictColumns`) the type definition `_spec.ID.Type` is `field.TypeUUID`, or `[16]byte` as you defined it, and the actual type of `_spec.ID.Value` is `uuid.UUID`.

But when upserting a row, while the type definition `_spec.ID.Type` remains the same `field.TypeUUID`/`[16]byte`, the actual type of `_spec.ID.Value` is `string`.

It looks like the function `sqlgraph.CreateNode` is doing something to `_spec`, because before it is called, `_spec.ID.Value`'s type is `uuid.UUID`, and after it's run the type changes to `string`.

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.17.1  |
| Ent         | 0.9.1   |
[...]

PR #1987, opened 2021-09-26
dialect/sql/sqlgraph: support sql.Scanner types when scanning IDs

a8m:
Fixed https://github.com/ent/ent/issues/1985


dialect/sql/sqlgraph/graph.go
```diff
@@ -1301,6 +1301,11 @@ func (c *creator) insertLastID(ctx context.Context, insert *sql.InsertBuilder) e
 		}
 		defer rows.Close()
 		if !c.ID.Type.Numeric() {
+			// If the ID implements the sql.Scanner
```
masseelch, line 1304: Just out of curiosity: What about numeric types that have a custom `sql.Sanner` implementation?

a8m: Good point. The `c.ID.Value.(field.ValueScanner)` check needs to move one level up. 

Thanks 🙏

entc/gen/template/dialect/sql/create.tmpl
```diff
@@ -21,7 +21,12 @@ func ({{ $receiver }} *{{ $builder }}) sqlSave(ctx context.Context) (*{{ $.Name
 	}
 	{{- if and $.ID.UserDefined (not $.ID.Type.Numeric) }}
```
masseelch, line 22: Same on this line, isn't it?

a8m:
Thanks @masseelch for the review ❤️

PR #2001, opened 2021-10-01
dialect/sql: wraps the sql.Rows with ColumnScanner interface

a8m:
Wrapping the `*sql.Rows` comes with a little memory overhead, but it makes it easier to extend the driver and to provide customized rows scanning. I'll add more details and examples before merging this PR.


a8m:
`entcache` is going to be OSS today and this feature is needed in order to make it available for everyone. 

image

PR #2009, opened 2021-10-05
doc/gql: update generated file names

yonidavidson:


a8m (approved):
Thanks @yonidavidson ❤️

Issue #2000, opened 2021-09-30
Upserts can cause Immutable Fields to Mutate


tankbusta:
When using `UpdateNewValues` with the Upsert feature flag, fields that have been marked Immutable in the schema can be mutated. 

- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

The `created_at` column in the database is modified on upsert

## Expected Behavior 🤔

The `created_at` column in the database should stay the same

## Steps to Reproduce 🕹

Example Schema:

```
func (DeploymentStep) Fields() []ent.Field {
    return []ent.Field{
        field.UUID("id", uuid.UUID{}).
            Default(uuid.New).
            Immutable().
            Unique(),
        field.String("name"),
        field.String("as").
            Optional(),
        field.String("ref").
            Optional(),
    }
}

func (DeploymentStep) Mixin() []ent.Mixin {
    return []ent.Mixin{
        mixins.TimeMixin{}, // `created_at` lives in here. Copy of the mixing provided in the Ent Documentation
    }
}
```

Example Mutation:
```
tx.DeploymentStep.
    Create().
    SetID(step.ID.UUID).
    SetName(step.Name).
    SetAs(step.As).
    SetDeploymentID(dep.Spec.ID.UUID).
    OnConflictColumns(deploymentstep.FieldID).
    UpdateNewValues().Exec(ctx)
```

1. Run Create (row will be created). Grab `created_at` from the resulting insert 
2. Wait a few moments
3. Run Create again (1 row will exist as expected)
4. Notice row's `created_at` was modified from Step 1

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.17  |
| Ent         |  `master`  (`v0.9.2-0.20210929153101-87e21993097e`) |
| Database    | Postgres  |
| Driver      | https://github.com/go-sql-driver/mysql        |

tankbusta:
@a8m as requested in our Discord conversation.

My 2c: Immutable is immutable and `UpdateNewValues` should be aware of that. It's a very convenient function that prevents a lot of `UpdateXXX` calls to the upsert.  I'm afraid folks might not notice if `UpdateMutableValues` was added and might be expecting the correct behavior.

masseelch:
I agree to @tankbusta 💯! Nice catch! I use the API a lot and did not notice all my creation timestamps were changing 😱

[...]

PR #2011, opened 2021-10-05
entc/gen: ignore immutable fields on Upsert<T>.UpdateNewValues

a8m:
Also, for some reason, the `TimeMixin.UpdateTime` was an immutable field, but this was incorrect because the codegen just skipped generating update setters for it, and it was still get updated by the `UpdateDefault` modifier. Removing the Immutable modifier allows users to set this field explicitly.

Fixed https://github.com/ent/ent/issues/2000

### DIFF
--- a/dialect/sql/builder_test.go
+++ b/dialect/sql/builder_test.go
@@ -1711,8 +1711,8 @@ func TestInsert_OnConflict(t *testing.T) {
 	t.Run("Postgres", func(t *testing.T) { // And SQLite.
 		query, args := Dialect(dialect.Postgres).
 			Insert("users").
-			Columns("id", "email").
-			Values("1", "user@example.com").
+			Columns("id", "email", "creation_time").
+			Values("1", "user@example.com", 1633279231).
 			OnConflict(
 				ConflictColumns("email"),
 				ConflictWhere(EQ("name", "Ariel")),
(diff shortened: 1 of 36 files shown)


dialect/sql/builder_test.go
```diff
@@ -1711,22 +1711,23 @@ func TestInsert_OnConflict(t *testing.T) {
...
 					u.SetIgnore("id")
+					u.SetIgnore("creation_time")
 					u.Add("version", 1)
 				}),
 				UpdateWhere(NEQ("updated_at", 0)),
 			).
 			Query()
-		require.Equal(t, `INSERT INTO "users" ("id", "email") VALUES ($1, $2) ON CONFLICT ("email") WHERE "name" = $3 DO UPDATE SET "id" = "users"."id", "email" = "excluded"."email", "version" = COALESCE("users"."version", 0) + $4 WHERE "updated_at" <> $5`, query)
-		require.Equal(t, []interface{}{"1", "user@example.com", "Ariel", 1, 0}, args)
+		require.Equal(t, `INSERT INTO "users" ("id", "email", "creation_time") VALUES ($1, $2, $3) ON CONFLICT ("email") WHERE "name" = $4 DO UPDATE SET "id" = "users"."id", "email" = "excluded"."email", "creation_time" = "users"."creation_time", "version" = COALESCE("users"."version", 0) + $5 WHERE "updated_at" <> $6`, query)
```
tankbusta, line 1729: Just out of curiosity, why include `"id" = "users"."id"` and `"creation_time" = "users"."creation_time"` instead of omitting them? I'm assuming it's more work to remove?

a8m: Note that, this is the internal package for building the SQL statement and not the generated code. But for your question, I prefer to use explicit ignore (assign to database values), because in the case the create command contains only immutable fields, omitting them will make the UPSERT statement invalid and the ON CONFLICT clause will need to be removed. This approach just simplifies the implementation and also triggers update hooks in the database.

masseelch:
I'll have a look tomorrow.

Issue #1990, opened 2021-09-27
Hook about data table name


storyicon:
I encounter large amounts of data, and they need to be dynamically placed into different data tables by time, or hash, etc. This means that different data tables need to be dynamically selected based on the data. 
These tables look like this.
```
...
block_2312
block_2314
block_2315
...
```
I am currently looking for relevant support in the documentation (e.g. Hook) but have not been able to get anywhere. 
Any help would be appreciated.

a8m:
Hey @storyicon 👋 

I think I have a solution for you, but before that, can you please elaborate on how do you (want) choose/pass the table name? Something like passing it in a context, or do you have other ideas?

```go
client.Block.Create().
    SetA(a).
    SetB(b).
    Exec(WithTableName("block_2312"))
```

storyicon:
Hey  @a8m 
I once participated in a project called cortex, which also used a split table design, like this:
https://github.com/cortexproject/cortex/blob/master/pkg/chunk/cassandra/storage_client.go#L534-L540
I think the most intuitive way is to have a TableName method:
```
client.Block.TableName("block_2312").Create().SetA(a).SetB(b).Save(ctx)
```
At the same time, the tablename can also be calculated and modified according to the field in the hook. It seems that the tablename cannot be changed in the mutation at present.
But I'm not sure yet if this design will conflict with other features.  Context is also an option, but it may make the API less obvious.

Thanks for your quick reply ~

poonman:
I think below api is more graceful
```
client.Block.Create().Table("block_2312").SetA(a).SetB(b).Save(ctx) 
```

shenghui0779:
I want this feature.

aonoa:
I want this feature.

PR #2020, opened 2021-10-09
feat(table): support modify table name dynamically

storyicon:
Signed-off-by: storyicon <yuanchao@bilibili.com>

Because there has been no further discussion for a long time, I took the initiative to initiate this PR.
It aims to solve: https://github.com/ent/ent/issues/1990.

I think ent needs this kind of ability, which is a very common practice in large projects.

### DIFF
--- a/entc/gen/template/builder/delete.tmpl
+++ b/entc/gen/template/builder/delete.tmpl
@@ -34,6 +34,12 @@ func ({{ $receiver }} *{{ $builder }}) Where(ps ...predicate.{{ $.Name }}) *{{ $
 	return {{ $receiver }}
 }
 
+// TableName is used to dynamically modify the table name used in sql builder
+func ({{ $receiver }} *{{ $builder }}) TableName(s string) *{{ $builder }} {
+	{{ $receiver }}.mutation.tableName = s
+	return {{ $receiver }}
+}
+
 // Exec executes the deletion query and returns how many vertices were deleted.
 func ({{ $receiver}} *{{ $builder }}) Exec(ctx context.Context) (int, error) {
 	var (
--- a/entc/gen/template/builder/mutation.tmpl
+++ b/entc/gen/template/builder/mutation.tmpl
@@ -66,6 +66,8 @@ type {{ $mutation }} struct {
 	done bool
 	oldValue func(context.Context) (*{{ $n.Name }}, error)
 	predicates []predicate.{{ $n.Name }}
+
+	tableName string
 }
 
 var _ ent.Mutation = (*{{ $mutation }})(nil)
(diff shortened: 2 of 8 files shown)


poonman:
My team now encountered this issue. pls support this feature as soon as possible. thanks very much

Yiling-J:
@poonman To make create work, I think you have to do some code generation. Here is my idea:
Let's use card as example:
```go
func (cc *CardCreate) sqlSave(ctx context.Context) (*Card, error) {
	_node, _spec := cc.createSpec()
	if err := sqlgraph.CreateNode(ctx, cc.driver, _spec); err != nil {
		if sqlgraph.IsConstraintError(err) {
			err = &ConstraintError{err.Error(), err}
		}
		return nil, err
	}
	id := _spec.ID.Value.(int64)
	_node.ID = int(id)
	return _node, nil
}

func (cc *CardCreate) createSpec() (*Card, *sqlgraph.CreateSpec) {
	var (
		_node = &Card{config: cc.config}
		_spec = &sqlgraph.CreateSpec{
			Table: card.Table,
			ID: &sqlgraph.FieldSpec{
				Type:   field.TypeInt,
				Column: card.FieldID,
			},
		}
	)
...
```
`Save` will call `sqlSave`, and `sqlSave` will call `createSpec`. Here is the interesting part, you can customize createSpec:
https://github.com/ent/ent/blob/f326c7dfd088f112d0bc3150161862115d6fff63/entc/gen/template/dialect/sql/create.tmpl#L50

So if you create a template, for example:
```go
{{ define "dialect/sql/create/spec/dynamictable" }}
if {{ $receiver }}.Table != "" {
	_spec.Table = {{ $receiver }}.Table
}
{{end}}
```
Then you can customize table name. But there is still one question, 
`{{ $receiver }}` is `CardCreate`, let's take a look:
```go
type CardCreate struct {
	config
	mutation *CardMutation
	hooks    []Hook
	conflict []sql.ConflictOption
}
```
No `Table` field here, but again, you can add from external template:
https://github.com/ent/ent/blob/f326c7dfd088f112d0bc3150161862115d6fff63/entc/gen/template/builder/create.tmpl#L32

This way, you can achieve your goal from **only external templates**, no need to override ent built-in templates.

storyicon:
> My team now encountered this issue. pls support this feature as soon as possible. thanks very much

With the current solution, another problem needs to be solved: some modifications need to be made in the `columnChecker` function of `ent.go`, to prevent dynamic table names can't pass check.

I do not insist on any solution, and I look forward to further discussion on this issue.

zuobee:
.Modify(func(s *sql.Selector) {
			s.From(sql.Table(table(uid)).Schema("userdb")).

zuobee:
Modify(func(s *sql.Selector) {
			s.From(sql.Table("table_01").Schema("dbname")).

PR #2025, opened 2021-10-11
entc/gen: fix example code in comment

mookjp:
Fix example code in a comment to have the correct syntax.


a8m:
Hey @mookjp and thanks for the contribution. Can you please run `go generate ./...` from the root of the project and update the PR? Thanks

PR #2039, opened 2021-10-18
website/blog: fix typo

mattn:


a8m:
Thanks for the contribution @mattn and for the awesome SQLite driver for Go 🙏 

However, this is actually not a typo, because `ent/dialect/sql.Open` creates a `dialect.Driver` for SQL, that can be wrapped with `entcache.Driver` or any other decorator. `ent.Open` is a shortcut for `sql.Open` and `ent.NewClient(ent.Driver(drv))`.

mattn:
Ah, make sense. Thanks.

PR #2042, opened 2021-10-19
entc/gen: generate `HasPrefixFold`, `HasSuffixFold` methods for string fields

### LINKED ISSUE #1849: HasPrefixFold, HasSuffixFold methods

## Summary 💡

We have the method on string fields like <FIELD>ContainsFold. It would be nice if we could add HasPrefixFold and HasSuffixFold. 

HasPrefixFold should work like `ILIKE %s+"%s"`  in postgres and HasSuffixFold should work like `ILIKE "%" + %s`

Add HasPrefixFold and HasSuffixFold to generated code

## Motivation 🔦

I am trying to build bot, in which user send a string, which is a name in the db. But the name can or cannot be in lower or upper case (for e.g. it can be String1 or string2), and user input can be in any case. I don't want to use ContainsFold as the social network for which I am building my bot don't allow > 40 characters, so I cannot guarantee full string equality from user text message.

tdakkota:
Fixes #1849.

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -1595,6 +1595,26 @@ func (p *Predicate) Like(col, pattern string) *Predicate {
 	})
 }
 
+// likeFold is a generic helper for building LIKE predicate with case-folding.
+func (p *Predicate) likeFold(col, arg string) *Predicate {
+	return p.Append(func(b *Builder) {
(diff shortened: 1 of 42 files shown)


rubenhazelaar:
Will this functionality still be supported?

PR #2080, opened 2021-10-27
Add unparam linter

alexander-melentyev:
Remove columns arg in matchSchema method, and delete postgres unused methods

```bash
`(*Postgres).matchSchema` - `columns` always receives `nil` (unparam)
```


alexander-melentyev:
@a8m please check this PR. Thank you!

PR #2095, opened 2021-10-30
Fix docs

alexander-melentyev:


a8m (approved):
Thanks, @alexander-melentyev for fixing these typos! 🙏

PR #2112, opened 2021-11-06
dialect/sql/sqlgraph: support creating nodes without scanning their ids

a8m:
This PR will make it possible to improve the efficiency of `<T>Create.Exec` in generated code.

---

Update:  `<T>Create.Exec` is implemented currently using `Save`. Changing `Exec` to not return the `ID` field (and the created entity), will break the behavior of hooks. i.e. A hook of type `OpCreate` will return an `ent.Value(nil)` after the mutation was called.

```go
		v, err := next.Mutate(ctx, m)
		// v is nil.
		return v, err
	})
}
```

Unfortunately, need to abandon this PR as this breaking change is unacceptable.


a8m:
cc @rotemtam, @tarrencev

PR #2151, opened 2021-11-21
ent/doc: load github buttons from cdnjs

hedwigz:
Looks like loading the script from GH pages may be slow at times.  
cdnjs should be the right tool for the task.  
If this doesn't help, consider adding `async: true` (see [docs](https://docusaurus.io/docs/api/docusaurus-config#scripts))


rotemtam (approved):

rotemtam:
Thanks @hedwigz !

PR #2157, opened 2021-11-22
Fix entql generation for ID types of Bytes, etc.

adayNU:
I was running into an issue trying to use a Bytes field as the ID field in a Schema, and it appears there is currently an issue with using a `Bytes` field as the ID field in the entql codegen (it was outputting a function `WhereID(p entql.[]byteP)`).

I believe this should fix the issue (I tested this by adding the `entql` flag to the custom id integration test).

I wasn't sure if the other cases I added (e.g. an ID field that is json type / time type / etc.) are actually sane ID field types, but if those don't make any sense I can remove some of those from the added conditionals.


adayNU:
Also if there is a better place to write a test(s) for this lmk and can happily add.

a8m (approved):
Looks good @adayNU! Thanks for the contribution 🙏 ❤️

PR #2162, opened 2021-11-24
Fix migrating Postgres fields with max character varchar custom type

naormatania:
Right now if we define in postgres field with max character varchar custom type
```
field.String("name", json.RawMessage{}).
			SchemaType(map[string]string{
				dialect.Postgres: "varchar(100)",
			}),
```
then first migration works as expected.
However the second time, migration operates and column already exists in the database, the migration fails with the following error:
```
sql/schema: creating changeset for \"my_table\": changing column type for \"name\" is invalid (varchar(100) != varchar):
```

This PR fixes the issue

### DIFF
--- a/dialect/sql/schema/postgres.go
+++ b/dialect/sql/schema/postgres.go
@@ -353,9 +353,9 @@ func (d *Postgres) cType(c *Column) (t string) {
 	case field.TypeUUID:
 		t = "uuid"
 	case field.TypeString:
-		t = "varchar"
+		t = c.scanTypeOr("varchar")
 		if c.Size > maxCharSize {
-			t = "text"
+			t = c.scanTypeOr("text")
 		}
 	case field.TypeTime:
 		t = c.scanTypeOr("timestamp with time zone")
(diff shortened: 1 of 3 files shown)


dialect/sql/schema/postgres.go
```diff
@@ -229,13 +229,14 @@ const maxCharSize = 10 << 20
...
-		defaults         sql.NullString
-		udt              sql.NullString
-		numericPrecision sql.NullInt64
-		numericScale     sql.NullInt64
+		nullable                      sql.NullString
+		defaults                      sql.NullString
+		udt                           sql.NullString
+		numericPrecision              sql.NullInt64
+		numericScale                  sql.NullInt64
+		numericCharacterMaximumLength sql.NullInt64
```
a8m, line 237:
```suggestion
		characterMaximumLen sql.NullInt64
```

The `numeric` prefix is not part of the column name.

a8m (approved):
Looks good @naormatania! Thanks a lot for your contribution ❤️ 🙏 

Please, fix the small comment and I'll merge it.

dialect/sql/schema/postgres.go
naormatania: done

dialect/sql/schema/postgres_test.go
```diff
@@ -886,6 +887,64 @@ func TestPostgres_Create(t *testing.T) {
 				mock.ExpectCommit()
 			},
 		},
+		{
+			name: "no modify fixed size varchar column",
```
a8m, line 891: Thanks for adding tests for this 😍

a8m:
@all-contributors please add @naormatania for code.

PR #2164, opened 2021-11-24
Correctly generate array field nillable setters

### LINKED ISSUE #2159: No SetNillable setter is generated for array Bytes field
If you use `field.Bytes` with an array Go type (e.g. `type ByteArray [12]byte`) and set that field as optional & nillable, you would expect there to be a `SetNillable` setter generated ().

## Current Behavior 😯

Right now only a non-nil setter is generated. (In the bug example the setter is `SetUserBytes(sb schema.UserBytes)`)

## Expected Behavior 🤔

There would be a nillable setters generated (in the bug example it would be `SetNillableUserBytes(sb *schema.UserBytes)`. I _think_ the fix is simply removing `reflect.Array` from [this](https://github.com/ent/ent/blob/master/schema/field/field.go#L1162) case.

## Steps to Reproduce 🕹

See here: https://github.com/adayNU/array-bug/commit/d2af2a87ae965856183f771910f5d41105b737f7

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.17.2  |
| Ent         | 0.9.1   |
| Database    | Postgres   |
| Driver      | https://github.com/lib/pq       |

adayNU:
This should resolve #2159. I'm hesitant to do this for a `JSON` field which is an Array type, but in theory those should also be nillable?


schema/field/field.go
```diff
@@ -1159,7 +1159,7 @@ func (d *Descriptor) goType(typ interface{}, expectType reflect.Type) {
 		},
 	}
 	switch t.Kind() {
-	case reflect.Slice, reflect.Array, reflect.Ptr, reflect.Map:
```
a8m, line 1162: 👍

entc/integration/ent/schema/fieldtype.go
```diff
@@ -257,7 +257,7 @@ func (FieldType) Fields() []ent.Field { //nolint:funlen
 			GoType(role.Priority(0)),
 		field.UUID("uuid", uuid.UUID{}).
 			Optional(),
-		field.UUID("nillable_uuid", uuid.UUID{}).
+		field.UUID("uuid_nillable", uuid.UUID{}).
```
a8m, line 260: why?

entc/gen/template/builder/setter.tmpl
```diff
@@ -32,7 +32,7 @@ in the LICENSE file in the root directory of this source tree.
 	}
 
 	{{/* Avoid generting nillable setters for nillable types. */}}
-	{{ if and (not $f.Type.Nillable) (or $f.Optional $f.Default) (not (and $updater $f.UpdateDefault)) }}
+	{{ if and (not (or $f.Type.Nillable $f.Nillable)) (or $f.Optional $f.Default) (not (and $updater $f.UpdateDefault)) }}
```
a8m, line 35: Not sure about this. Can you explain? 

The idea with `and`-ing `{{ not $f.Type.Nillable }}` was to not generate `SetNillable<F>` for `nillable` values. e.g.

```
SetIntSlice(nil) // Valid, and no need for SetNillableIntSlice.
SetStructPtr(nil) // Valid, and no need for SetNillableStructPtr.
```

entc/integration/edgefield/ent/post_create.go
```diff
@@ -36,14 +36,6 @@ func (pc *PostCreate) SetAuthorID(i int) *PostCreate {
...
 }
 
-// SetNillableAuthorID sets the "author_id" field if the given value is not nil.
-func (pc *PostCreate) SetNillableAuthorID(i *int) *PostCreate {
-	if i != nil {
-		pc.SetAuthorID(*i)
-	}
-	return pc
-}
-
```
a8m, line 46: ??

PR #2165, opened 2021-11-24
docs: use go get -d for installation

ernado:
https://golang.org/doc/go-get-install-deprecation

```console 
$ go get entgo.io/ent/cmd/ent
go get: installing executables with 'go get' in module mode is deprecated.
        To adjust and download dependencies of the current module, use 'go get -d'.
        To install using requirements of the current module, use 'go install'.
        To install ignoring the current module, use 'go install' with a version,
        like 'go install example.com/cmd@latest'.
        For more information, see https://golang.org/doc/go-get-install-deprecation
        or run 'go help get' or 'go help install'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ with large data-models.
 
 ## Quick Installation
 ```console
-go get entgo.io/ent/cmd/ent
+go get -d entgo.io/ent/cmd/ent
 ```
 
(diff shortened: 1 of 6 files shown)


a8m (approved):
Thanks @ernado ❤️

Issue #2166, opened 2021-11-25
when I use uint as the field type, Update().Add<T>  only accepts the uint type


medivh-jay:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

Add a field to Fields in any schema, such as this:
```
field.Uint("inventory"),
```
and then generate the code. part of the generated code:
```
the func (<x>u *<T>Update) AddInventory(u uint) *<T>Update
```
this will not support int type parameters such as -1
## Expected Behavior 🤔

because there is a need to use numerical decrement here, it is expected to be able to use Update ().AddInventory (-1)

## Steps to Reproduce 🕹

Steps:

1. Add a field to Fields in any schema, such as this:
```
field.Uint("inventory"),
```
2. generate the code
3. call AddInventory(-1)

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.17.2  |
| Ent         | 0.9.1   |
| Database    | MySQL   |
| Driver      | https://github.com/go-sql-driver/mysql        |

a8m:
Hey @medivh-jay, and thanks for opening this issue.

This issue should be resolved, I just need to decide on how to address it. One option is to allow passing the signed version of an unsigned field. For example: `AddInventory(int)` for `field.Uint("inventory")`, and the second option is to provide another method for subtracting values. I prefer that first approach as it keeps the API without changes.

masseelch:
> Hey @medivh-jay, and thanks for opening this issue.
> 
> This issue should be resolved, I just need to decide on how to address it. One option is to allow passing the signed version of an unsigned field. For example: `AddInventory(int)` for `field.Uint("inventory")`, and the second option is to provide another method for subtracting values. I prefer that first approach as it keeps the API without changes.

Why not both? Sub can then simply call `Add<T>(-input)`.

a8m:
Because there are no `Sub` modifiers in ent builders atm. The reason is that it can be achieved with `Add(-v)` (e.g. https://pkg.go.dev/time#Time.Add).

a8m:
Issue resolved with #2169. Thanks for suggesting this @medivh-jay 🙏

medivh-jay:
Thank you very much for your prompt reply. 😄 @a8m

PR #2169, opened 2021-11-25
entc/gen: add support for subtracting values from unsgined fields

a8m:
Fixed https://github.com/ent/ent/issues/2166


masseelch (approved):
I'd add a test for the required variant. Other than that LGTM :-)

a8m:
Thanks for the review @masseelch

Issue #2170, opened 2021-11-26
Feature Request: Set `imports.LocalPrefix` when sort imports


Thearas:
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

Sames as `goimports -local "<my-pkg>/ent"`.

## Motivation 🔦

Current:

```go
import (
	"context"
	"fmt"

	"entgo.io/ent/dialect/sql"
	"entgo.io/ent/dialect/sql/sqlgraph"
	"entgo.io/ent/schema/field"
	"<my-pkg>/ent/internal"
	"<my-pkg>/ent/predicate"
	"<my-pkg>/ent/testbed"
)
```

What I want:

```go
import (
	"context"
	"fmt"

	"entgo.io/ent/dialect/sql"
	"entgo.io/ent/dialect/sql/sqlgraph"
	"entgo.io/ent/schema/field"

	"<my-pkg>/ent/internal"
	"<my-pkg>/ent/predicate"
	"<my-pkg>/ent/testbed"
)
```

We could set `imports.LocalPrefix` to `graph.Config.Package`.

a8m:
Hey @Thearas, and thanks for the suggestion!

TBH, I don't think a new flag or option is needed, because the same functionality can be achieved today. If you're using the `ent/entc.go`-based config option, the `LocalPrefix` can be set before running codegen:

```go
//go:build ignore

package main

// ...

func main() {
	imports.LocalPrefix = "<config-goes-here>"
	// ...
	err = entc.Generate("./schema", &gen.Config{}, opts...)
	if err != nil {
		log.Fatalf("running ent codegen: %v", err)
	}
}
```

If you're using the command line version (that is executed by `go generate`), you can just add another line for executing `goimports` after `ent generate`. 

```go
package ent

//go:generate go run -mod=mod entgo.io/ent/cmd/ent generate ./schema
//go:generate go run -mod=mod golang.org/x/tools/cmd/goimports -local <config-goes-here> -w .
```

masseelch:
> ```go
> //go:build ignore
> 
> package main
> 
> // ...
> 
> func main() {
> 	imports.LocalPrefix = "<config-goes-here>"
> 	// ...
> 	err = entc.Generate("./schema", &gen.Config{}, opts...)
> 	if err != nil {
> 		log.Fatalf("running ent codegen: %v", err)
> 	}
> }
> ```

[...]

PR #2171, opened 2021-11-26
sort imports using LocalPrefix

Thearas:
Implement https://github.com/ent/ent/issues/2170.

The second commit is generated by `go generate ./...`.

### DIFF
--- a/entc/gen/graph.go
+++ b/entc/gen/graph.go
@@ -753,7 +753,6 @@ func (a assets) write() error {
 // format runs "goimports" on all assets.
 func (a assets) format(localPrefix string) error {
 	imports.LocalPrefix = localPrefix
-
 	for _, file := range a.files {
 		path := file.path
 		src, err := imports.Process(path, file.content, nil)
--- a/entc/gen/graph.go
+++ b/entc/gen/graph.go
@@ -231,7 +231,7 @@ func generate(g *Graph) error {
 	// We can't run "imports" on files when the state is not completed.
 	// Because, "goimports" will drop undefined package. Therefore, it's
 	// suspended to the end of the writing.
-	return assets.format()
+	return assets.format(g.Config.Package)
 }
 
 // addNode creates a new Type/Node/Ent to the graph.
@@ -751,7 +751,9 @@ func (a assets) write() error {
 }
 
 // format runs "goimports" on all assets.
-func (a assets) format() error {
+func (a assets) format(localPrefix string) error {
+	imports.LocalPrefix = localPrefix
+
 	for _, file := range a.files {
 		path := file.path
 		src, err := imports.Process(path, file.content, nil)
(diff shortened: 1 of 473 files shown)


entc/gen/graph.go
```diff
@@ -751,7 +751,9 @@ func (a assets) write() error {
 }
 
 // format runs "goimports" on all assets.
-func (a assets) format() error {
+func (a assets) format(localPrefix string) error {
+	imports.LocalPrefix = localPrefix
+
```
masseelch, line 756:
```suggestion
```

masseelch (approved):
I like very much. Let's see what @a8m thinks.

a8m:
> I like very much. Let's see what @a8m thinks.

Thanks for your contribution @Thearas, and for your review @masseelch ❤️ 

Please, see my comment on https://github.com/ent/ent/issues/2170#issuecomment-980009054, and let's continue the discussion there.

entc/integration/gremlin/ent/fieldtype.go
```diff
@@ -17,8 +17,9 @@ import (
 	"entgo.io/ent/dialect/gremlin"
 	"entgo.io/ent/entc/integration/ent/role"
 	"entgo.io/ent/entc/integration/ent/schema"
-	"entgo.io/ent/entc/integration/gremlin/ent/fieldtype"
 	"github.com/google/uuid"
+
+	"entgo.io/ent/entc/integration/gremlin/ent/fieldtype"
```
a8m, line 22: In such cases, the generated package imports a "local package". Therefore, the grouping should be based on the Go modules (project root). In this case, it's `entgo.io/ent`. i.e. if we have 2 import paths:
```go
<root-pkg>/ent/user
<root-pkg>/pkg/types
```

We want to set `-local` to `<root-pkg>`, and not `<root-pkg>/ent`, right?

masseelch: Ahm, yeah. `goimports` needs an option to add a 4th section :-D

Thearas: Yeah, I agree😁. How to get the `<root-pkg>`?

Thearas: Another idea, `entc` will set `LocalPrefix` to `<root-pkg>` only if the `LocalPrefix` is empty.

a8m: We can add some code for loading the Go module information (see my PR - https://github.com/ent/ent/pull/2175) and use it as a LocalPrefix, but I agree it should be set only if `LocalPrefix` is empty. Let me know what do you think.

a8m: `Module.Path` should return the go module path. But again, tbh, I think there are no real guidelines on how package paths should be grouped. I, for example, like to group all my company/product import paths in the same block. Thoughts?

Thearas: I agree. I have already fixed my problem with `imports.LocalPrefix = "<config-goes-here>"`. This PR could be closed😂.

Thearas: Thanks for your solution @a8m !

PR #2183, opened 2021-11-30
update faq.md, add new question

idc77:
question: How to change the character set and/or collation of a MySQL table?

Not sure if the slug path is correct.

### DIFF
--- a/doc/md/faq.md
+++ b/doc/md/faq.md
@@ -763,7 +763,7 @@ func (d *multiDriver) Close() error {
 #### How to change the character set and/or collation of a MySQL table?
 
 By default for MySQL the character set `utf8mb4` is used and the collation of `utf8mb4_bin`.
-However if you'd like to change the schema's character set and/or collaction you will need to use the annotation feature.
+However if you'd like to change the schema's character set and/or collation you need to use an annotation.
 
 Here's an example where we set the character set to `ascii` and the collation to `ascii_general_ci`.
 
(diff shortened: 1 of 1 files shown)


doc/md/faq.md
```diff
@@ -759,3 +760,22 @@ func (d *multiDriver) Close() error {
 }
 ```
 
+#### How to change the character set and/or collation of a MySQL table?
+
+By default for MySQL the character set `utf8mb4` is used and the collation of `utf8mb4_bin`.
+However if you'd like to change the schema's character set and/or collaction you will need to use the annotation feature.
```
masseelch, line 766:
```suggestion
However if you'd like to change the schema's character set and/or collation you need to use an annotation.
```

masseelch (approved):
Thanks!

doc/md/faq.md
```diff
@@ -22,6 +22,7 @@ sidebar_label: FAQ
 [How to add `CHECK` constraints to table?](#how-to-add-check-constraints-to-table)  
 [How to define a custom precision numeric field?](#how-to-define-a-custom-precision-numeric-field)  
 [How to configure two or more `DB` to separate read and write?](#how-to-configure-two-or-more-db-to-separate-read-and-write)
```
a8m, line 24:
```suggestion
[How to configure two or more `DB` to separate read and write?](#how-to-configure-two-or-more-db-to-separate-read-and-write)  
```

a8m: A line break is needed here.

a8m:
Thanks for your contribution @idc77 🙏 ❤️ 

@all-contributors please add @idc77 for docs.

PR #2191, opened 2021-12-01
dialect/schema: support float for mysql

yonidavidson:


a8m:
Thanks for working on it @yonidavidson. Can you please add a test case for it? 😃

PR #2211, opened 2021-12-10
entc/gen: don't rely on descriptor when retrieving enum values or nam…

masseelch:
…es to ease extension testibility

Trying to add tests to some extensions and this is complicating things a lot because `Field.def` is unexported. Not taking the length of `def.Enums` but `Enums` directly fixes this. It technically is also more correct, because `f.Enums` could be greater than `f.def.Enums`.


a8m (approved):
Thanks @masseelch 💯

Issue #2209, opened 2021-12-09
modernc.org/sqlite driver fails to migrate after the first migration


plunkettscott:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

When using the the [modernc.org/sqlite](https://pkg.go.dev/modernc.org/sqlite) driver, the first SQLite `Schema.Create()` execution works properly and migrates the database. However, the second (and subsequent) times it fails on INTEGER columns, providing the following error message (with no changes to the schema between executions):

```
creating changeset for "discord_roles": invalid type "INTEGER" for column "color"
```

## Expected Behavior 🤔

Subsequent runs of `Schema.Create()` on startup should not produce an error, like the [github.com/mattn/go-sqlite3](https://pkg.go.dev/github.com/mattn/go-sqlite3) driver.

## Steps to Reproduce 🕹

https://github.com/plunkettscott/ent-sqlite-migration-bug

Steps:

1. Run `go test ./...`

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.17.3  |
| Ent         | 0.9.1   |
| Database    | SQLite   |
| Driver      | [modernc.org/sqlite](https://pkg.go.dev/modernc.org/sqlite)       |

plunkettscott:
Just an update to this, I dropped it down to `modernc.org/sqlite` @v1.14.1 (which uses SQLite 3.36.0) and this works correctly. It looks like a potential problem with SQLite v3.37.0 in Ent.

a8m:
Thanks for reporting this @plunkettscott, I was able to reproduce this and found the issue.

It looks like SQLite changed the `table_info` output.

v3.36.0:

![image](https://user-images.githubusercontent.com/7413593/145677154-b38ed9bf-6e49-40a2-a06d-75176883c1a0.png)

v3.37.0:

![image](https://user-images.githubusercontent.com/7413593/145677166-07fba6a0-fe53-40f7-b04a-dc3cc8c3cf51.png)

We can easily support this if we [lowercase the type](https://github.com/ent/ent/blob/master/dialect/sql/schema/sqlite.go#L292) before we test it. Would you like to contribute a fix? Thanks 🙏

m3hm3t:
I could not test with [repo](https://github.com/plunkettscott/ent-sqlite-migration-bug).

If you share me how, I can do that.

[...]

PR #2222, opened 2021-12-18
convert column type string to lower case

m3hm3t:
PR for issue [2209](https://github.com/ent/ent/issues/2209)

### DIFF
--- a/dialect/sql/schema/sqlite.go
+++ b/dialect/sql/schema/sqlite.go
@@ -289,7 +289,6 @@ func (d *SQLite) scanColumn(c *Column, rows *sql.Rows) error {
 	if err != nil {
 		return err
 	}
-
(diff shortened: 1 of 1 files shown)


dialect/sql/schema/sqlite.go
```diff
@@ -289,7 +289,8 @@ func (d *SQLite) scanColumn(c *Column, rows *sql.Rows) error {
 	if err != nil {
 		return err
 	}
-	switch parts[0] {
+
```
a8m, line 292: rm redundant \n

a8m (approved):
Thanks for the contribution @m3hm3t 🙏🏻

dialect/sql/schema/sqlite.go
a8m, line 292:
```suggestion
```

PR #2251, opened 2022-01-05
fix(entc/gen): there was a typo in Tx interfaces

masseelch:
PR fixes a typo in the generate GoDoc for the transaction interfaces


a8m (approved):
👍

PR #2254, opened 2022-01-09
dialect/sql/schema: do not sort primary-key columns before diff

a8m:
PKs are indexes and therefore, the order does matter.
Thanks to ariga.io/atlas we found this issue.

### DIFF
--- a/dialect/sql/schema/migrate.go
+++ b/dialect/sql/schema/migrate.go
@@ -9,7 +9,6 @@ import (
 	"crypto/md5"
 	"fmt"
 	"math"
-	"sort"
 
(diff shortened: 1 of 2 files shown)


yonidavidson (approved):

a8m:
Merging this PR even though the "migration" check (in CI) is falling, because (join) tables are always created with the same order. However, when running multiple tests (with multiple databases) in the CI, the previous logic was changing the order of the generated schema (in `ent/migrate.Tables`) so future tests were creating the tables in a different order, and this caused the tests to fail due to mismatched primary-key order.

PR #2258, opened 2022-01-10
add support for reading into, storing in, and querying using a specific timezone

rubensayshi:
Opening this PR to discuss this as a suggestion, if you feel this is going in a direction you could consider merging then I'll add tests as well!

The use case is: we strongly prefer storing all time in UTC in our database, but it's a huge pain having to make sure we do `.In(time.UTC)` everywhere before storing data and very error prone.  

Even worse is querying, since you need to make sure that if the time in the database is stored in UTC that you're querying using a `time.Time` which is in UTC as well... 

This PR solves both of these issues by introducing two runtime configurable options: `StoreInTz` and `ReadIntoTz`

`StoreInTz` will ensure any timestamps in mutations will be set to that timezone, and any querying done will use that timezone.

`ReadIntoTz` will ensure any timestamps read from the database are converted into the desired timezone.

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	"entgo.io/ent/dialect"
 )
@@ -2892,10 +2893,11 @@ func (n Queries) Query() (string, []interface{}) {
 type Builder struct {
 	sb        *strings.Builder // underlying builder.
 	dialect   string           // configured dialect.
-	args      []interface{}    // query parameters.
-	total     int              // total number of parameters in query tree.
-	errs      []error          // errors that added during the query construction.
-	qualifier string           // qualifier to prefix identifiers (e.g. table name).
+	tz        *time.Location
+	args      []interface{} // query parameters.
+	total     int           // total number of parameters in query tree.
+	errs      []error       // errors that added during the query construction.
+	qualifier string        // qualifier to prefix identifiers (e.g. table name).
 }
 
 // Quote quotes the given identifier with the characters based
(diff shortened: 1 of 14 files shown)


dialect/sql/builder.go
```diff
@@ -2892,10 +2893,11 @@ func (n Queries) Query() (string, []interface{}) {
 type Builder struct {
 	sb        *strings.Builder // underlying builder.
 	dialect   string           // configured dialect.
-	args      []interface{}    // query parameters.
-	total     int              // total number of parameters in query tree.
-	errs      []error          // errors that added during the query construction.
-	qualifier string           // qualifier to prefix identifiers (e.g. table name).
+	tz        *time.Location
```
rubensayshi, line 2896: I couldn't find a better place to put this to have it accessible in the predicates, but I think it could be considered similar to `dialact` so it's not a terrible place to put this

entc/integration/ent/card.go
```diff
@@ -115,13 +115,23 @@ func (c *Card) assignValues(columns []string, values []interface{}) error {
 			if value, ok := values[i].(*sql.NullTime); !ok {
 				return fmt.Errorf("unexpected type %T for field create_time", values[i])
 			} else if value.Valid {
-				c.CreateTime = value.Time
+				if c.config.readIntoTz != nil {
+					c.CreateTime = value.Time.In(c.config.readIntoTz)
+				} else {
+					c.CreateTime = value.Time
+				}
+
```
rubensayshi, line 123: could replace this with a `InReadIntoTz()` helper or something to keep it simple here and move the if into the `config` ?

masseelch:
Why don't you set the timezone to the underlying database connection?

rubensayshi:
hmm, I guess that does cover half the problem...

I actually want to store (and query) in UTC but have the values read from the database in my own timezone

PR #2259, opened 2022-01-10
read time.Time into configured timezone

rubensayshi:
simpler version of https://github.com/ent/ent/pull/2258 after @masseelch pointed out that you can set `timezone=UTC` (postgres) or `loc=UTC` (mysql) to ensure that timestamps are stored in UTC and that when querying your passed in timestamp is converted to UTC.

only thing left now are the changes which convert a returned timestamp into a configured timezone, this would allow having a different / specific timezone returned while keeping UTC in the database (which afaik is considered best practice).

### DIFF
--- a/entc/gen/template/config.tmpl
+++ b/entc/gen/template/config.tmpl
@@ -49,6 +49,16 @@ type config struct {
 			{{- end }}
 		{{- end }}
 	{{- end }}
+
+	readIntoTz *time.Location
+}
+
+// timeInTz will convert the read time.Time into the configured timezone (if configured).
+func (c *config) timeInTz(t time.Time) time.TIme {
+	if c.readIntoTz != nil {
+		return t.In(c.readIntoTz)
+	}
+	return t
 }
 
 // hooks per client, for fast access.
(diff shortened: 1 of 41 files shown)


masseelch:
Hey @rubensayshi, me again :-) 

I am still not sure if this is really needed. Storing timestamps in UTC is indeed a good idea but the `timezone` and `loc` options of the driver already parse timezone. 

Displaying the date in the your local timezone should be done in your "view", whatever that might be.

rubensayshi:
In general it's nice and convenient that you don't have to convert it to a timezone anymore at the edge (view) in systems where you just always want to display in a single timezone, but I kinda agree that it's often just better to convert it at the edge since then you can potentially display it in the user's desired timezone.  


But I think the context in which this becomes actually useful / important is in a backend system which does a lot of processing etc.  
Like for example a billing system, where there isn't a view / edge to convert to the user's timezone but instead all the timestamps you're dealing with need to be in a very specific timezone.

masseelch:
I guess we'll have to hear @a8m opinion on this matter.

PR #2269, opened 2022-01-13
add WhereIf API for conditional query

gitive:
Usually, we are writing some services may facing the conditional query, such as:
```go
query := s.client.UserAccount.Query().Where(useraccount.TenantId(1))
if withPositiveOnly {
	query.Where(useraccount.MoneyGT(0))
}
users, err := query.All(s.ctx)
```
It's very inconvenient, in many cases with more then three conditions.
It may be graceful, if it can be wrote like this:
```go
users, err := s.client.UserAccount.Query().Where(useraccount.TenantId(1)).WhereIf(withPositiveOnly, useraccount.MoneyGT(0)).All(s.ctx)
```
It's very clear and concise.
Think you.

### DIFF
--- a/entc/gen/template/builder/delete.tmpl
+++ b/entc/gen/template/builder/delete.tmpl
@@ -34,6 +34,14 @@ func ({{ $receiver }} *{{ $builder }}) Where(ps ...predicate.{{ $.Name }}) *{{ $
 	return {{ $receiver }}
 }
 
+// WhereIf appends a list predicates to the {{ $builder }} builder if b is true.
+func ({{ $receiver }} *{{ $builder }}) WhereIf(b bool, ps ...predicate.{{ $.Name }}) *{{ $builder }} {
+	if b {
+	    {{ $mutation }}.Where(ps...)
+	}
+	return {{ $receiver }}
+}
+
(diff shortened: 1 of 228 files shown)


gitive (commented):
three files changed:
entc/gen/template/builder/delete.tmpl
entc/gen/template/builder/query.tmpl
entc/gen/template/builder/update.tmpl
and the others are gen by ent.

gmhafiz:
A test can be added to. For example

`integration_test.go`

```go

func Conditional(t *testing.T, client *ent.Client) {
	ctx := context.Background()

	builders := []*ent.UserCreate{
		client.User.Create().SetName("A").SetAge(1).SetPhone("0000"), // Duplicate
		client.User.Create().SetName("B").SetAge(1).SetPhone("1111"), // New row.
	}
	client.User.CreateBulk(builders...).
		OnConflictColumns(user.FieldPhone).
		UpdateNewValues().
		ExecX(ctx)

	users, err := client.User.Query().
		WhereIf(true, func(b *ent.UserQuery) {
			b.
				Limit(1).
				Offset(1)
		}).
		All(ctx)
	require.Nil(t, err)
	require.Equal(t, len(users), 1)
}
```

PR #2270, opened 2022-01-14
[migrate]: allow to use external ent_types. [WIP]

giautm:


giautm:
I need feedback for naming the EntTypesStore interface and its methods.

a8m:
Thanks for the contribution @giautm. I'll give it a look later, but please note that this change may conflict with my changes in https://github.com/ent/ent/tree/atlas. We can talk to arrange this.

giautm:
The idea is just allow to more control how to generate Universel ID when running migration.

PR #2274, opened 2022-01-18
fix: update only/onlyID comments

iamnande:
this patch updates the Only/OnlyID methods in the builder template from:

> // Returns a *NotSingularError when exactly one {{ $.Name }} entity is not found.
to
> // Returns a *NotSingularError when more than exactly one {{ $.Name }} ID is found.

in an effort to be more explicit in the explanation of the behavior of the methods.

I caught myself reading the difference between the lines more than a few times just to make sure I truly understood the behavior. Maybe others agree?

### DIFF
--- a/entc/gen/template/builder/query.tmpl
+++ b/entc/gen/template/builder/query.tmpl
@@ -144,7 +144,7 @@ func ({{ $receiver }} *{{ $builder }}) FirstIDX(ctx context.Context) {{ $.ID.Typ
 }
 
 // Only returns a single {{ $.Name }} entity found by the query, ensuring it only returns one.
-// Returns a *NotSingularError when more than exactly one {{ $.Name }} entity is found.
+// Returns a *NotSingularError when more than one {{ $.Name }} entity is found.
 // Returns a *NotFoundError when no {{ $.Name }} entities are found.
 func ({{ $receiver }} *{{ $builder }}) Only(ctx context.Context) (*{{ $.Name }}, error) {
 	nodes, err := {{ $receiver }}.Limit(2).All(ctx)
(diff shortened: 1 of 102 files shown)


entc/gen/template/builder/query.tmpl
```diff
@@ -171,7 +171,7 @@ func ({{ $receiver }} *{{ $builder }}) OnlyX(ctx context.Context) *{{ $.Name }}
 }
 
 // OnlyID is like Only, but returns the only {{ $.Name }} ID in the query.
-// Returns a *NotSingularError when exactly one {{ $.Name }} ID is not found.
+// Returns a *NotSingularError when more than exactly one {{ $.Name }} ID is found.
```
rotemtam, line 174:
```suggestion
// Returns a *NotSingularError when more than one {{ $.Name }} ID is found.
```

entc/gen/template/builder/query.tmpl
```diff
@@ -144,7 +144,7 @@ func ({{ $receiver }} *{{ $builder }}) FirstIDX(ctx context.Context) {{ $.ID.Typ
 }
 
 // Only returns a single {{ $.Name }} entity found by the query, ensuring it only returns one.
-// Returns a *NotSingularError when exactly one {{ $.Name }} entity is not found.
+// Returns a *NotSingularError when more than exactly one {{ $.Name }} entity is found.
```
hilakashai, line 147:
```suggestion
// Returns a *NotSingularError when zero or more than one {{ $.Name }} entity is found.
```

entc/gen/template/builder/query.tmpl
hilakashai, line 174:
```suggestion
// Returns a *NotSingularError when zero or more than one {{ $.Name }} ID is found.
```

a8m: Thanks for the review @rotemtam. @iamnande, Can you please fix the PR and run codegen?

iamnande:
i was able to run `go generate ./...` just fine. 100% of the files committed in [7111fab](https://github.com/ent/ent/pull/2274/commits/7111fabfff515515478259313935c0d825f2e894) are the result of running it.

a8m:
Thanks for your contribution @iamnande 🙏 ❤️ 

@all-contributors please add @iamnande for code

PR #2294, opened 2022-01-25
ci: bump golangci-lint version

sashamelentyev:


a8m:
Thanks for the contribution @sashamelentyev

Issue #1631, opened 2021-06-09
Support descending index


ColetteContreras:
```sql
CREATE TABLE t (
  c1 INT, c2 INT,
  INDEX idx1 (c1 ASC, c2 ASC),
  INDEX idx2 (c1 ASC, c2 DESC),
  INDEX idx3 (c1 DESC, c2 ASC),
  INDEX idx4 (c1 DESC, c2 DESC)
);
```

- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

```go
func (MyTable) Indexes() []ent.Index {
	return []ent.Index{
		index.Fields("c1 DESC", "c2 ASC"),
	}
}

OR

func (MyTable) Indexes() []ent.Index {
        directionalIndex1 := map[string]string{ "c1": "DESC", "c2", "ASC" }
	return []ent.Index{
		index.DirectionalFields(directionalIndex1),
	}
}
```

## Motivation 🔦

I have a table which ID type is not int, but UUID. I'll use `ORDER BY time DESC` to retrieve latest N records. It says my sorting is `Backward index scan` when I try to optimize my query using `EXPLAIN` statement. After doing a little research I realized that the index which created  by `ent` is an **ascending** index, but what I really needed is descending index. It'll be slower than **descending** index approximately 15%.

a8m:
Hey @ColetteContreras and thanks for proposing this.
We should add support for descending indexes. I just added an annotation option for indexes (in https://github.com/ent/ent/pull/1632), and this should make it simple to support this option. 

cc @rotemtam

a8m:
Thanks for proposing this issue @ColetteContreras, it was resolved with https://github.com/ent/ent/pull/2304.

You can define descending indexes if you're using the [Atlas migration](https://entgo.io/docs/migrate#atlas-integration). The API is as follows:

```go
index.Fields("a").
	Annotation(entsql.Desc())
// CREATE INDEX ... ON "table" ("a" DESC)

index.Fields("a", "b", "c").
	Annotation(entsql.DescColumns("a", "b"))
// CREATE INDEX ... ON "table" ("a" DESC, "b" DESC, "c")
```

badsector998:
> Thanks for proposing this feature @ColetteContreras, it was resolved with #2304.
> 
> You can define descending indexes if you're using the [Atlas migration](https://entgo.io/docs/migrate#atlas-integration). The API is as follows:
> 
> ```go
[...]

PR #2304, opened 2022-01-31
sql: add support for DESC indexes using annotation and atlasgo.io

a8m:
Fixed https://github.com/ent/ent/issues/1631

### DIFF
--- a/dialect/entsql/annotation.go
+++ b/dialect/entsql/annotation.go
@@ -202,6 +202,29 @@ type IndexAnnotation struct {
 	//	CREATE INDEX `table_c1_c2_c3` ON `table`(`c1`(100), `c2`(200), `c3`)
 	//
 	PrefixColumns map[string]uint
+
+	// Desc defines the DESC clause for a single column index.
+	// In MySQL, the following annotation maps to:
+	//
+	//	index.Fields("column").
+	//		Annotation(entsql.Desc())
+	//
+	//	CREATE INDEX `table_column` ON `table`(`column` DESC)
+	//
+	Desc bool
+
+	// DescColumns defines the DESC clause for columns in a multi column index.
+	// In MySQL, the following annotation maps to:
+	//
+	//	index.Fields("c1", "c2", "c3").
+	//		Annotation(
+	//			entsql.DescColumn("c1"),
+	//			entsql.DescColumn("c2"),
+	//		)
+	//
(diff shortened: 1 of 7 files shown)


dialect/entsql/annotation.go
```diff
@@ -202,6 +202,29 @@ type IndexAnnotation struct {
...
+	//
+	//	index.Fields("c1", "c2", "c3").
+	//		Annotation(
+	//			entsql.DescColumn("c1"),
+	//			entsql.DescColumn("c2"),
+	//		)
+	//
+	//	CREATE INDEX `table_c1_c2_c3` ON `table`(`c1` DESC, `c2` DESC, `c3`)
+	//
+	DescColumns map[string]bool
```
rotemtam, line 227: isn't order lost by using a map?

rotemtam (commented):
looks good, see question

dialect/entsql/annotation.go
a8m: The order has no meaning here as I use `DescColumns[c.Name]` to check if a `DESC` option was defined (not looping over it).

dialect/entsql/annotation.go
```diff
@@ -202,6 +202,28 @@ type IndexAnnotation struct {
...
+	// In MySQL, the following annotation maps to:
+	//
+	//	index.Fields("c1", "c2", "c3").
+	//		Annotation(
+	//			entsql.DescColumns("c1", "c2"),
+	//		)
+	//
+	//	CREATE INDEX `table_c1_c2_c3` ON `table`(`c1` DESC, `c2` DESC, `c3`)
+	//
+	DescColumns map[string]bool
```
masseelch, line 226: The more "goish" (and in my opinion cleaner way) is going with `map[string]struct{}`. Not really sure though, the check adds some boilerplate to it. 

Will this be used by the user directly? Looks like it is unlikely.

masseelch (approved):
Minor comment.

dialect/entsql/annotation.go
a8m: You still have the option to add annotations like:
```go
&entsql.IndexAnnotation{
	Prefix: 10,
	Desc: true,
}
```

a8m:
Thanks @rotemtam and @masseelch for the review ❤️

PR #2307, opened 2022-02-02
Add check that model file doesn't already exist on ent init

adayNU:
Previously if you ran `ent init` it would overwrite an existing file  f that already existed. Now  checks if that file already exists, and returns an error if so.


a8m (approved):
Looks great. Thanks for the contribution @adayNU 🙏🏻❤️

Issue #2290, opened 2022-01-24
GoType on field named "id" 


lynkas:
I'm trying to override default id field with own type called "ID"

```
func (IDMixin) Fields() []ent.Field {
	return []ent.Field{
		field.Int64("id").GoType(sid.ID("")).DefaultFunc(sid.New).Unique().Immutable(),
	}
}

```
where sid.ID is a type implemented ValueScanner interface ( simplified ):

```

type ID string

func (i ID) Value() (driver.Value, error) {

	return i.ToInt64()
}

func (i *ID) Scan(src interface{}) error {
	switch v := src.(type) {
	case nil:
		return nil
	case int64:
		*i = ID(Int64ToString(v))
		return nil
	}
	return errors.New("not a valid base62")
}

func New() ID {
	return NewLength(10)
}
func NewLength(len int) ID {
	return ID(utils.RandomString(len))
}
```
After generation I found there is an error in table_create.go, which located in ``func (acb *<table>CreateBulk) Save(ctx context.Context) ([]*<table>, error) `` function :

```
mutation.id = &nodes[i].ID
mutation.done = true
if specs[i].ID.Value != nil && nodes[i].ID == 0 {      <-error here
	id := specs[i].ID.Value.(int64)
	nodes[i].ID = sid.ID(id)
}
return nodes[i], nil
```
the error is `Invalid operation: nodes[i].ID == 0 (mismatched types sid.ID and untyped int)`

Did I misuse this feature, like the id field cannot be override with custom types?

a8m:
Hey @lynkas, what ent version do you use? Can you check if you get this on master? Use `go get entgo.io/ent@master` to update.

lynkas:
@a8m I was using v0.9.1. It still occurs on master.

crossworth:
Hello @lynkas, could you try using the `field.Other` as on this example using the `master` version:
https://github.com/ent/ent/blob/a7beb5c834f0b81de7afd15350da9c42d8eb6b88/entc/integration/customid/ent/schema/other.go#L22-L29

lynkas:
@crossworth Thank you for your contribution! It works perfectly on an individual table. However some problems occurs when it queries a foreign key.
![image](https://user-images.githubusercontent.com/22864418/153693600-f9b0c164-03b3-4a8c-9f55-32490eaaeeda.png)
It tried to cast sid input type (a string) to sid store type (int64), and error occurred. 

[...]

PR #2309, opened 2022-02-03
Allow ID to use field.Other with custom types

crossworth:
When using field.Other the generated code would cause problems on two instances:

https://github.com/crossworth/entissue2290/blob/042ee692b298775220878f08082e6592575eef58/ent/user_create.go#L231
image

https://github.com/crossworth/entissue2290/blob/042ee692b298775220878f08082e6592575eef58/ent/user.go#L56
image
Line 56: `conversion from int64 to ID (string) yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)`

I'm not sure if this is the best way to solve the problem, but I think we can assume that when the user use a `field.Other` it would not be `~int64`.

This **WILL NOT** fix the problem when using `GoType`. The same problem happens, but I don't know how to solve, maybe a better approach would be to check the type.

Partial fix for: #2290


a8m (approved):
Amazing @crossworth ❤️ 

Thanks for the contribution!

a8m:
@all-contributors please add @crossworth for code

Issue #2331, opened 2022-02-15
Missing unique field in query Clone()


MrParano1d:
The `unique` Field from a query builder is ignored when the query gets cloned.

- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

The query template doesn't clone the `unique` field from the query.

```
// Clone returns a duplicate of the {{ $builder }} builder, including all associated steps. It can be
// used to prepare common query builders and use them differently after the clone is made.
func ({{ $receiver }} *{{ $builder }}) Clone() *{{ $builder }} {
	if {{ $receiver }} == nil {
		return nil
	}
	return &{{ $builder }}{
		config: 	{{ $receiver }}.config,
		limit: 		{{ $receiver }}.limit,
		offset: 	{{ $receiver }}.offset,
		order: 		append([]OrderFunc{}, {{ $receiver }}.order...),
		predicates: append([]predicate.{{ $.Name }}{}, {{ $receiver }}.predicates...),
		{{- range $e := $.Edges }}
			{{ $e.EagerLoadField }}: {{ $receiver }}.{{ $e.EagerLoadField }}.Clone(),
		{{- end }}
		// clone intermediate query.
		{{ $.Storage }}: {{ $receiver }}.{{ $.Storage }}.Clone(),
		path: {{ $receiver }}.path,
	}
}
```

## Expected Behavior 🤔

When I do a `client.SomeModel.Query().Unique(false).Clone()` `unique` is `true` again.

It should be `false`.

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.17.2  |
| Ent         | 0.10.0   |
| Database    | PostgreSQL   |
| Driver      | https://github.com/jackc/pgx        |

a8m:
Fixed with #2332, Thanks!

PR #2332, opened 2022-02-15
fix(query): fix missing unique field in query builder template

MrParano1d:
when a query gets cloned the unique field doesn't get cloned

#2331


a8m:
Thanks, @MrParano1d for the contribution. It's appreciated 🙏 ❤️

a8m:
@all-contributors please add @MrParano1d for code

PR #2337, opened 2022-02-16
dialect/sql/schema: versioned migrations

masseelch:
This PR introduces the ability to have Ent (Atlas actually) create migration files. The files itself by default follow the format understood by golang-migrate/migrate without any extra configuration. Flyway can be configured to work with them. However, the file format itself is configurable by passing in a custom `migrate.Formatter` with the `WithFormatter()` builder.

Versioned migrations only work, if the new migration engine based on Atlas is enabled.

There are two ways to use the new feature:

1. Ent will create a new file `ent/migrate/main.go` which can be executed to create a new migration file. It also serves as an example of how to use the generated Client to create new files programatically.
2. There is a new cli command added: `ent migrate diff` which loads the graph and calculates the diff by connection to a dev database. 

Documentation will follow in another PR.


cmd/ent/ent_test.go
```diff
@@ -30,8 +33,29 @@ func TestCmd(t *testing.T) {
...
+	require.NoError(t, os.MkdirAll("migrations", 0750))
+	defer os.RemoveAll("migrations")
+
+	cmd = exec.Command(
+		"go", "run", "entgo.io/ent/cmd/ent",
+		"migrate", "diff",
+		"./ent/schema",
+		"--dir", "migrations",
+		"--dsn", "file:ent?mode=memory&_fk=1",
+		"--driver", "sqlite3",
```
a8m, line 51: Use atlas/golang-migrate format? Where the driver/dialect is used as a prefix? 

```suggestion
		"--dsn", "sqlite://file:ent?mode=memory&_fk=1",
```

cmd/internal/base/base.go
```diff
@@ -110,6 +114,58 @@ func DescribeCmd() *cobra.Command {
 	}
 }
 
+func DiffCmd() *cobra.Command {
+	var (
+		dsn string
+		drv string
+		dir string
```
a8m, line 121:
```suggestion
		dir, dsn string
```
drv is not needed (see previous comment),

cmd/internal/base/base.go
```diff
@@ -110,6 +114,58 @@ func DescribeCmd() *cobra.Command {
...
+		cmd = &cobra.Command{
+			Use: "migrate [flags] command",
+		}
+	)
+	cmd.AddCommand(
+		&cobra.Command{
+			Use:   "diff [flags] path",
+			Short: "create a new migration file by comparing the database / migration directory state with your schema graph",
+			Example: examples(
+				"ent diff ./ent/schema --dir migrations --dsn sqlite://file:ent.db?_fk=1",
```
a8m, line 131: Here you don't use the driver btw.

cmd/internal/base/base.go
```diff
@@ -16,7 +17,10 @@ import (
 	"text/template"
 	"unicode"
 
+	migrate2 "ariga.io/atlas/sql/migrate"
```
a8m, line 20: is it really necessary?

dialect/sql/schema/atlas.go
```diff
@@ -61,8 +60,8 @@ func WithDiffHook(hooks ...DiffHook) MigrateOption {
 	}
 }
 
-// SkipChanges allows skipping/filtering list of changes
-// returned by the differ before executing migration planning.
+// WithSkipChanges allows skipping/filtering list of changes
```
a8m, line 63: 👍

dialect/sql/schema/migrate.go
```diff
@@ -159,6 +161,19 @@ func (m *Migrate) Create(ctx context.Context, tables ...*Table) error {
...
+// Changes will be written to migration files by the configures Planner.
+func (m *Migrate) Diff(ctx context.Context, tables ...*Table) error {
+	if m.atlas.dir == nil {
+		return errors.New("no migration directory given")
+	}
+	plan, err := m.atDiff(ctx, m, tables...)
+	if err != nil {
+		return err
+	}
+	return migrate.New(nil, m.atlas.dir, m.atlas.fmt).WritePlan(plan)
```
a8m, line 174: May I suggest changing `migrate.New` to `migrate.NewPlanner`?

a8m: Passing `nil` driver feels odd to me. I don't want to abuse functional options, but consider use something like:
```go
migrate.NewPlanner(
    migrate.PlanDir(...),
    migrate.PlanFormatter(...),
    migrate.PlanDriver(...),
    ...
)
```

go.mod
```diff
@@ -32,6 +32,7 @@ require (
 	github.com/google/go-cmp v0.5.6 // indirect
 	github.com/hashicorp/hcl/v2 v2.10.0 // indirect
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
+	github.com/kr/pretty v0.2.1 // indirect
```
a8m, line 35: ?

a8m (commented):
Looks very very good @masseelch! I'm super excited about it.

cmd/ent/ent_test.go
masseelch: I thought about that. But ent users are using the dsn without driver prefix everywhere and I wanted to stay consistent. 

If I make it clear in the docs though going with the prefixed version should be okay.

cmd/ent/ent.go
```diff
@@ -10,6 +10,10 @@ import (
 	"entgo.io/ent/cmd/internal/base"
 
 	"github.com/spf13/cobra"
+
+	_ "github.com/go-sql-driver/mysql"
+	_ "github.com/lib/pq"
+	_ "github.com/mattn/go-sqlite3"
```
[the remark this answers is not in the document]
masseelch: Do we want to remove the command in Ent then? I am not sure. Yes it sucks, but there is no other way I know 😐

cmd/internal/base/base.go
masseelch: Yes. To create the `LocalDir`.

cmd/ent/ent_test.go
a8m: Agree, but cli is different from programming one, and tbh, I wanted to change the `client.Open` API to "prefixed dsns" a long time ago, without breaking users' code.

cmd/internal/base/base.go
a8m: I meant the package alias. You can give a different name to the `migrate` variable, but if you find it prettier/readable this way, keep it.

a8m (approved):
Awesome work @masseelch! ❤️ 🙏

PR #2350, opened 2022-02-21
entc/gen: singularize feature flag name for versioned migrations

masseelch:


a8m (approved):
💯 

Fix red CI before landing

PR #2353, opened 2022-02-22
dialect/entsql: add support for index-type annotation 

a8m:
In future PRs, we can add support for `Search` predicates to simplify constructing the WHERE expression manually. Also, we should consider add the common index types to entsql:

```go
entsql.FullText()
entgql.GIN()
...
```

### DIFF
--- a/dialect/sql/schema/mysql.go
+++ b/dialect/sql/schema/mysql.go
@@ -951,5 +951,22 @@ func (d *MySQL) atIndex(idx1 *Index, t2 *schema.Table, idx2 *schema.Index) error
 		}
 		idx2.AddParts(part)
 	}
+	if t, ok := indexType(idx1, dialect.MySQL); ok {
+		idx2.AddAttrs(&mysql.IndexType{T: t})
+	}
 	return nil
+}
+
+func indexType(idx *Index, d string) (string, bool) {
+	ant := idx.Annotation
+	if ant == nil {
+		return "", false
+	}
+	if ant.Type != "" {
+		return idx.Annotation.Type, true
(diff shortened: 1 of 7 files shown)


dialect/sql/schema/mysql.go
```diff
@@ -951,5 +951,22 @@ func (d *MySQL) atIndex(idx1 *Index, t2 *schema.Table, idx2 *schema.Index) error
...
+
+func indexType(idx *Index, d string) (string, bool) {
+	ant := idx.Annotation
+	if ant == nil {
+		return "", false
+	}
+	if ant.Type != "" {
+		return ant.Type, true
+	}
+	if ant.Types != nil && ant.Types[d] != "" {
```
masseelch, line 968: I'd expect `Types` to take precedence. Understood `Type` as default. And `Types` to override eg SQLite

yonidavidson (approved):

entc/integration/migrate/entv2/schema/user.go
```diff
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"entgo.io/ent"
+	"entgo.io/ent/dialect"
```
masseelch, line 11: Indention seems wrong

dialect/sql/schema/mysql.go
a8m: done

entc/integration/migrate/entv2/schema/user.go
a8m: why?

PR #2357, opened 2022-02-23
doc/website/blog: wrong module path for ogent

masseelch:


masseelch:
Added vanity URL to ariga.io. This PR is not needed.

PR #2358, opened 2022-02-23
Add FieldDefault method to Mutation interface and entc template.

freb:
The Save/SaveX methods on the generated <T>Create and <T>Update mutation wrappers set field defaults before hooks are run so code has no way to tell which fields were explicitly set and which are using default values.

This change adds a FieldDefault method to Mutation that is similar to the existing Field method, only it returns a fields default value. This allows code to compare the current and default value and infer if the
field was explicitly set to a non-default value.

A few things to note:

1. The template code will produce an empty switch statement if there are no field default values. I did not see an easy way to check for this, but if it is a deal breaker I can look the fields to check before proceeding.
2. `FieldDefault` has interaction with `ent/runtime` just like the internal `defaults` method for <T>Create and <T>Update mutation wrappers. Instead of exposing an error I am simply returning `nil, true` when the function has not been properly set. I figured this issue can be caught elsewhere, such as during the save call (which will generally precede use of FieldDefault) and it is not worth complicating the API by also returning an error value. Just let me know if this is a mistake and I should add in an error return parameter like `defaults` does.


freb:
My motivation for making this change was to allow privacy rules to protect specific fields from alteration. I have user attribute fields (`is_staff`, `is_superuser`) that need to be set during user creation by an admin and should later be alterable by an admin, but should never be editable by anyone else even though the user may alter other fields on their own User entity.

My User policy looks like this:

~~~go
func (User) Policy() ent.Policy {
	return privacy.Policy{
		Query: privacy.QueryPolicy{
			rule.AllowIfSuper(),
			rule.QueryUserFilter(),
		},

		Mutation: privacy.MutationPolicy{
			rule.AllowIfSuper(),
			rule.DenyIfNotActiveOrgAdmin(),
			rule.DenyFields("is_staff", "is_super", "password_hash"),
			// ...more filtering
		},
	},
}
~~~

I also have a `rule.AllowFields` that will deny access if a field is being set that is not in the list.

Here is one of the rules that makes use of the new `m.FieldDefault` method:

~~~go
// DenyFields checks the create/update mutation for set fields and denies
// access if any set fields are found in the fields parameter.
func DenyFields(fields ...string) privacy.MutationRule {
	return privacy.MutationRuleFunc(func(ctx context.Context, m ent.Mutation) error {
		ctx = disableRules(ctx)
		view := viewer.FromContext(ctx)
		if view == nil {
			return fmt.Errorf("DenyFields: %w", NoViewerError)
		}

		illegal := []string{}

		changed := m.Fields()

		for _, c := range changed {
			for _, f := range fields {
				if c == f {
					// If create, see if value is equal to default.
					if m.Op().Is(ent.OpCreate) {
						cur, _ := m.Field(c)
						def, _ := m.FieldDefault(c)
						if cur == def {
							continue
						}
					}
					// Defaults aren't applied on update calls, so only check if field was set.
					illegal = append(illegal, c)
				}
			}
		}

		if len(illegal) > 0 {
			return privacy.Denyf("viewer cannot set field(s): %s", strings.Join(illegal, ", "))
		}

		return privacy.Skip
	})
}
~~~

freb:
- asdf

PR #2369, opened 2022-02-28
dialect/sql/schema: fix no change condition in atlas

zeevmoney:


a8m:
Closed with #2374, I added you as a co-author of the PR.

image

PR #2375, opened 2022-03-04
dialect/sql/schema: add name to versioned migration files

imhuytq:
Currently, `golang-migrate` will not be able to scan versioned migration files generated because the filename does not match this pattern: [^([0-9]+)_(.*)\.(down|up)\.(. *)$](https://github.com/golang-migrate/migrate/blob/57aead3125fa6354d48199ac37f5bca5293ee600/source/parse.go#L21). So to satisfy the above pattern, we need to add names for the versioned migration files.

Another thing is that we should not execute `Planner.WritePlan` if there are no changes. Otherwise, it will generate empty migration files.


masseelch:
Hey @imhuytq, thank you your contribution.

@all-contributors please add @imhuytq for code.

PR #2406, opened 2022-03-16
entc/gen: move column default quoting to template

a8m:


entc/gen/template/migrate/schema.tmpl
```diff
@@ -38,8 +38,8 @@ var (
 				{{- with $c.Size }} Size: {{ . }},{{ end }}
 				{{- with $c.Attr }} Attr: "{{ . }}",{{ end }}
 				{{- with $c.Enums }} Enums: []string{ {{ range $e := . }}"{{ $e }}",{{ end }} },{{ end }}
-				{{- if not (isNil $c.Default) }} Default: {{ $c.Default }},{{ end }}
-				{{- if ne (len $c.Collation) 0 }} Collation: {{ $c.Collation }},{{ end }}
```
masseelch, line 42: Previously Collation was without "", not it is with. By intention? Looks like not.

a8m: I removed the quoting from `gen.Field.Column`. Intentional.

masseelch (approved):

PR #2414, opened 2022-03-19
Update getting-started.md

fgm:
Since Go 1.18 is now released, `go get` is no longer the way to download a command for any supported version: updated to use go install instead.


a8m (approved):
💯

PR #2416, opened 2022-03-20
Update schema-def to use proper method

fgm:
The exemple incorrectly adds an ignored `User.Index` method, although ent actually needs a `User.Indexes` method as part of `ent.Interface`


a8m (approved):
Thanks for the contribution @fgm 🙏

PR #2417, opened 2022-03-20
entc/gen: use join for loading m2m relationship

a8m:
Rewrite the way Ent eager-loads M2M edges, by replacing two queries with a single JOIN query. The main reasons are:
1. This change is necessary for the nested-pagination implementation in `entgql` in order to use SQL `WINDOW` functions and limit the number of edges returned for each entity.
2. In the average case, Ent users limit the number of records they query from the database using entgql/entproto pagination or use limit+offset, and the previous implementation used to scan all edges even though it was not necessary (e.g. `With<E>` configured with `Limit`, `Order` or `Where`). 
3. If we want to keep the previous implementation and avoid scanning all edges, we should add `JOIN` to the destination table which makes it irrelevant. 
4. In case there are many edges, the previous implementation is reached the database parameters limit (e.g. 65535 in PostgreSQL).  

### Test plan

Added 10K users, and connected each one of them to a unique group with attribute `active=false`. Then, created another 10 groups (with `active=true`) and connected them to all users.

The results are as follows:

```
TestMySQL/8 (branch:withm2m)
users->groups(active:false)	104.482914ms
users->groups			394.802676ms
groups(active:false)->users	105.381099ms
groups->users			104.649884ms

TestMySQL/8 (branch:master)
users->groups(active:false)	176.937021ms
users->groups			180.452625ms
groups(active:false)->users	111.331736ms
groups->users			165.322574ms
```

When querying `users->groups`, the new JOIN of `users.id` + `groups` columns generate 99990 duplicate rows, unlike the previous solution, that used 2 queries. One for getting all edges (without filtering) from join-table, and another one for querying the destination table (the one requested on `With`). 
However, in cases like `groups(active:false)` when there are no duplicate rows at all, the numbers are much better. 

### Future work

If we'll feel it is necessary, we can provide users an option to configure the way relationships are queried using "runtime annotation/hints". Some rough ideas I have in mind:

```go
client.User.Query().
    WithGroups(sql.LoadWithJoin()).
    WithFriends(sql.LoadWithSubQuery()).
    All(ctx)
```

### DIFF
--- a/entc/gen/template/dialect/sql/query.tmpl
+++ b/entc/gen/template/dialect/sql/query.tmpl
@@ -23,7 +23,7 @@ in the LICENSE file in the root directory of this source tree.
 {{ $builder := pascal $.Scope.Builder }}
 {{ $receiver := receiver $builder }}
 
-func ({{ $receiver }} *{{ $builder }}) sqlAll(ctx context.Context) ([]*{{ $.Name }}, error) {
+func ({{ $receiver }} *{{ $builder }}) sqlAll(ctx context.Context, hooks ...queryHook) ([]*{{ $.Name }}, error) {
 	var (
 		nodes = []*{{ $.Name }}{}
 		{{- with $.UnexportedForeignKeys }}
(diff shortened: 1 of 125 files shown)


entc/gen/template/dialect/sql/query.tmpl
```diff
@@ -272,74 +271,58 @@ func ({{ $receiver }} *{{ $builder }}) sqlQuery(ctx context.Context) *sql.Select
...
-					Table: {{ $.Package }}.{{ $e.TableConstant }},
-					Columns: {{ $.Package }}.{{ $e.PKConstant }},
-				},
-				Predicate: func(s *sql.Selector) {
-					s.Where(sql.InValues({{ $.Package }}.{{ $e.PKConstant }}[{{ if $e.IsInverse }}1{{ else }}0{{ end }}], fks...))
-				},
+			query.Where(func(s *sql.Selector) {
+				joinT := sql.Table({{ $.Package }}.{{ $e.TableConstant }})
+				{{- $edgeid := print $e.Type.Package "." $e.Type.ID.Constant }}
+				{{- $fk1idx := 1 }}{{- $fk2idx := 0 }}{{ if $e.IsInverse }}{{ $fk1idx = 0 }}{{ $fk1idx = 1 }}{{ end }}
```
masseelch, line 285: `{{ $fk1idx = 0 }}{{ $fk1idx = 1 }}`

That part looks not right. One should be `fk2idx`, correct?

masseelch (changes requested):

entc/gen/template/dialect/sql/query.tmpl
a8m: Correct.

masseelch (approved):
The tests did not pick up the bug, did they? Should we add a test for it?

masseelch (approved):
:100:

PR #2432, opened 2022-03-28
update example import

cjraa:
Generator example is importing `github.com/ariga/ogent` instead of `ariga.io/ogent`


masseelch (approved):
Thank you @cjraa, nice catch!

PR #2440, opened 2022-04-01
init db version before executing diff

CharlesGe129:
What's the problem: Under versioned migration on MySQL 8.0, unique string field would be generated as type varchar(191) instead of varchar(255). This is because in generating migration files by NamedDiff(), sqlDialect's init() method is not called. 

Further concern: For init() function, MySQL and Postgres's implementation checks the database version which is what I want. But SQLite's implementation checks the foreign_keys support. So I think there might be a better may to check the database version before doing Diff().

### DIFF
--- a/dialect/sql/schema/migrate.go
+++ b/dialect/sql/schema/migrate.go
@@ -173,6 +173,9 @@ func (m *Migrate) NamedDiff(ctx context.Context, name string, tables ...*Table) 
 	if m.atlas.dir == nil {
 		return errors.New("no migration directory given")
 	}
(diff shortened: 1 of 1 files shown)


masseelch (approved):

masseelch:
@all-contributors please add @CharlesGe129 for code.

masseelch:
@CharlesGe129 Thank you for this. Indeed calling `init()` is what we should do.

CharlesGe129:
@masseelch Thanks! Glad to be helpful.

PR #2441, opened 2022-04-01
Fix incorrect Go code in GoType() doc comments

fgm:
All examples used repeated argument names, which do not compile


a8m:
Thanks for the fix @fgm! Can you please update the numeric.tmpl and run codegen (go generate ./schema/field/…).

fgm:
Done, sorry I hadn't noticed it was a generated file.

a8m:
Thanks 🙏🏻

PR #2470, opened 2022-04-13
dialect/sql/schema: option to enable atlas sum file

masseelch:
Documentation will follow in another PR describing why this is useful and how to use it (and fix failures).


dialect/sql/schema/atlas.go
```diff
@@ -264,6 +264,13 @@ func WithFormatter(fmt migrate.Formatter) MigrateOption {
 	}
 }
 
+// EnableSumFile instructs atlas to generate a migration directory integrity sum file as well.
+func EnableSumFile() MigrateOption {
```
a8m, line 268:
```suggestion
func WithSumFile() MigrateOption {
```

Maybe use `With` to stay consistent with other `MigrateOption`s.

a8m (approved):
:shipit:

PR #2471, opened 2022-04-13
doc/md: document atlas sum file usage

masseelch:


doc/md/versioned-migrations.md
```diff
@@ -241,3 +241,115 @@ func main() {
    }
 }
 ```
+
+## Atlas migration directory integrity file
+
+Suppose you have multiple teams develop a feature in parallel and both of them need a migration. If Team A and Team B do
+not check in with each other, they might end up with a broken migration directory since new files do not raise a merge
+conflict in a version control system like git. To prevent concurrent creation of new migration files and guard against
```
a8m, line 249: Maybe give a scenario example? e.g. Both teams tried to add the same column => will cause a migration failure.

doc/md/versioned-migrations.md
```diff
@@ -241,3 +241,115 @@ func main() {
...
+accidental changes in the migration history, Atlas has a feature called __Migration Directory Integrity File__, which simply is
+another file in your migration directory called `atlas.sum`.
```
a8m: I'd also explain why this is good for this use case, how it prevents these scenarios, and how to troubleshoot such conflicts using this file. Please, let me know if you need help with this

doc/md/versioned-migrations.md
```diff
@@ -241,3 +241,115 @@ func main() {
...
+   "ariga.io/atlas/sql/sqltool"
+   "entgo.io/ent/dialect/sql"
+   "entgo.io/ent/dialect/sql/schema"
+   "entgo.io/ent/entc"
+   "entgo.io/ent/entc/gen"
+)
+
+func main() {
+   // Load the graph.
+   graph, err := entc.LoadGraph("/.schema", &gen.Config{})
```
a8m, line 284: I'm curious why you don't use `<project>/ent/migrate.Tables` instead? It's defined statically, and also promises that the codegen already was executed for the schema.

doc/md/versioned-migrations.md
```diff
@@ -241,3 +241,115 @@ func main() {
...
+   // Open connection to the database.
+   dlct, err := sql.Open("mysql", "root:pass@tcp(localhost:3306)/test")
+   if err != nil {
+      log.Fatalln(err)
+   }
+   // Inspect it and compare it with the graph.
+   m, err := schema.NewMigrate(dlct, schema.WithDir(d),
+      // highlight-start
+	  // Enable the Atlas Migration Directory Integrity File.
+      schema.WithSumFile(),
```
a8m, line 306:
```suggestion
      // Enable the Atlas Migration Directory Integrity File.
      schema.WithSumFile(),
```

doc/md/versioned-migrations.md
```diff
@@ -241,3 +241,169 @@ func main() {
...
+   }
+   if err := m.Diff(context.Background(), tbls...); err != nil {
+      log.Fatalln(err)
+   }
+}
+```
+
+In addition to the usual `.sql` migration files the migration directory will contain the `atlas.sum` file. Every time
+you let Ent generate a new migration file, this file is updated for you. However, every manual change made to the
+mitration directory will render the migration directory and the `atlas.sum` file out-of-sync. With the Atlas CLI you can
```
a8m, line 370:
```suggestion
migration directory will render the migration directory and the `atlas.sum` file out-of-sync. With the Atlas CLI you can
```

a8m (approved):
Looks good 💯

PR #2482, opened 2022-04-18
doc/md - added usage note for golang-migrate

thmeitz:


masseelch:
🙏

doc/md/versioned-migrations.md
```diff
@@ -241,3 +241,17 @@ func main() {
...
+
+```
+"user:password@tcp(host:port)/dbname?multiStatements=true"
+```
+
+**Why?**
+
+Without the `multiStatements=true` parameter, the migration will fail if there is more than one SQL statement in the migration file.
+
+Please note that all other options you use for your DSN must be added as well!
```
masseelch, line 257: I'd drop this sentence and change the above to something like "you need to add the `multiStatements` parameter to true as in the example below" and then take the DSN we used in the documents with the param applied. WDYT?

thmeitz: I'm fine with your change.

masseelch (approved):

Issue #2477, opened 2022-04-15
ogent / ent  - type generation problems


thmeitz:
- [ x ] The issue is present in the latest release.
- [ x ] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

1. ogent generates wrong type
`schema.go` definition `field.Uint32("id").StorageKey("client_id"),`
`ogent` generates `int32` type
- see: one of the 600 errors in the example repo
2. `ent` generation problem if table contains `table` column
- see: https://github.com/thmeitz/entisp/blob/647e31671d6a07919c4ab19d210c8b1b14923e37/ent/iptable/where.go#L102
3. `ent` generation problem if table name is `client`
renamed schema  to `customer` in example repo. rename it back to client and you'll get a `go generate` error
4. `ogent` req.TmpData.Get undefined (type []byte has no field or method Get)
- NewOptByteSlice will not be generated by `ogent`
- see: https://github.com/thmeitz/entisp/blob/3c4366726268c071416b9815ce0205e0d5554200/ent/ogent/responses.go#L860
## Expected Behavior 🤔

1. expecting type `uint32`
2. expecting thats using a pre- or suffix? dont have an idea how we should fix this
3. expecting thats working (see 2.)
4. NewOptByteSlice should be generated

## Steps to Reproduce 🕹

Steps:

1. checkout the repo [ispent](https://github.com/thmeitz/entisp)

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.18  |
| Ent         | v0.10.2-0.20220321093754-edd968490ea2  |

a8m:
Thanks for reporting this issue @thmeitz 🙏 

One of us will give it a look next week.

masseelch:
Hey @thmeitz,

1. OpenAPI does not allow unsigned integers. However, for unsigned int we should add a min value of 0 and fix the bug you detected.
2. The llinked PR now skips generation of the `Table` predicate
3. Some keywords are used internally, `Client` being one of them. I extended the documentation to mention this as well.
4. Fixed in the linked PR
5. See 1

PR #2486, opened 2022-04-20
entc/gen: skip `Table()` predicate and warn about reserved schema name (`Client`)

masseelch:
#2477


rotemtam (commented):
LGTM but I'll let @a8m approve

doc/md/schema-def.md
```diff
@@ -58,6 +58,11 @@ the root directory of your project, and can be generated by `entc` as follows:
 go run entgo.io/ent/cmd/ent init User Group
 ```
 
+:::note 
+Pleae note, that `Client` is a reserved Go struct and cannot be used as schema name. For database tables
+named `client`, please use an annotation as mentioned [here](schema-annotations.md#custom-table-name).
+:::
```
a8m, line 64: What's so special about `Client`? Maybe expand the godoc of [gen.ValidSchemaName](https://pkg.go.dev/entgo.io/ent/entc/gen#ValidSchemaName) and link to it?

a8m (approved):

PR #2506, opened 2022-04-28
dialect/sql: support string based pk for mysql56 indexes (prevent err…

masseelch:
…or 1071)


a8m (approved):

entc/integration/customid/ent/schema/revision.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+// Revision holds the schema definition for the Revision entity.
+type Revision struct {
+	ent.Schema
+}
+
+// Fields of the Revision.
+func (Revision) Fields() []ent.Field {
+	return []ent.Field{field.String("id")}
```
a8m, line 20:
```suggestion
	return []ent.Field{
		field.String("id"),
	}
```

a8m (approved):
💯

Issue #2484, opened 2022-04-19
Multiple JOIN aggregation


Marahin:
Hi,

I'm looking for support when it comes to doing multiple JOIN aggregation. Currently I gave up and wrote raw sql: 

```go
		rows, err := db.Client.Debug().QueryContext(context.Background(),
			`
			select
				players.name,
				players_with_profits.profit as "profit",
				players_with_experience.experience as "experience",
				players.level,
				players.vocation,
				players.world,
				g.name as "guild"
			from
							players
			left join (
				select
								players.id as id,
								sum(pdp.profit) as profit
				from
								players
				join player_daily_profits pdp on
								pdp.player_id = players.id
				where
								pdp.created_at >= $1
				group by
								players.id) players_with_profits on
							players_with_profits.id = players.id
			left join (
				select
								players.id as id,
								sum(er.relative) as experience
				from
								players
				join experience_records er on
								er.player_id = players.id
				where
								er.created_at >= $1
				group by
								players.id) players_with_experience on
							players_with_experience.id = players.id
			left join guilds g on
				g.id = players.guild_players
			where
				experience is not null and profit is not null
			order by
				experience desc,
				profit desc
			LIMIT $2`, beginningDay, limit,
		)
```

But I would prefer to stick to ent's built in queries interface. Is it currently possible to break down this query so it contains as little raw sql as possible?

yonidavidson:
Hey @Marahin  , have you had a look at these examples https://entgo.io/docs/predicates/#get-all-users-with-a-tesla-car ?

PR #2509, opened 2022-04-30
Sub queries

crossworth:
This PR adds a few example of the sub-query usage, I could not think a better example for the join case.
Related: https://github.com/ent/ent/issues/2484#issuecomment-1108492471

I have changed the `builder.go` as well otherwise a empty `FROM` is always generated:
image

### DIFF
--- a/entc/integration/integration_test.go
+++ b/entc/integration/integration_test.go
@@ -689,6 +689,16 @@ func Select(t *testing.T, client *ent.Client) {
 	}).Int(ctx)
 	require.NoError(err)
 	require.Equal(1, i)
+	// example with join
+	u = client.User.Create().SetName("crossworth").SetAge(28).SaveX(ctx)
+	userTable := sql.Table(user.Table)
+	subQuery = sql.Select("id").From(userTable).Where(sql.EQ(userTable.C(user.FieldName), "crossworth"))
+	users = client.User.Query().Modify(func(s *sql.Selector) {
+		s.Select("*").From(userTable)
+		s.Join(subQuery).On(userTable.C("id"), subQuery.C("id"))
+	}).AllX(ctx)
+	require.Len(users, 1)
+	require.Equal(u.ID, users[0].ID)
 }
 
 func ExecQuery(t *testing.T, client *ent.Client) {
--- a/entc/integration/integration_test.go
+++ b/entc/integration/integration_test.go
@@ -692,15 +692,15 @@ func Select(t *testing.T, client *ent.Client) {
 		Int(ctx)
 	require.NoError(err)
 	require.Equal(1, i)
-	
+
 	// Select with join.
 	u = client.User.Create().SetName("crossworth").SetAge(28).SaveX(ctx)
 	id := client.User.
 		Query().
-		Modify(func(s *sql.Selector) {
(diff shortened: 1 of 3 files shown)


entc/integration/integration_test.go
```diff
@@ -680,6 +680,25 @@ func Select(t *testing.T, client *ent.Client) {
...
+	// example with join
+	u = client.User.Create().SetName("crossworth").SetAge(28).SaveX(ctx)
+	userTable := sql.Table(user.Table)
+	subQuery = sql.Select("id").From(userTable).Where(sql.EQ(userTable.C(user.FieldName), "crossworth"))
+	users = client.User.Query().Modify(func(s *sql.Selector) {
+		s.Select("*").From(userTable)
+		s.Join(subQuery).On(userTable.C("id"), subQuery.C("id"))
+	}).AllX(ctx)
+	require.Len(users, 1)
+	require.Equal(u.ID, users[0].ID)
```
a8m, line 701:
```suggestion
	require.Equal(1, i)
	
	// Select with join.
	u = client.User.Create().SetName("crossworth").SetAge(28).SaveX(ctx)
	id := client.User.
		Query().
		Modify(func(s *sql.Selector) {
			subQuery := sql.Select(user.FieldID).
				From(sql.Table(user.Table)).
				Where(sql.EQ(userT.C(user.FieldName), "crossworth"))
			s.Join(subQuery).On(s.C(user.FieldID), subQuery.C(user.FieldID))
		}).
		OnlyIDX(ctx)
	require.Equal(u.ID, id)
```

entc/integration/integration_test.go
```diff
@@ -680,6 +680,25 @@ func Select(t *testing.T, client *ent.Client) {
 	require.Len(gs, 2)
 	require.Equal(hub.QueryUsers().CountX(ctx), gs[0].UsersCount)
 	require.Equal(lab.QueryUsers().CountX(ctx), gs[1].UsersCount)
+
+	// Select Subquery.
+	t.Log("select subquery")
+	subQuery := sql.SelectExpr(sql.Raw("1"))
+	i, err := client.User.Query().Modify(func(s *sql.Selector) {
+		s.Select("*").From(subQuery.As("s"))
+	}).Int(ctx)
```
a8m, line 689:
```suggestion
	i, err := client.User.
		Query().
		Modify(func(s *sql.Selector) {
			subQuery := sql.SelectExpr(sql.Raw("1")).As("s")
			s.Select("*").From(subQuery)
		}).
		Int(ctx)
```

Style preference. I prefer to keep subqueries in the scope they are used (if it's possible), and add extra line breaks if it keeps the code more readable (don't have a strong rule here). I hope you don't mind 😃

entc/integration/integration_test.go
a8m: `OnlyID` expects for exactly one ID. 

Also, I think the `Modifier` can be simplified there and may be replaced with a [custom predicate](https://entgo.io/docs/predicates/#custom-predicates) (`Where(func(...))` because it filters the results.

a8m (commented):
Looks good, I've added a few minor comments.

Thanks for the contribution @crossworth 🙏 ❤️

a8m (approved):

PR #2538, opened 2022-05-10
cmd/internal: support go1.18 workspace in generate.go command

### LINKED ISSUE #2438: support golang 1.18 workspaces
in workspace enabled repo
```
 $ go generate ./ent                                                                                                                                                                                           ☸ rancher-desktop on ☁️  
go: -mod may only be set to readonly when in workspace mode, but it is set to "mod"
        Remove the -mod flag to use the default readonly value,
        or set GOWORK=off to disable workspace mode.
```

yonidavidson:
fix #2438

### DIFF
--- a/cmd/internal/base/base.go
+++ b/cmd/internal/base/base.go
@@ -210,7 +210,11 @@ func createDir(target string) error {
 	if target != defaultSchema {
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


JK-97:
when merge it?

PR #2539, opened 2022-05-10
Workaround for max params eager loading

rubensayshi:
opening this PR to discuss a possible fix for this problem, I can polish this if we can settle on a good direction for this.

the issue we're having is that if we're querying more than 65535 records and doing eager-loading then we'll hit:
```
pq: got 70000 parameters but PostgreSQL only supports 65535 parameters
```

the simplest solution is probably to just document a warning for this and that the user (me in this case) should chunk the requests, but that comes with some risks... (ie a record being inserted in between etc.)

the solution I'm proposing here is fairly simple as well, though it needs to be cleaned up more ... not sure where these helper functions should go? maybe on the Driver?

an alternative I ran into which might be even better (but more work / complexity): https://klotzandrew.com/blog/postgres-passing-65535-parameter-limit

### DIFF
--- a/dialect/sql/helpers.go
+++ b/dialect/sql/helpers.go
@@ -1,0 +1,34 @@
+package sql
+
+import (
+	"entgo.io/ent/dialect"
+)
+
+func MaxParams(dialectName string) int {
+	switch dialectName {
+	case dialect.Postgres:
+		return 65535
+	default:
+		return -1
+	}
+}
+
+func Chunked(fn func(start, end int) error, n int, maxPerChunk int) error {
+	if maxPerChunk == -1 {
+		return fn(0, n)
+	}
+
+	for start := 0; start < n; start += maxPerChunk {
+		end := start + maxPerChunk
+		if end > n {
+			end = n
+		}
+
+		err := fn(start, end)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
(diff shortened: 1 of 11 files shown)


crossworth:
Hello, please see this comment:
https://discord.com/channels/885059418646003782/885059419107360779/950507215230480504

I think the ORM should handle cases like this, but I'm not sure the best way to do it.
I do think that it should support `INSERT` `UPDATE` `DELETE` as well, since the problem is not limited `QUERY` only.

rubensayshi:
ye, the `unnest` option that I linked to is actually a good option for WHERE statements I think, but gets more complicated for inserts...

I think maybe a good first step would be to make the `In` predicate use `unnest` for postgres?  
the `unnest` approach requires specifying the type in the unnest (ie `UNNEST($1::[]int)`), which can be know at that time, and this would also immediately apply to `WHERE` clauses of `UPDATE` and `DELETE` statements.

then the only cases which still need to be covered independently are: INSERTs and UPDATEs with an insane amount of SET params, the latter being an extreme edge case, the INSERT being a valid case but something which requires a different approach to do it properly...

I'll try to find some time to take a stab at the `unnest` in the predicate, should have a few hours for this either today or tomorrow and share that.

karankohli-cf:
any update on this?

hachh:
Any update on this issue ? Eager loading fails when there are more than 65535 rows...

PR #2540, opened 2022-05-11
M2M supports association table already defined in schema

tsingsun:
```
// Edges of the User.
    return []ent.Edge{
        edge.To("groups", Group.Type).
            // wile generate a new table "groups"
            StorageKey(
                edge.Table("groups"),
                edge.Columns("user_id", "group_id"),
                edge.Symbols("groups_id1", "groups_id2")
            ),
    }
}
```
in Edges method, StorageKey only support generate `to` and `form` key column,but in fact, the association table has to need more colunms In some cases.

if I defined a schema and the table name must as same as StorageKey's, run `generate`, `migrate/shcema.go` has  error because there two XXXXXTable and XXXXColumns with the same name

 so reuse complex mode defined in schema would be better.

1. primary key allow defined in schema.
2. foreign keys force to reset.

### DIFF
--- a/entc/gen/graph.go
+++ b/entc/gen/graph.go
@@ -497,7 +497,7 @@ func (g *Graph) Tables() (all []*schema.Table, err error) {
 					Symbol:     fkSymbol(e, owner, ref),
 				})
 			case M2M:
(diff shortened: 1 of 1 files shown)


tsingsun:
commit: e1c52774

Issue #2551, opened 2022-05-18
Consider support for Neo4j


fbiville:
I see there is support for Gremlin, which makes me think there is room to support Neo4j and Cypher.
I'm happy to work on the initial integration for the Cypher dialect and other related areas.

fbiville:
After a quick POC, I already have some questions.

Some specific Cypher queries cannot run in an explicit transaction (`CALL {...} IN TRANSACTIONS` [Neo4j 4.4 or later] and `... USING PERIODIC COMMIT` [Neo4j up to 4.4]).

The current API seems very transaction-focused, so I'm not sure if there is a way for the user to explicit flag specific queries as auto-commit or not?

PR #2552, opened 2022-05-18
Create POC for Neo4j integration

fbiville:
This is a small, probably very wrong experiment for Neo4j support (following up on https://github.com/ent/ent/issues/2551).
I time-boxed this to 1 hour without any knowledge of the codebase.

I am happy to get this further, with proper testing and more guidance, as what I've done here is most likely not enough at all (and there are issues we would need to discuss as well if the current implementation remains).


fbiville:
Some implementation notes/questions:

## `context.Context`
- `context.Context` is not used in the latest release of the Neo4j Go driver (but will be in 5.0)
- some APIs such as `Commit` and `Rollback` do not accept a `context.Context` but their Neo4j counterparts will, in 5.0 (since `COMMIT` and `ROLLBACK` are messages sent over the wire). Is there a plan to add `Context` there?

## Transactions
Before 5.0, there is no easy way for the user to retry transactions unless they use what we call transaction functions (i.e. managed transaction callbacks). 
However, the current API seems to align with explicit transactions and that's what the POC does.

Also, some specific Cypher queries CANNOT run with explicit transactions (`CALL {} IN TRANSACTIONS` with Neo4j 4.4 and later, `USING PERIODIC COMMIT` with Neo4j up to 4.4).
They need to be run with `Session.Run`. Is there any way for the executor to get that info so that it picks the right Neo4j API?

Issue #2564, opened 2022-05-25
Non-ASCII characters in field names


hax10:
Hi! I'm just getting started with Ent and have had a very positive experience so far.

I have a question about whether non-ASCII characters are allowed in ```ent.Field``` names.

For the project I'm working on, it would be convenient to use foreign language field names in the database. I've had preliminary success doing this as follows:

```go
func (SomeEntity) Fields() []ent.Field {
	return []ent.Field{
		field.String("X量"),
    ...
```

The above snippet compiles and Ent can create the relevant DB table and Go struct definitions, and insert data into the DB without any problems.

Two things to note:

1. The entity name ```SomeEntity``` can only contain ASCII, because Ent uses this name in import statements. Go only allows a subset of ASCII chars in import paths (see https://github.com/golang/go/issues/43052).

2. The field name must start with a char in ```/[a-zA-Z]/```, otherwise Ent's code generation fails with an ```illegal UTF-8 encoding (and 16 more errors)```, hence the initial "X" before the Chinese character "量".

My question is: **do Ent's field names support non-ASCII chars?**

My concern is that, although the example above runs successfully on both SQLite and MariaDB, the field names in my DB have become garbled text.

Using the example of "X量" above, in SQLite, Ent has generated the following field name (in hex) ```78 c3 a9 c2 87 c2 8f```. Now ```78``` is lowercase "x" (Ent has lower-cased the initial char), and the six bytes after it appear to be "量". However, in proper UTF-8 encoding, this should be ```78 e9 87 8f```, with only three bytes rather than six for the Chinese character. Even UTF-32 doesn't need 6 bytes to encode one Unicode code point.

I've had a quick look at the code in [/entc/gen/func.go](https://github.com/ent/ent/blob/c071898395eb49a32ffdb7072e04b78cee499ef1/entc/gen/func.go), and it appears to be Unicode-aware, but clearly Ent is doing something in the field name generation code that is mangling non-ASCII chars. Is anyone able to explain what's going on or point me in the right direction?

Many thanks!

PR #2568, opened 2022-05-26
Fix snake casing code for multibyte (i.e. non-ASCII) characters

hax10:
This PR resolves the problem in https://github.com/ent/ent/issues/2564

Ent users who already have non-ASCII chars in their DB field names will need to recreate/migrate their database tables (hopefully this shouldn't affect many users)

### DIFF
--- a/entc/gen/func.go
+++ b/entc/gen/func.go
@@ -175,14 +175,20 @@ func snake(s string) string {
 		j int
 		b strings.Builder
 	)
-	for i := 0; i < len(s); i++ {
-		r := rune(s[i])
-		// Put '_' if it is not a start or end of a word, current letter is uppercase,
+	// Split a string on Unicode code point boundaries
+	// e.g. "Beløb" => [B e l ø b]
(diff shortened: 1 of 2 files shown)


a8m:
Thanks for contributing this fix @hax10. Re-reading this now, it sounds like a backward-incompatible change so I need to think about how we can detect this type of column renaming in migration to avoid data loss.

Would you mind checking before if the [`StorageKey`](https://entgo.io/docs/schema-fields/#storage-key) option solves your issue?

hax10:
Amazing! Thank you. I can indeed use ```StorageKey()``` to force the DB column name to be some non-ASCII value, i.e.:

```go
field.String("<ASCII Variable Name>").StorageKey("<non-ASCII database column name>")
```

If you guys leave this issue open and I find some spare time, I will poke around Atlas and the Ent migration engine, and re-submit this PR with some migration code. It should be fairly straightforward, since column renaming is well supported on all database engines, even SQLite.

PR #2607, opened 2022-06-04
Rename examples dir to _examples

frederikhors:


rotemtam:
Hi,
Thanks for the suggestion. Personally, I don't think _examples looks nicer than plain old "examples".

Otherwise, consider that many conversations around the community link to these examples, while not strictly an API, those link would break and instead of being helpful to newcomers they will now add confusion.

Is there some technical reason for this change? In general, having some context in the PR body is helpful for reviews 🙏

Thanks again

frederikhors:
Hi @rotemtam! Thank you for your work! 😃

> I don't think _examples looks nicer than plain old "examples"

I don't care how much more beautiful it is. I created this PR for the two points I wrote: in order not to download all those files that also deceive IDEs such as VSCode with wrong imports from examples.

The directories with "_" before the name are excluded by default with Go.

I did the same for gqlgen: https://github.com/99designs/gqlgen/pull/1734

frederikhors:
> many conversations around the community link to these examples

We can put a Readme file in examples dir if you like that alerts about new path.

a8m:
> The directories with "_" before the name are excluded by default with Go.

Can you elaborate? Excluded by what?

Another alternative which I prefer is to make it a separate Go module. Thoughts?

frederikhors:
> Excluded by what?

They are treated like not important file in IDEs and in certain cases not downloaded at all with "go get/install".

I do not think another module is needed. It's just how Go works: underscore before directories that are not code for this module. gqlgen has shrinked a lot with my PR. And other projects too.

a8m:
> certain cases not downloaded at all with "go get/install".

Can you elaborate on which cases?

> I do not think another module is needed. It's just how Go works: underscore before directories that are not code for this module. gqlgen has shrinked a lot with my PR. And other projects too.

I’m not sure this is right because I just tried it on a few repositories. However, “go get” won’t download this directory if it’s a different module.

Please, give it a try or share with us some references/links if you have some. Thanks for the help 🙏🏻

frederikhors:
> Can you elaborate on which cases?

I do not remember because was years ago.

What's the deal with underscore aside from redirecting links in discussions?

Whether it's in a new module or not is not important, the important thing is that IDEs don't see it as code for THIS module (and with underscore it works).

The separate module seems to me a lot more work.

PR #2610, opened 2022-06-05
Fix misprint

frederikhors:


a8m:
Hey @frederikhors and thanks for the contribution 🙏🏻

The WithWhereFilters option is actually deprecated. Use the latest master of ent/contrib for now and we’ll release a new version soon. 

https://github.com/ent/contrib/blob/master/entgql/extension.go#L124

PR #2617, opened 2022-06-06
feat(entc/gen): add preallocate for scope map

sashamelentyev:


a8m:
Thanks for the contribution @sashamelentyev. Please, squash all these small commits into one. 

Closing.

PR #2633, opened 2022-06-09
Fix compile error caused by entc template when Entity name has initials TX

hax10:
When creating an Entity with initials "TX", e.g. TumorXrays,

the code generated by entc cannot compile, because of a name clash in the generated transaction (**tx**) ```Unwrap()``` function:

```go
func (tx *TumorXrays) Unwrap() *TumorXrays {
	tx, ok := tx.config.driver.(*txDriver)
	...
```


a8m:
Thanks for the contribution @hax10 🙏

a8m:
@all-contributors please add @hax10 for code

PR #2636, opened 2022-06-09
dialect/sql/schema: introduce type store interface

masseelch:
This will allow us to store information about pk ranges allocated per type in a different location thant the database. This is useful for deterministic pk range computation for things like generating atlas.hcl out of the Ent schema.


a8m (approved):
💯

PR #2644, opened 2022-06-13
dialect/sql/schema: file based type store

### LINKED ISSUE #2587: Versioned migrations do not work with GlobalUniqueID
Suppose trying to generate versioned migrations files for a global unique id enabled schema: 

```go
client.Schema.Diff(ctx, schema.WithDir(dir), schema.WithGlobalUniqueID(true))
```

will result into the following error:

```text
insert into ent_types: Error 1146: Table 'migrate.ent_types' doesn't exist
```

## Current Behavior 😯

There is an error.

## Expected Behavior 🤔

A new set of migration files are created that also allocate the correct ranges for the primary keys.

masseelch:
This PR adds support for a file based type storage when using versioned migrations. The file called `.ent_types` is written to the migration directory alongside the migration files and will be kept in sync for every migration file generation run.

In order to not break existing code, where the type storage might differ for different deployment, global unique ID mut be enabled by using a new option. This will also be raised as an error to the user when attempting to use versioned migrations and global unique ID.

Documentation will be added to this PR once feedback on the code is gathered.

Fixes #2587 

----- 

I have discussed with Ariel already, that we will move the configuration for allocated pk ranges to the schema level. We haven't yet figured out, how to do this best, but it will come.

### DIFF
--- a/dialect/sql/schema/atlas.go
+++ b/dialect/sql/schema/atlas.go
@@ -7,10 +7,10 @@ package schema
 import (
 	"context"
 	"database/sql"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io/fs"
+	"io/ioutil"
 	"sort"
 	"strings"
 
@@ -588,6 +588,17 @@ func (m *Migrate) ensureTypeTable(next Differ) Differ {
 			}
 			desired.Tables = append(desired.Tables, at)
 		}
+		// If there is a drift between the types stored in the database and the ones stored in the file,
+		// stop diffing, as this is potentially destructive. This will most likely happen on the first diffing
+		// after moving from online-migration to versioned migrations if the "old" ent types are not in sync with
+		// the deterministic ones computed by the new engine.
+		if len(m.dbTypeRanges) > 0 && len(m.fileTypeRanges) > 0 && !equal(m.fileTypeRanges, m.dbTypeRanges) {
+			return nil, fmt.Errorf(
+				"type allocation range drift detected: %v <> %v: see %s for more information",
+				m.dbTypeRanges, m.fileTypeRanges,
+				"https://entgo.io/docs/versioned-migrations#moving-from-auto-migration-to-versioned-migrations",
+			)
+		}
 		changes, err := next.Diff(current, desired)
 		if err != nil {
 			return nil, err
@@ -651,6 +662,8 @@ type dirTypeStore struct {
 	dir migrate.Dir
 }
 
+const atlasDirective = "atlas:sum ignore\n"
+
 // load the types from the types file.
 func (s *dirTypeStore) load(context.Context, dialect.ExecQuerier) ([]string, error) {
 	f, err := s.dir.Open(entTypes)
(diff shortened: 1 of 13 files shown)


dialect/sql/schema/atlas.go
```diff
@@ -326,6 +347,18 @@ func (m *Migrate) setupAtlas() error {
...
 	}
+	if m.universalID && m.atlas.dir != nil {
+		// If global unique ids and a migration directory is given, enable the file based type store for pk ranges.
+		m.typeStore = &dirTypeStore{dir: m.atlas.dir}
+		// The file based pk range store is not backwards compatible, since the allocated ranges were computed
+		// dynamically when computing the diff between a deployed database and the current schema. In cases where there
+		// exist multiple deployments, the allocated ranges for the same type might be different from each other,
+		// depending on when the deployment took part. To guard the user against bugs, the file based type store must
+		// be enabled by an option.
+		if !m.atlas.typeStoreConsent {
```
a8m, line 358: This is a copy of the comment above.

dialect/sql/schema/migrate.go
```diff
@@ -182,11 +183,45 @@ func (m *Migrate) NamedDiff(ctx context.Context, name string, tables ...*Table)
...
 	plan, err := m.atDiff(ctx, m, name, tables...)
 	if err != nil {
 		return err
 	}
+	if m.universalID {
+		// Add the required entries to the TypeTable.
+		for _, t := range m.typeStore.(*dirTypeStore).newTypes {
+			plan.Changes = append(plan.Changes, &migrate.Change{
+				Cmd:     m.atTypeRangeSQL(t),
+				Comment: `add pk range for "t" table`,
```
a8m, line 219:
```suggestion
				Comment: fmt.Sprintf("add pk range for "%q" table", TypeTable),
```

lagzi:
Hey, I'm getting `validating migration directory: checksum mismatch` when trying to create a new migration (`.Diff(..)`) because it seems that the `.ent_types` file changed but the checksum isn't updated.

masseelch:
> Hey, I'm getting `validating migration directory: checksum mismatch` when trying to create a new migration (`.Diff(..)`) because it seems that the `.ent_types` file changed but the checksum isn't updated.

Good catch. I will both add a test and a fix for that :-)

masseelch:
@lagzi I changed added a commit, that should fix your problem. If it doesn't please ping me on [Discord](https://discord.gg/qZmPgTE6RX) so I can help you resolve it :-)

dialect/sql/schema/atlas.go
```diff
@@ -537,6 +568,46 @@ func (m *Migrate) aIndexes(b atBuilder, t1 *Table, t2 *schema.Table) error {
...
+					return nil, fmt.Errorf("unexpected missing type range for %q", t)
+				}
+			}
+			// Override the types file created in the diff process with the "old" allocated types ranges.
+			if err := m.typeStore.(*dirTypeStore).save(m.dbTypeRanges); err != nil {
+				return nil, err
+			}
+			// Change the type range allocations since they will be added to the migration files when
+			// writing the migration plan to migration files.
+			m.typeRanges = m.dbTypeRanges
```
a8m, line 605: 💯

dialect/sql/schema/atlas.go
```diff
@@ -574,3 +645,52 @@ func descIndexes(idx *Index) map[string]bool {
...
 	return descs
 }
+
+const entTypes = ".ent_types"
+
+// dirTypeStore stores and read pk information from a text file stored alongside generated versioned migrations.
+// This behaviour is enabled automatically when using versioned migrations.
+type dirTypeStore struct {
+	dir migrate.Dir
+	// newTypes []string // store types added in this run
```
a8m, line 655: ??

a8m:
Looks good 🚀 

Please, fix the broken tests and I'll approve.

masseelch:
Currently this PR is blocked on a checksum problem: `.ent_types` is added to the checksum as well and this has to be resolved on Atlas.

a8m (approved):
I like the code but can't see `WithDeterministicGlobalUniqueID`. It's too long for me 🙈

PR #2657, opened 2022-06-16
Fix problem when model maps integer id to a GoType

stoikheia:
Two errors occurred when I use integer PK in a DB and associate GoType in a model.

```go
// Fields of the IntSid.
func (IntSid) Fields() []ent.Field {
	return []ent.Field{
		field.Int64("id").
			GoType(sid.New()).
			Unique().
			Immutable(),
	}
}
```

`sid` has been decrared in `/ent/entc/integration/customid/sid/sid.go`.

### 1. 

`intsid_create.go : (*IntSidCreateBulk) Save()`

image

```
ent/intsid_create.go:394:48: invalid operation: nodes[i].ID == 0 (mismatched types sid.ID and untyped int)
```

### 2. 

`intsid.go : (*IntSid) assignValues()`

image

```
ent/intsid.go:92:12: conversion from int64 to ID (string) yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
ent/intsid.go:98:26: conversion from int64 to ID (string) yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
```

### 1.

I took a way is similar to sqlSave() in *_create.go below to fix this problem.

https://github.com/ent/ent/blob/7017cbc898e8734920c5116a694606124eda7618/entc/gen/template/dialect/sql/create.tmpl#L23
https://github.com/ent/ent/blob/7017cbc898e8734920c5116a694606124eda7618/entc/gen/template/dialect/sql/create.tmpl#L33-L35

Because, Code generator will check GoType implements ValueScanner.

https://github.com/ent/ent/blob/7017cbc898e8734920c5116a694606124eda7618/schema/field/field.go#L1177-L1182

### 2.

I fixed it so that it is assigned as the value of the column type of model when it is GoType.

It has the same behavior as String, Bytes, UUID and Other.

And I aggregated check condition IsUUID() and IsOther() into HasGoType().

Because it indicates true when HasGoType() is called if the column is set field.Other and field.UUID.


giautm (approved):
This PR look good to merge.

@a8m

entc/gen/template/dialect/sql/create.tmpl
```diff
@@ -180,7 +180,11 @@ func ({{ $receiver }} *{{ $builder }}) Save(ctx context.Context) ([]*{{ $.Name }
 				{{- if $.HasOneFieldID }}
 					mutation.{{ $.ID.BuilderField }} = &nodes[i].{{ $.ID.StructField }}
 					{{- if or $.ID.IsString $.ID.IsUUID $.ID.IsBytes $.ID.IsOther }}
-						{{- /* Do nothing, because these 2 types must be supplied by the user. */ -}}
+						{{- /* Do nothing, because these 4 types must be supplied by the user. */ -}}
+					{{- else if or $.ID.Type.ValueScanner (not $.ID.Type.Numeric) }}
```
a8m, line 184: Why do you check `not $.ID.Type.Numeric`? Isn't the first check enough (i.e. `$.ID.Type.ValueScanner`)?

entc/gen/template/dialect/sql/create.tmpl
```diff
@@ -180,7 +180,11 @@ func ({{ $receiver }} *{{ $builder }}) Save(ctx context.Context) ([]*{{ $.Name }
...
+						if specs[i].ID.Value != nil {
+							nodes[i].ID.Scan(specs[i].ID.Value)
```
a8m, line 186: We need to handle the error returned by `Scan` here.

entc/integration/customid/ent/schema/intsid.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"entgo.io/ent"
+	"entgo.io/ent/entc/integration/customid/sid"
+	"entgo.io/ent/schema/edge"
+	"entgo.io/ent/schema/field"
+)
+
+// IntSid holds the schema definition for the IntSid entity.
+type IntSid struct {
```
a8m, line 15:
```suggestion
type IntSID struct {
```

entc/gen/template/dialect/sql/create.tmpl
stoikheia: Thanks!
Yes, `$.ID.Type.ValueScanner` only is enough.

entc/gen/template/dialect/sql/create.tmpl
stoikheia: I got it.

entc/integration/customid/ent/schema/intsid.go
```diff
@@ -0,0 +1,36 @@
...
+			Immutable(),
+	}
+}
+
+// Edges of the IntSid.
+func (IntSID) Edges() []ent.Edge {
+	return []ent.Edge{
+		edge.To("parent", IntSID.Type).
+			Unique(),
+		edge.From("children", IntSID.Type).Ref("parent"),
```
a8m, line 34:
```suggestion
		edge.From("children", IntSID.Type).
			Ref("parent"),
```

entc/integration/customid/customid_test.go
```diff
@@ -116,6 +118,27 @@ func CustomID(t *testing.T, client *ent.Client) {
...
+	iSIDBulk[0] = client.IntSID.Create().SetParent(iSID)
+	iSIDBulk[1] = client.IntSID.Create().SetParent(iSID)
+	iSIDChildren := client.IntSID.CreateBulk(iSIDBulk...).SaveX(ctx)
+	if strings.HasPrefix(t.Name(), "TestPostgres/") {
+		require.Equal(t, sid.ID("2"), iSIDChildren[0].ID)
+		require.Equal(t, sid.ID("3"), iSIDChildren[1].ID)
+	} else {
+		require.Equal(t, sid.ID("101"), iSIDChildren[0].ID)
+		require.Equal(t, sid.ID("102"), iSIDChildren[1].ID)
+	}
```
a8m, line 137: Can you elaborate on why we have this `TestPostgres` check? Why it's different than the rest?

entc/integration/customid/customid_test.go
```diff
@@ -116,6 +118,27 @@ func CustomID(t *testing.T, client *ent.Client) {
...
+	iSID := client.IntSID.Create().SaveX(ctx)
+	require.Equal(t, sid.ID("1"), iSID.ID)
+	iSIDChildID := sid.ID("100")
+	iSIDChild := client.IntSID.Create().SetID(iSIDChildID).SetParent(iSID).SaveX(ctx)
+	require.Equal(t, iSIDChildID, iSIDChild.ID)
+	require.Equal(t, iSID.ID, iSIDChild.QueryParent().OnlyX(ctx).ID)
+	iSIDBulk := make([]*ent.IntSIDCreate, 2)
+	iSIDBulk[0] = client.IntSID.Create().SetParent(iSID)
+	iSIDBulk[1] = client.IntSID.Create().SetParent(iSID)
+	iSIDChildren := client.IntSID.CreateBulk(iSIDBulk...).SaveX(ctx)
```
a8m, line 130:
```suggestion
	root := client.IntSID.Create().SaveX(ctx)
	require.EqualValues(t, "1", root.ID)
	cid := sid.ID("100")
	child := client.IntSID.Create().SetID(cid).SetParent(root).SaveX(ctx)
	require.Equal(t, cid, child.ID)
	require.Equal(t, rott.ID, child.QueryParent().OnlyX(ctx).ID)
	children := client.IntSID.CreateBulk(
		client.IntSID.Create().SetParent(root),
		client.IntSID.Create().SetParent(root)
	).SaveX(ctx)
```

Give these variables shorter names, please 🙏

a8m:
Looks great! Last minor comments, please.

entc/integration/customid/customid_test.go
stoikheia: First, a reason why this has exists. It checks returned values from ent whether they are set returned values from DB.

A reason of difference of code each RDBS is below.

Mysql and sqlite assign the next value which is from maximum value of the column when auto increment.

> https://dev.mysql.com/doc/refman/8.0/en/example-auto-increment.html
> "When you insert any other value into an AUTO_INCREMENT column, the column is set to that value and the sequence is reset so that the next automatically generated value follows sequentially from the largest column value. "
> https://www.sqlite.org/autoinc.html
> "The ROWID chosen for the new row is at least one larger than the largest ROWID"
> 

Auto-increment of Postgres is represented as a type named Serial. ( off course, you know. ) And it behaves as a function `nextval（' tablename_colname_seq'）`
https://www.postgresql.org/docs/13/datatype-numeric.html#DATATYPE-SERIAL
https://www.postgresql.org/docs/13/functions-sequence.html

And ent creates a table by SQL below when test of IntSID run.
```sql
CREATE TABLE "int_si_ds" (
  "id" bigint NOT NULL GENERATED BY DEFAULT AS IDENTITY,
  "int_sid_parent" bigint NULL, PRIMARY KEY ("id"),
  CONSTRAINT "int_si_ds_int_si_ds_parent" FOREIGN KEY ("int_sid_parent") REFERENCES "int_si_ds" ("id") ON DELETE SET NULL
)
```

> https://www.geeksforgeeks.org/postgresql-identity-column/#:~:text=The%20GENERATED%20ALWAYS%20instructs%20PostgreSQL,value%20for%20the%20identity%20column.
> "The GENERATED AS IDENTITY constraint is the SQL standard-conforming variant of the PostgreSQL’s SERIAL column."

Unfortunately, Serial will not associate with column value user inserted.

Therefore, SaveX() will not return "101" or "102" as ID after insertion of sid.ID("100") on Postgres.

Even worse, Postgres will occer an error if insert ID "2" instead of "100". Because it will conflict between sequence and exsisting value.

```text
panic: ent: constraint failed: insert nodes to table "int_si_ds": pq: duplicate key value violates unique constraint "int_si_ds_pkey"
```

Therefore, I cannot unify the test Postgres and the others.

entc/integration/customid/customid_test.go
stoikheia: ok, but some variable names will conflict.
I'll wrap it by sub test.

in this way.
https://github.com/stoikheia/ent/pull/11/commits/1bfabcfb9621fd2fcdbb075e3d147d855b4f3e1e#

stoikheia: I pushed this commit.
https://github.com/ent/ent/pull/2657/commits/5f5b11a835e234000fe40d6d857321a2e8b0ae5b
thanks

entc/integration/customid/customid_test.go
stoikheia: I unified the tests by a way of avove.
Because, fundamental purpose is just checking children IDs. This if-statement is noisy.
https://github.com/ent/ent/pull/2657/commits/94163af0a4d9378d2a6cf7e224e485f5f9975fb9

a8m (approved):
🚀 

Amazing @stoikheia! Thanks for working on it and I'll love to see you on our  [Discord server](https://discord.gg/qZmPgTE6RX).

PR #2665, opened 2022-06-19
entc/gen: allow opening and testing concurrent enttest clients

### LINKED ISSUE #2662: enttest.Open is not goroutine-safe
Concurrent calls to `enttest.Open()` (which you might encounter when parallelizing test runs via `t.Parallel()`) encounter a data race in the lazy/unsychronized initialization of t.columns in `func (m *Migrate) setupTable(t *Table) {`.

a8m:
image

Fixed #2662


tmc:
This change looks reasonable to me. Nice wall time speed up!

Issue #2685, opened 2022-06-24
Invalid imports created when generating code for an object using more than one goType


rltvty:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

might be related to https://github.com/ent/ent/issues/1565

## Current Behavior 😯

In the attached bug-repo, there are two versioned proto packages : 
* `ent_multiple_proto_bug.bar.v1alpha`
* `ent_multiple_proto_bug.foo.v1alpha`

The generated code for these libraries end up at the following paths:
* go package: `barv1alpha` path: `gen/ent-multiple-proto-bug/bar/v1alpha`
* go package: `foov1alpha` path: `gen/ent-multiple-proto-bug/foo/v1alpha`

The *ValueScanner* interface is implemented for both packages in the `*.pb.ent.go` files. 

Both packages are used in the `ent/schema/user.go` object:
```
func (User) Fields() []ent.Field {
	return []ent.Field{
		field.Int("age"),
		field.String("name"),
		field.Bytes("foo").GoType(&foov1alpha.Foo{}),
		field.Bytes("bar").GoType(&barv1alpha.Bar{}),
	}
}
```
(see: https://github.com/rltvty/ent-multiple-proto-bug/blob/master/ent/schema/user.go for full code)

After running `go generate ./ent`,  the generated code has invalid imports and will not compile. In all cases, an extra, non-aliased, import of `"github.com/rltvty/ent-multiple-proto-bug/gen/ent-multiple-proto-bug/foo/v1alpha"` seems to be the issue.

The following files in the `./ent` folder have the issue:
* `mutation.go`: https://github.com/rltvty/ent-multiple-proto-bug/blob/master/ent/mutation.go#L14
* `runtime.go`: https://github.com/rltvty/ent-multiple-proto-bug/blob/master/ent/runtime.go#L6
* `user_delete.go`: https://github.com/rltvty/ent-multiple-proto-bug/blob/master/ent/user_delete.go#L14
* `user_query.go`: https://github.com/rltvty/ent-multiple-proto-bug/blob/master/ent/user_query.go#L15
* `user_update.go`: https://github.com/rltvty/ent-multiple-proto-bug/blob/master/ent/user_update.go#L16
* `user.go`: https://github.com/rltvty/ent-multiple-proto-bug/blob/master/ent/user.go#L12

## Expected Behavior 🤔

[...]

PR #2686, opened 2022-06-24
entc/gen: use gotype pkgname as alias in case it does not match pkgpath

a8m:
Fixed https://github.com/ent/ent/issues/2685


rotemtam (approved):

rotemtam:
Maybe add a test?

a8m:
> Maybe add a test?

Definitely, working on it. See https://github.com/ent/ent/issues/2685#issuecomment-1165939830 😎

a8m:
Thanks for the review @rotemtam

PR #2723, opened 2022-07-06
entc/gen: fix name conflicts in type package (#2548)

### LINKED ISSUE #2548: schema import package name conflicts check not handle major version
Schema import a package conflicts with generate package name, but has major version postfix not been handled.

This function only check base name, not check major version
https://github.com/ent/ent/blob/v0.10.1/entc/gen/type.go#L889
``` go
// aliases adds package aliases (local names) for all type-packages that
// their import identifier conflicts with user-defined packages (i.e. GoType).
func aliases(g *Graph) {
	names := make(map[string]*Type)
	for _, n := range g.Nodes {
		names[n.PackageDir()] = n
	}
	for _, n := range g.Nodes {
		for _, f := range n.Fields {
			if !f.HasGoType() {
				continue
			}
			name := f.Type.PkgName
			if name == "" && f.Type.PkgPath != "" {
				name = path.Base(f.Type.PkgPath)
			}
			// An user-defined type already uses the
			// package local name.
			if n, ok := names[name]; ok {
				// By default, a package named "pet" will be named as "entpet".
				n.alias = path.Base(g.Package) + name
			}
		}
	}
}
``` 
maybe use this function?
https://github.com/golang/tools/blob/v0.1.10/internal/imports/fix.go#L1128
``` go
// ImportPathToAssumedName returns the assumed package name of an import path.
// It does this using only string parsing of the import path.
// It picks the last element of the path that does not look like a major
// version, and then picks the valid identifier off the start of that element.
// It is used to determine if a local rename should be added to an import for
// clarity.
// This function could be moved to a standard package and exported if we want
// for use in other tools.
func ImportPathToAssumedName(importPath string) string {
	base := path.Base(importPath)
	if strings.HasPrefix(base, "v") {
		if _, err := strconv.Atoi(base[1:]); err == nil {
			dir := path.Dir(importPath)
			if dir != "." {
				base = path.Base(dir)
			}
		}
	}
	base = strings.TrimPrefix(base, "go-")
	if i := strings.IndexFunc(base, notIdentifier); i >= 0 {
		base = base[:i]
	}
	return base
}
```
- [x ] The issue is present in the latest release.
- [x ] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯
Generate code with import package conflicts error.

## Expected Behavior 🤔
Package name conflicts with major version been handled.

## Steps to Reproduce 🕹

Steps:

1. schema field with JSON type from a package with major version postfix.
2. generate code.

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.17  |
| Ent         | 0.10.1  |
| Database    | MySQL   |
| Driver      | https://github.com/go-sql-driver/mysql        |

LiuPai:
Fixed #2548

### DIFF
--- a/entc/gen/type.go
+++ b/entc/gen/type.go
@@ -13,8 +13,10 @@ import (
 	"path"
 	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 	"unicode"
+	"unicode/utf8"
 
 	"entgo.io/ent"
 	"entgo.io/ent/dialect/entsql"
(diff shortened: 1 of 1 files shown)


entc/gen/type.go
```diff
@@ -932,7 +963,7 @@ func aliases(g *Graph) {
 			}
 			name := f.Type.PkgName
 			if name == "" && f.Type.PkgPath != "" {
-				name = path.Base(f.Type.PkgPath)
+				name = importPathToAssumedName(f.Type.PkgPath)
```
a8m, line 966: Can't we extract it from the Go identifier, instead of parsing the import path? e.g. `pkg.type` => `pkg`. Because import path does not always reflect the package name.

```go
import (
    "a"   // a.T
    "b/a" // b.T
)
```

The above import is valid if the 2 packages have different names.

PR #2726, opened 2022-07-06
entc/gen: improve edge-schema updates

a8m:
Also, added additional example where an edge schema has another edge
to a type that holds information about the relationship.

The only reason this example exists is to allow users to reduce the storage
occupied by the join-table and allow connect (via M2O) multiple edge-schemas
to an 'information'/'description' node.


Liooo:
tenor-63418738

PR #2730, opened 2022-07-08
feat: add convert to uppercase instead of error

sashamelentyev:


sashamelentyev:
@a8m, i close this?

a8m:
Thanks for submitting this PR @sashamelentyev. I prefer to not guess the schema/type name and ask the user to pass it explicitly.

PR #2735, opened 2022-07-10
dialect/sql/builder: make sql.In() with empty args fallback to False()

Liooo:
Encountered a SQL syntax error when executing `sql.In("col", []interface{})`. The new test case compiles into `"SELECT * FROM "users" WHERE "a" = $1 OR"` without this PR change.

entc generated code at `ent/<modelname>/where.go` has the same `FALSE` fallback logic but `sql.In` didn't, and I'm assuming this is not the expected behavior.

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -1535,9 +1535,9 @@ func In(col string, args ...interface{}) *Predicate {
 
 // In appends the `IN` predicate.
 func (p *Predicate) In(col string, args ...interface{}) *Predicate {
-    if len(args) == 0 {
-        return p.False()
-    }
+	if len(args) == 0 {
+		return p.False()
+	}
 	return p.Append(func(b *Builder) {
 		b.Ident(col).WriteOp(OpIn)
 		b.Nested(func(b *Builder) {
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -1535,7 +1535,7 @@ func In(col string, args ...interface{}) *Predicate {
 
 // In appends the `IN` predicate.
 func (p *Predicate) In(col string, args ...interface{}) *Predicate {
-	// if not arguments were provided, append the FALSE constants,
+	// if no arguments were provided, append the FALSE constants,
 	// since we can't apply "IN ()". This will make this predicate falsy.
 	if len(args) == 0 {
 		return p.False()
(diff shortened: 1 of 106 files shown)


a8m (approved):
Looks good @Liooo! Thanks for working on it.

dialect/sql/builder.go
```diff
@@ -1536,7 +1536,7 @@ func In(col string, args ...interface{}) *Predicate {
 // In appends the `IN` predicate.
 func (p *Predicate) In(col string, args ...interface{}) *Predicate {
 	if len(args) == 0 {
-		return p
+		return p.False()
```
a8m, line 1539:
```suggestion
		// If no arguments were provided, append the FALSE constants,
		// since we can't apply "IN ()". This will make this predicate falsy.
		return p.False()
````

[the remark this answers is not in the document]
Comment (a8m) [reply dialect/sql/builder.go:1539]: We can also remove this if-statement from the codegen.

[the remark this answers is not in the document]
Comment (Liooo) [reply dialect/sql/builder.go]: > We can also remove this if-statement from the codegen.

Really agreed, fixed in d03c9ef 🙏

Comment (Liooo) [reply dialect/sql/builder.go:1541]: thanks added in [f1fcc1c](https://github.com/ent/ent/pull/2735/commits/f1fcc1c9238a3b82ecdcf5e32246007641ab46f3) 👍

File: dialect/sql/builder.go [hunk 2]
```diff
@@ -1535,8 +1535,10 @@ func In(col string, args ...interface{}) *Predicate {
 
 // In appends the `IN` predicate.
 func (p *Predicate) In(col string, args ...interface{}) *Predicate {
+	// if not arguments were provided, append the FALSE constants,
```
a8m, line 1538:
```suggestion
	// If no arguments were provided, append the FALSE constants,
```

entc/gen/template/dialect/sql/predicate.tmpl
```diff
@@ -18,12 +18,6 @@ in the LICENSE file in the root directory of this source tree.
 	{{- $storage := $.Scope.Storage -}}
 	func(s *sql.Selector) {
 		{{- if $op.Variadic }}
-			// if not arguments were provided, append the FALSE constants,
-			// since we can't apply "IN ()". This will make this predicate falsy.
-			if len({{ $arg }}) == 0 {
-				s.Where(sql.False())
-				return
-			}
```
a8m, line 26: Run `go generate ./...` from the root of the project.

a8m (approved):
🚀

Thanks for the contribution @Liooo

Liooo:
@a8m 

thanks for the review 🙌

PR #2737, opened 2022-07-11
add When(), a conditional query builder

gmhafiz:
Lots of files affected but only 5 files are relevant

- delete.tmpl
- query.tmpl
- update.tmpl
- integration_test.go
- crud.md


gmhafiz:
More details on why in discussion section: https://github.com/ent/ent/discussions/2370

PR #2738, opened 2022-07-11
Add Between() and NotBetween() as shorthand functions for SQL

gmhafiz:
Idea floated in Discord: https://discord.com/channels/885059418646003782/885059419107360779/971312595019632751 @masseelch

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -1643,6 +1643,28 @@ func (p *Predicate) Like(col, pattern string) *Predicate {
 	})
 }
 
+// Between is a shorthand for WHERE column BETWEEN ? AND ?
+func Between(col string, start, end interface{}) *Predicate {
+	p := P()
+	return p.Append(func(b *Builder) {
+		b.WriteString(fmt.Sprintf("%s BETWEEN ", col))
+		b.Arg(start)
+		b.WriteString(" AND ")
+		b.Arg(end)
+	})
+}
+
+// NotBetween is a shorthand for WHERE column NOT BETWEEN ? AND ?
+func NotBetween(col string, start, end interface{}) *Predicate {
+	p := P()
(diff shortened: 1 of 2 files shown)


dialect/sql/builder.go
```diff
@@ -1643,6 +1643,28 @@ func (p *Predicate) Like(col, pattern string) *Predicate {
 	})
 }
 
+// Between is a shorthand for WHERE column BETWEEN ? AND ?
+func Between(col string, start, end interface{}) *Predicate {
+	p := P()
+	return p.Append(func(b *Builder) {
+		b.WriteString(fmt.Sprintf("%s BETWEEN ", col))
```
a8m, line 1650: Use `Ident` instead and add tests for both sqlite/mysql and postgres.

a8m (commented):
💯

dialect/sql/builder.go
gmhafiz: done. like so:

```go
func Between(col string, start, end interface{}) *Predicate {
	p := P()
		b.Ident(col)
		b.WriteString(" BETWEEN ")
		b.Arg(start)
		b.WriteString(" AND ")
		b.Arg(end)
	})
}
```

I also added test for sqlite on top of postgres and mysql.


Do you think i should add an `OpBetween` and `OpNotBetween` in `ops`? 

```go
	OpMul:     "*",
	OpDiv:     "/",
	OpMod:     "%",
 	OpBetween:     "BETWEEN",
	OpNotBetween:     "NOT BETWEEN",
```

So it can be this instead
```go
b.Ident(col)
b.WriteOp(OpBetween)
b.Arg(start)
```

gmhafiz: Also does this operation needs to be documented in say `crud.md`?

gmhafiz:
it's ready for review @a8m

PR #2746, opened 2022-07-13
doc/md: fixing enum Status in Add Fields To The Schema

nine-hundred:
When run 'go generate ./ent', got error on [create type Todo: invalid default value for enum field "status"]

Because todo.graphql on tutorial-todo-gql.md says enum Status is "IN_PROGRESS"


a8m (approved):
💯 
Thanks for the contribution @nine-hundred

PR #2757, opened 2022-07-14
dialect/sql/builder: make sql.NotIn() with empty args fallback to NOT FALSE

adayNU:
This is basically the identical change to #2735, but for NotIn(). Also added tests for both cases.

This bug currently prevents anyone using NotIn() from upgrading from v0.10.x to v0.11.x.

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -1587,10 +1587,10 @@ func NotIn(col string, args ...interface{}) *Predicate {
 
 // NotIn appends the `Not IN` predicate.
 func (p *Predicate) NotIn(col string, args ...interface{}) *Predicate {
-	// If no arguments were provided, append the FALSE constant, since
-	// we cannot apply "NOT IN ()". This will make this predicate falsy.
+	// If no arguments were provided, append the NOT FALSE constant, since
+	// we cannot apply "NOT IN ()". This will make this predicate truthy.
 	if len(args) == 0 {
-		return p.False()
+		return Not(p.False())
 	}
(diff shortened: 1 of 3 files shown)


adayNU:
fwiw i also ran `go mod tidy`

a8m:
Hey @adayNU 👋 

A small comment on this behavior, the `x NOT IN (...)` operator is equivalent to `NOT (x IN (...)).` Hence, the predicate should be true and not false in the case of an empty set. Thoughts?

The SQL standard requires at least one value, but SQLite allows this:
```console
sqlite> select 1 in ();
0
sqlite> select 1 not in ();
1
```

> This bug currently prevents anyone using NotIn() from upgrading from v0.10.x to v0.11.x.

Can you elaborate on why?

adayNU:
@a8m sorry for the delay here. Updated based on your feedback.

So for the first question - this is obviously correct lol. I (naively) compared the output of `client.Debug` from v0.10.x and v.0.11.x and saw the difference was that v0.10.x had the `FALSE` keyword. So fwiw, it looks like this was also a bug in v0.10.x, just a different one :) (in v0.11.x you end up with an invalid query, in v0.10.x you end up with a silent bug).

For the latter question, given the statement above, it doesn't really feel like a major blocker given it was incorrect in both versions and you can workaround it in v0.11.x by simply ensuring you don't call `NotIn()` w/o arguments.

a8m (approved):
💯

a8m:
Thanks for the contribution @adayNU 🙏🏻

PR #2778, opened 2022-07-20
docs: Correct expression on documentation

42Atomys:
An invalide sql statement is provider in documentation, just a missing `)`


a8m (approved):
Thanks for the contribution @42Atomys 🙏

PR #2786, opened 2022-07-23
chore: resort imports by goimports rules

sashamelentyev:


a8m:
Thanks for the contribution @sashamelentyev. The current ordering is fine as we usually use this order: std, local (can be configured in goimports) and external. 

Appreciate the contribution but it's not needed in this case. Thanks

Issue #2755, opened 2022-07-14
Calling same relay connection with different aliases is not working


amrnt:
For example, if I have this edge on `Post`

```golang
        edge.To("comments", Comment.Type).
            Annotations(entgql.RelayConnection()),
```

When I call this query, the results for both `publishedComments` and `draftComments` will be same as `publishedComments`.

```graphql
node(id: "xxx") {
  ... on Post {
    title
    body

    publishedComments: comments(where: { status: PUBLISHED }) {
      edges {
        node {
          id
          status
        }
      }
    }

    draftComments: comments(where: { status: DRAFT }) {
      edges {
        node {
          id
          status
        }
      }
    }
  }
}
```

I think this issue was mentioned briefly by @a8m here: https://github.com/ent/contrib/pull/344#pullrequestreview-1020415004

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.18  |
| Ent         | master  |
| Database    | Postgtres   |
| Driver      | pgx  |

a8m:
Thanks for reporting this, @amrnt 🙏 

We currently have a limitation in entql's fields-collection where it expects only one GraphQL field per edge. @giautm and I discussed this and we thought about adding some API to Ent (enabled by a FF) that can be integrated to entgql and solve this issue:

```go
posts, err := client.Post.Query().
    WithNamedComments("published", func(q *ent.CommentQuery) {
        q.Where(comment.StatusEQ(comment.StatusPublished))
    })
    WithNamedComments("draft", func(q *ent.CommentQuery) {
        q.Where(comment.StatusEQ(comment.StatusDraft))
    }).
    Paginate(...)
    
// Then, get each named loaded edge as follows:
for _, p := range posts {
    published, err := p.Edges.NamedComments("published")
    // ...
    draft, err := p.Edges.NamedComments("draft")
    // ...
}
```

This API should be transparent to entgql users and a query like the one above should be loaded correctly in an efficient way.

a8m:
Thanks @amrnt for the great bug report. The issue was addressed both in ent and ent/contrib.

Closed with https://github.com/ent/contrib/pull/355

PR #2790, opened 2022-07-23
entc/gen: move eager-loading to method

a8m:
This is preparation work for `WithNamed<E>` API. For more context, see the proposal at https://github.com/ent/ent/issues/2755.

### DIFF
--- a/entc/gen/template/dialect/sql/query.tmpl
+++ b/entc/gen/template/dialect/sql/query.tmpl
@@ -75,9 +75,13 @@ func ({{ $receiver }} *{{ $builder }}) sqlAll(ctx context.Context, hooks ...quer
 		return nodes, nil
 	}
 	{{- range $e := $.Edges }}
-		{{- with extend $ "Rec" $receiver "Edge" $e }}
-			{{ template "dialect/sql/query/eagerloading" . }}
-		{{- end }}
+		if query := {{ $receiver }}.{{ $e.EagerLoadField }}; query != nil {
+			if err := {{ $receiver }}.load{{ $e.StructField }}(ctx, query, nodes, {{ if and (not $e.M2M) (not $e.O2M) }}nil{{ else }}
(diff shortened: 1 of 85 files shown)


entc/gen/template/dialect/sql/query.tmpl
```diff
@@ -88,6 +92,137 @@ func ({{ $receiver }} *{{ $builder }}) sqlAll(ctx context.Context, hooks ...quer
 	return nodes, nil
 }
 
+{{/* Generate a method to eager-load each edge. */}}
+{{- range $e := $.Edges }}
+	func ({{ $receiver }} *{{ $builder }}) load{{ $e.StructField }}(ctx context.Context, query *{{ $e.Type.QueryName }}, nodes []*{{ $.Name }}, init func(*{{ $.Name }}), assign func(*{{ $.Name }}, *{{ $e.Type.Name }})) error {
+		{{- if $e.M2M }}
+			edgeids := make([]driver.Value, len(nodes))
+			byid := make(map[{{ $.ID.Type }}]*{{ $.Name }})
```
masseelch, line 100:
```suggestion
			byids := make(map[{{ $.ID.Type }}]*{{ $.Name }})
```

Stay consistent

masseelch (commented):
:100:

masseelch (approved):

entc/gen/template/dialect/sql/query.tmpl
giautm: Should be `edgeIDs` and `byIDs`

giautm (approved):
💯

entc/gen/template/dialect/sql/query.tmpl
a8m: I actually prefer `byid` as it describes the operation for getting a node by its ID and that's why this variable is used for.  I use this singular form quite often in ent and atlas. I'll keep it this way, but will change the naming to camelCase.

PR #2793, opened 2022-07-24
chore: use error.New when error without format

sashamelentyev:


sashamelentyev:
@a8m, hello!
Please forgive me, but this case again

Issue #892, opened 2020-10-28
doc: Hook Helpers example doesn't work due to circular dependency


mgabeler-lee-6rs:
The Hook Helpers example here won't work: https://github.com/facebook/ent/blob/master/doc/md/hooks.md#hook-helpers

The example shows importing the _generated_ `hook` package from the `schema` package. But to load the schema package to run generation, it has to parse OK ... which won't work if generation hasn't run.  If you try, you get an error like this:
```
load schemas dir: <project-dir>/ent/schema/counter.go:7:2: could not import <project>/ent/hook (invalid package name: "")
```

Thus, the circular dependency.

AFAICT, the hook helpers in their current form cannot ever be used from schema or mixin hooks (absent maybe some weird forms of indirection?) due to this.

However it also seems worth observing that the majority of the generated `hook.go` has no dependency on the individual schema, only on the generic interfaces & constants from the main `ent` package. These could be related to a package in the main `ent` module (with a different name to avoid collisions) and thus most of the circular dependency problems would be resolved. This would also have the advantage of being able to write mixin hooks leveraging them in shared code that isn't dependent on the schema in use.

a8m:
Hey @mgabeler-lee-6rs!

Can you please report what version do you use at the moment for ent? 

> AFAICT, the hook helpers in their current form cannot ever be used from schema or mixin hooks (absent maybe some weird forms of indirection?) due to this.

Please see the yellow note in this doc [entgo.io/docs/hooks/#schema-hooks](https://entgo.io/docs/hooks/#schema-hooks), and the testing examples ([1](https://github.com/facebook/ent/blob/master/entc/integration/hooks/ent/schema/user.go), [2](https://github.com/facebook/ent/blob/master/entc/integration/hooks/ent/schema/card.go)) that use the hook API in the schema.

Now, about circular dependency, let me explain how the schema-hooks work in ent, and you're more than welcome to propose improvements to the existing solution.
1. If users use hooks (or privacy) in the schema, we assume they may import the generated ent-package, and the runtime-stitching logic will be generated to a new package named `runtime` to avoid the circular dependency scenario  
2. Users need to empty import the `runtime` package in the main of the application to register these hooks (see [`enttest`](https://github.com/facebook/ent/blob/master/entc/integration/hooks/ent/enttest/enttest.go#L14) for example)


[...]

PR #2801, opened 2022-07-26
Add `BuildFlags` to schema loader configuration

joewreschnig:
This allows passing arbitrary build flags to the schema package loader,
which in turn can be used to pass build constraints affecting what
schema types and/or methods are visible.

As a particular use case, this should make it easier to implement
two-pass code generation to deal with circular imports in hooks (#892).


a8m (approved):
Thanks for adding this functionality, and sorry for the delayed review, @joewreschnig.

 > As a particular use case, this should make it easier to implement
two-pass code generation to deal with circular imports in hooks (https://github.com/ent/ent/issues/892).

Would you mind commenting on #892 with your workaround for this? e.g. how you pass the flags, example config, etc.

Thanks 🙏

PR #2847, opened 2022-08-13
entc/gen: allow setting required fields as nillable

a8m:


doc/md/schema-fields.md
```diff
@@ -439,13 +439,12 @@ func (User) Fields() []ent.Field {
...
-Sometimes you want to be able to distinguish between the zero value of fields
-and `nil`; for example, if the database column contains `0` or `NULL`.
-The `Nillable` option exists exactly for this.
+Sometimes you want to be able to distinguish between the zero value of fields and `nil`.
+For example, if the database column contains `0` or `NULL`.  The `Nillable` option exists exactly for this.
 
 If you have an `Optional` field of type `T`, setting it to `Nillable` will generate
 a struct field with type `*T`. Hence, if the database returns `NULL` for this field,
-the struct field will be `nil`. Otherwise, it will contain a pointer to the actual data.
+the struct field will be `nil`. Otherwise, it will contain a pointer to the actual value.
```
Albert-Gao, line 447: Hi, can we add an example in terms of the impact of this `Nillable()` call? From the perspective of json serialisation.

For example, what does the final json look like, before and after this `Nillable()` call, as a reader, I can understand the impact to the generated Struct, but not quite sure about the post-jsonify stage. :) would be better if we have this information here, thanks :)

PR #2860, opened 2022-08-15
dialect/sql: escape EqualFold on Postgres

crossworth:
Hello 👋, `EqualFold` currently does not escape the argument, this can lead to problems, like returning multiple results when the first/last character of the argument is `%` and `pq: LIKE pattern must not end with escape character` when the first/last  character of the argument is a escape character.

While at it, I revised the points raised on #1484 (from the Postgres point of view). There seems to be a lot of different information about the use of `ILIKE` vs `LOWER` and I could not reach a conclusion.

I think we just leave as is but we should keep in mind that may exists a faster way available, when we have a user that needs it, we could look at it again.

- https://stackoverflow.com/a/41691993
- https://dba.stackexchange.com/questions/89901/lower-vs-ilike-on-postgresql
- https://til.hashrocket.com/posts/e65bb49204-lower-is-faster-than-ilike


a8m (approved):
💯 

Thanks, Pedro!

PR #2885, opened 2022-08-29
dialect/sql/builder: added FullJoin

Liooo:
please review 🙏

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -2331,6 +2331,11 @@ func (s *Selector) RightJoin(t TableView) *Selector {
 	return s.join("RIGHT JOIN", t)
 }
 
+// FullJoin appends a `Full JOIN` clause to the statement.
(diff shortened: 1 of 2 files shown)


dialect/sql/builder.go
```diff
@@ -2331,6 +2331,11 @@ func (s *Selector) RightJoin(t TableView) *Selector {
 	return s.join("RIGHT JOIN", t)
 }
 
+// FullJoin appends a `Full JOIN` clause to the statement.
```
a8m, line 2334:
```suggestion
// FullJoin appends a `FULL JOIN` clause to the statement.
```

a8m (approved):
💯 

Thanks for the contribution , @Liooo

dialect/sql/builder.go
Liooo: 🙏

a8m (approved):
🚀

PR #2890, opened 2022-08-30
entc/gen: remove repeated template fragment

Weilence:


a8m (approved):
Thanks @Weilence 👏🏻 🚀

PR #2900, opened 2022-09-06
entc/gen: support custom Acronym

Weilence:


a8m (approved):
💯 
Thanks for the contribution @Weilence

Issue #2761, opened 2022-07-16
ent generates invalid code when resolving the pkg name to Go map type


ivanvanderbyl:
- [X] The issue is present in the latest release.
- [X] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

After upgrading from 0.10.1 to 0.11.1, when I regenerate my app I get this weird output in the dependency imports of some files e.g. `<T>_delete.go`:

```go
import (
	"context"
	"errors"
	"fmt"
	"math"
	"strings"
	"time"
		"github.com/airheartdev/api/ent/predicate"
			map[string]models "github.com/airheartdev/api/internal/models"
			"github.com/airheartdev/api/internal/models"
			"github.com/airheartdev/api/internal/models"
	"entgo.io/ent"
	"entgo.io/ent/dialect"
			"entgo.io/ent/dialect/sql"
			"entgo.io/ent/dialect/sql/sqlgraph"
			"entgo.io/ent/schema/field"
		stdsql "database/sql"

)
```

This seems to be a result of having a JSON field that uses a map type that has a value in another pkg, for example:

```go
field.JSON("occupancy_pricing", map[string]models.OccupancyPricing{})
```

I've bisected all the commits and this regression was introduced with https://github.com/ent/ent/pull/2686

## Expected Behavior 🤔

It generates the correct pkg import name.

## Workaround

A quick workaround for anyone else coming across this: define a custom map type:

```go
type OccupancyPricingMap map[string]OccupancyPricing
```

cc @a8m

a8m:
Thanks for reporting this @ivanvanderbyl 🙏 

Updating to the latest master should solve the issue.

ivanvanderbyl:
Thanks for the quick fix @a8m

PR #2926, opened 2022-09-13
entc/gen: allow package with alias on ID field

crossworth:
Hello, was reported in Discord [1](https://discord.com/channels/885059418646003782/885059419107360779/1018909745165381652), [2](https://discord.com/channels/885059418646003782/1018920317495820359/1019235767760994444), that it's not possible to use `github.com/oklog/ulid/v2` as a `UUID` type.

The example using `ent/bug` worked correctly, but after updating to the latest version a problem with [duplicated imports](https://github.com/crossworth/ulid-as-uuid/blob/ee4f3c51e8ba3f0cdac547e746d05a72acd89b11/ent/user.go#L11-L12) (one with the alias and the other without) occurs.

This seems related to #2765 (could not test it).

Any way this PR tries to solve this problem, it also adds a _test_ for the problem of #2761 to ensure we don't get a regression.

NOTE:
- Maybe is possible to create an action that updates `ent/bug` to use the latest version of Ent.
- The only reason I created the `entgen` folder is to be able to debug the code, let me know if I should remove it.


crossworth:
It seems I'm having trouble with the generate command, it's adding new lines on my side I guess.

entc/integration/ulid/ent/schema/mixin/id_mixin.go
```diff
@@ -0,0 +1,18 @@
+package mixin
+
+import (
+	"entgo.io/ent"
+	"entgo.io/ent/schema/field"
+	"entgo.io/ent/schema/mixin"
+	"github.com/oklog/ulid/v2"
```
a8m, line 7: A simpler solution for testing it and also to avoid creating a new module and a directory is to create a dummy package under `integration/ent/schema/*` and use it there. It will also allow us to test it without downloading `oklog/ulid` as a dependency. WDYT?

crossworth: I started with that, I created a custom package only to be imported by the test, but after while I thought that it might be better to use the `ulid` package since it was what the user wanted and it would work as an example of using a compatible uuid type/`ulid`.

But I don't really have a opinion about it, I guess it's better to avoid having to download it only to run the tests.

I'm not sure about the part of avoiding creating a new module, what I want to test is packages that use version like `v2` on the import path, I think I would need to create a new module and if I create a new module, it will be added to the upper `go.mod`, so I have to create yet another module on the "local test".

Or maybe I'm missing something.

Example using local packages (to compare if its better or I can do something different):
- https://github.com/crossworth/ent/tree/import-bug-a/entc/integration/packagewithalias

Let me know what you think.

a8m: Hey @crossworth! I didn't have enough time to dig into this but would love to have this fix added. Regarding the tests, we already have this logic in line 47 and it works fine, right? I think we have a similar test that covers this here: https://github.com/ent/ent/blob/master/entc/integration/ent/schema/dir/dir.go#L7, because the reflection package path does not aware of modules.

crossworth: Hello again, I'm not sure if I understood your message (could not find the line 47), but I have checked and the problem is present without the modules, so I have made changes to keep this PR simpler (and without new dependencies).

Please, see if you this version is better, on my side I think it is, it still have the same objectives but uses less lines of code.

entc/gen/template/import.tmpl
```diff
@@ -19,7 +19,8 @@ import (
 		"{{ $.Config.Package }}/predicate"
 		{{- if $.HasOneFieldID }}
 			{{- with $.ID.Type.PkgPath }}
-				"{{ . }}"
+				{{- $name := $.ID.Type.PkgName }}
+				{{ if ne $name (base $.ID.Type.PkgPath) }}{{ $name }} {{ end}}"{{ $.ID.Type.PkgPath }}"
```
giautm, line 23:
```suggestion
				{{ if ne $name (base $.ID.Type.PkgPath) }}{{ $name }} {{ end}}"{{ . }}"
```

giautm (approved):
Tested with my reproduce the bug. It removes the duplicate import.

https://github.com/giautm/ent-bug-ulid/commit/7dd035263e4f087e724791278795ed250884c7a6

a8m (approved):
💯 

Looks good. See @giautm’s comment.

a8m:
Thanks, @crossworth ❤️🙏🏻

PR #2935, opened 2022-09-18
doc: minor fix Japanese README

eltociear:
* URL fix
* Correct your expressions more fluently


a8m:
Thanks for the contribution @eltociear

PR #2942, opened 2022-09-20
doc/md: add atlas migration import to versioned docs

masseelch:


a8m (approved):
Looks good, but let's remember to remove it in the next version or the one after.

PR #2958, opened 2022-09-26
doc: make mutation and variable match todo.graphql

kortschak:
Please take a look. This fixes the mutation query to match the todo.graphql file. It also removes a section in the gqlgen.yml that doesn't appear during the tutorial.

Curiously, the diff section in the gqlgen.yml file is already present, so I am not entirely sure what should be happening there.


a8m (approved):
💯 

Thanks, @kortschak

PR #2960, opened 2022-09-26
doc: make mutation and variable match todo.graphql

kortschak:
Please take a look. Same change as for the previous page.


a8m (approved):
Thanks, Dan 🙏🏻

PR #2961, opened 2022-09-26
docs: fix file name for required change

kortschak:
Please take a look.

(As an aside, I did not get what this page was trying to get across as the code as provided did not generate 11 queries and there is no indication that I could see on the page for how to alter the **disabling fields collection** behaviour).


a8m (approved):
Thanks for the fix @kortschak.

> (As an aside, I did not get what this page was trying to get across as the code as provided did not generate 11 queries and there is no indication that I could see on the page for how to alter the disabling fields collection behaviour).

Yes, this document was written at the begging before field collection was enabled by default. It can be disabled using the `Unbind()` option, and I'll find the time to update this doc.

PR #2965, opened 2022-09-27
docs/md: various fixes

kortschak:
This is a collection of fixes that look like are needed. I have some doubts because some fixes here are added in later pages (notable example being the addition of `entgql.WithSchemaGenerator()` to entc.go in doc/md/tutorial-todo-gql-filter-input.md, which is required to make entc.go successfulliy complete, but which is added in the tutorial in a later page, https://entgo.io/docs/tutorial-todo-gql-schema-generator — though there are others that I've found). My suspicion is that the page order has changed over the construction of the tutorial.

Please take a look.

### DIFF
--- a/doc/md/tutorial-todo-gql-filter-input.md
+++ b/doc/md/tutorial-todo-gql-filter-input.md
@@ -47,14 +47,15 @@ go run ./cmd/todo/
 
 ### Configure Ent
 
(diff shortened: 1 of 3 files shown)


a8m (approved):
💯 

Appreciate the help, @kortschak.

PR #2972, opened 2022-09-28
dialect/sql/sqljson: add dialect-aware Append function

a8m:


dialect/sql/sqljson/dialect.go
```diff
@@ -0,0 +1,182 @@
+// Copyright 2019-present Facebook Inc. All rights reserved.
+// This source code is licensed under the Apache 2.0 license found
+// in the LICENSE file in the root directory of this source tree.
+
+package sqljson
```
a8m, line 5: We got to the stage where drivers have more differences than in common, and I plan to move the rest of the dialect-specific code to this file.

giautm (approved):
💯

PR #2994, opened 2022-10-07
dialect/sql: add Window function to be able to customize selector portion for window functions

timoha:


dialect/sql/builder.go
```diff
@@ -2960,7 +2960,16 @@ type WindowBuilder struct {
...
-	return &WindowBuilder{fn: "ROW_NUMBER"}
+	return WindowSelector(Select("ROW_NUMBER()"))
+}
+
+// WindowSelector returns a new window clause with a custom selector allowing
+// for custom windown functions.
+//
+//	s1 := WindowSelector(SelectExpr(Expr("ROW_NUMBER()")))
+//	s2 := WindowSelector(Select("ROW_NUMBER()"))
+func WindowSelector(s *Selector) *WindowBuilder {
```
a8m, line 2971: The first argument/grammar-rule holds the window function, right? So I'm not sure using a `Selector` is right here. 

What do you think about implementing it like predicates? i.e. `P(func(*sql.Builder)`. The reason it gets a function and not a builder is that the function is invoked only on `Query()` call and may use the same dialect-builder.

```go
func RowNumber() *WindowBuilder {
	return Window(func(b *sql.Builder) {
		b.WriteString("ROW_NUMBER")
	})
}
```

We can create helper functions like `RowNumber` in the future if we feel the API is not intuitive.

a8m (commented):
Thanks for the contribution, @timoha ❤️

dialect/sql/builder.go
timoha: Updated! let me know if it's something you are looking for

dialect/sql/builder.go
```diff
@@ -2951,7 +2951,7 @@ func (*WithBuilder) view() {}
 // only to query rows-limited edges in pagination.
 type WindowBuilder struct {
 	Builder
-	fn        string // e.g. ROW_NUMBER(), RANK().
+	selector  func(*Builder)
```
a8m, line 2954:
```suggestion
	fn  func(*Builder) // e.g. ROW_NUMBER(), RANK().
```
It's not a selector anymore.

dialect/sql/builder.go
```diff
@@ -2960,7 +2960,19 @@ type WindowBuilder struct {
...
+}
+
+// Window returns a new window clause with a custom selector allowing
+// for custom windown functions.
+//
+//	Window(func(b *Builder) {
+//		b.WriteString(Sum(posts.C("duration")))
+//	}).PartitionBy("author_id").OrderBy("id"), "duration").
+func Window(selector func(*Builder)) *WindowBuilder {
+	return &WindowBuilder{selector: selector}
```
a8m, line 2975:
```suggestion
func Window(fn func(*Builder)) *WindowBuilder {
	return &WindowBuilder{fn: fn}
```

dialect/sql/builder.go
```diff
@@ -3000,8 +3012,8 @@ func (w *WindowBuilder) OrderExpr(exprs ...Querier) *WindowBuilder {
 
 // Query returns query representation of the window function.
 func (w *WindowBuilder) Query() (string, []any) {
-	w.WriteString(w.fn)
-	w.WriteString("() OVER ")
+	w.selector(&w.Builder)
```
a8m, line 3015:
```suggestion
	w.fn(&w.Builder)
```

dialect/sql/builder_test.go
```diff
@@ -2146,6 +2146,22 @@ func TestWindowFunction(t *testing.T) {
 	require.Equal(t, []any{2}, args)
 }
 
+func TestWindowFunction_Select(t *testing.T) {
+	posts := Table("posts")
+
```
a8m, line 2151:
```suggestion
```

a8m (approved):
Looks good! Please, fix the minor comments, and I'll merge it.

Thanks for the contribution. It's appreciated.

PR #2996, opened 2022-10-08
Remove unused variable and type redundancy

marwan-at-work:
Hi, this is a small PR that does two things:
1. Removes the `ctx` variable from the query.path function since it is being unused.
2. Changes `struct{}{}` to `{}` in map assignment. 

Mainly to make my linter happy and since this is a generated file, I can't put ignore-comments on top of those lines.


a8m (approved):
💯 

Thanks Marwan!

PR #3050, opened 2022-10-28
add note to import correct modules in tutorial

niwla23:
This adds a note to import the correct modules when adding the fields to the tutorial.

Just had to take a look at the examples to find the right module imports so I thought it would be good to have it in the tutorial.

### DIFF
--- a/doc/md/tutorial-todo-crud.md
+++ b/doc/md/tutorial-todo-crud.md
@@ -50,6 +50,16 @@ func (Todo) Fields() []ent.Field {
 }
 ```
 
+Update your imports to look like this:
+```
+import (
+	"time"
+
+	"entgo.io/ent"
+	"entgo.io/ent/schema/field"
+)
+```
+
 After adding these fields, we need to run the code-generation as before:
 
 ```console
--- a/doc/md/tutorial-todo-crud.md
+++ b/doc/md/tutorial-todo-crud.md
@@ -51,7 +51,7 @@ func (Todo) Fields() []ent.Field {
 ```
 
 Update your imports to look like this:
-```
+```go
 import (
 	"time"
 


doc/md/tutorial-todo-crud.md
```diff
@@ -50,6 +50,16 @@ func (Todo) Fields() []ent.Field {
...
+Update your imports to look like this:
+```
+import (
+	"time"
+
+	"entgo.io/ent"
+	"entgo.io/ent/schema/field"
+)
+```
+
```
a8m, line 62: Can you edit the example above instead?

a8m (commented):
Thanks for the contribution, @niwla23. See the comment below 🙏

doc/md/tutorial-todo-crud.md
niwla23: It would not make sense because the example is only a snippet. It is missing the struct defintion, so adding the imports on top would be confusing I think

a8m: Most users use `goimports`, but if you still think it's necessary, let's keep it as one code block. Thanks 🙏 
~~~
```go title="ent/schema/todo.go"
package schema

import (
	"time"
	"entgo.io/ent"
	"entgo.io/ent/schema/field"
)

 // ...

func (Todo) Fields() []ent.Field {
	// ...
}
```
~~~

niwla23: :+1: I am quite new to go and didn't know it can auto import so I got confused by it

a8m: Checkout https://pkg.go.dev/golang.org/x/tools/cmd/goimports and let me know if we can close this one. Thanks

niwla23: yeah I think it can be closed then

a8m: Thanks 🙏🏻

PR #3051, opened 2022-10-28
dialect/sql/sqlgraph: handle edge-schema in M2O/O2Oi calls entql

a8m:


a8m:
> @a8m tested on our codebase and confirmed this fixes the panic we were seeing!

Thanks for verifying this, @adayNU 🙏🏻

PR #3052, opened 2022-10-29
fix comments lost in generated DDL.

chaosue:


a8m (commented):
Thanks for adding this, @chaosue. It's appreciated! 🙏 

I think this change should be added to [EntSQL annotation](https://github.com/ent/ent/tree/master/dialect/entsql), because I prefer not to trigger migration changes to all users that use the `Comment` option today. Also, some users, like myself, care only about the code comments and do not need database comments.

a8m:
Hey @chaosue! Any update on this PR?

a8m:
After a discussion with Cheng on Discord, we think the configuration can be simplified by providing a single annotation in the `entsql` package to indicate that columns' comments should also be stored on the database schema. For example:

```go
func (T) Annotations() []schema.Annotation {
    return []schema.Annotation{
        // WithComments indicates that all entity comments
        // should be stored in the database schema as well.
        entsql.WithComments(true),
    }
}
```

Table comments can be implemented once https://github.com/ent/ent/pull/3164 is merged, and an `entsql.Comment(string)` annotation can be added in case there are requests for mixed comments (one for code generation, and one for the database).

huiwanggo:
`entsql.Comment(string)`  When will this feature be implemented?

huiwanggo:
table comment ? entsql.WithComments(true), schema.Comment(string) SQL DDL not work。Will this feature be supported? @a8m

suncle1993:
> 

if you want to generate table comment and column comment, you can upgrade your entgo to master version. And then use like this:

```go
type T struct {
	ent.Schema
}

func (T) Annotations() []schema.Annotation {
	return []schema.Annotation{
		entsql.Annotation{Table: "t"},
		entsql.WithComments(true),
		schema.Comment("t is like business unit table"),
	}
}

func (T) Fields() []ent.Field {
	return []ent.Field{
		field.String("name").Comment("unique name"),
		field.String("description").Default("").Comment("xxx"),
	}
}
```

suncle1993:
> table comment ? entsql.WithComments(true), schema.Comment(string) SQL DDL not work。Will this feature be supported? @a8m


if you want to generate table comment and column comment, you can upgrade your entgo to master version. And then use like this:

```go
type T struct {
	ent.Schema
}

func (T) Annotations() []schema.Annotation {
	return []schema.Annotation{
		entsql.Annotation{Table: "t"},
		entsql.WithComments(true),
		schema.Comment("t is like business unit table"),
	}
}

func (T) Fields() []ent.Field {
	return []ent.Field{
		field.String("name").Comment("unique name"),
		field.String("description").Default("").Comment("xxx"),
	}
}
```

PR #3071, opened 2022-11-08
Add encrypt function

ugurkorkmaz:
I added a simple method to encrypt data of type String.


rotemtam:
Hi there @ugurkorkmaz 

Thanks for the suggestion.

Encryption is a big topic which personally I believe calls for a large API surface to work in production settings.

We have an open community discussion on the topic you're more than welcome to join:

https://github.com/ent/ent/discussions/2364

Issue #3069, opened 2022-11-07
Default annotations aren't working if the default value is a literal


Southclaws:
The following field produces no default value:

```go
field.Text("kyc_result").Annotations(&entsql.Annotation{Default: "'not_initiated'::text"}),
```

Whereas these produce the desired result:

```go
...Annotations(&entsql.Annotation{Default: "CURRENT_TIMESTAMP"}),
...Annotations(&entsql.Annotation{Default: "gen_random_uuid()"})
```

- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

![image](https://user-images.githubusercontent.com/1636971/200363649-9e6c897d-dc56-44b5-8005-0507316c13aa.png)

## Expected Behavior 🤔

![image](https://user-images.githubusercontent.com/1636971/200363675-84fe9373-0873-406f-b900-2afa879e9b28.png)

## Steps to Reproduce 🕹

Just spin up an example schema with a `Text` field that has an annotated default value as in the above snippet.

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.19  |
| Ent         | entgo.io/ent v0.11.5-0.20221105052850-a91aabe3ef4c  |
| Database    | Postgres 13 |
| Driver      | pgx        |

a8m:
Thanks for suggesting this, @Southclaws. This was resolved with #3085.

Please, see the examples there (or in the doc) and let me know if you still need help with it 🙏

Southclaws:
Awesome, thanks!

PR #3085, opened 2022-11-11
dialect/sql/entsql: support setting expression as column default value

a8m:
Fixed https://github.com/ent/ent/issues/3069


giautm (approved):
💯

PR #3113, opened 2022-11-24
dialect/sql: add except and intersect for builder

Liooo:
This PR adds EXCEPT and INTERSECT set operators.

- didn't add any dialect checking in this PR, what do you think?
    - EXCEPT and INTERSECT was added in [MySQL 8.0.31](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-31.html#mysqld-8-0-31-sql-syntax)
    - some clause combinations aren't supported in sqlite (e.g. `UNION DISTINCT` and `SUBTRACT ALL`)
   - I'm not really sure about gremlin, please tell me if there's anything to be done

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -2138,7 +2138,7 @@ type Selector struct {
 	limit     *int
 	offset    *int
 	distinct  bool
-	union     []union
+	setops    []setOperation
 	prefix    Queries
 	lock      *LockOptions
 }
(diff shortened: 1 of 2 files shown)


dialect/sql/builder.go
```diff
@@ -2424,42 +2424,112 @@ func (s *Selector) join(kind string, t TableView) *Selector {
...
-// unionType describes an UNION type.
-type unionType string
+// dupStrategy describes set operation's duplication handling strategy
+type dupStrategy string
 
 const (
-	unionAll      unionType = "ALL"
-	unionDistinct unionType = "DISTINCT"
+	dupStrategyAll      dupStrategy = "ALL"
+	dupStrategyDistinct dupStrategy = "DISTINCT"
```
Liooo, line 2432: if there's any better name 🙏

a8m:
I added the SQLite checks, removed/renamed some parts, and merged to master.
The commit is also on you:

image

I preferred to do this in code review, but since it was urgent to you, I didn't want to block you that much and avoid the CR loop. Thanks for the contribution; it's appreciated ❤️

Liooo:
@a8m thank you 🎉

Issue #2594, opened 2022-06-01
Two primary keys?


frederikhors:
I'm wondering if it's possible to have two Primary Keys for some tables, like:

```sql
CREATE TABLE "public"."players" (
  "id" int8 NOT NULL DEFAULT nextval('players_id_seq'::regclass),
  "tenant_id" int8 NOT NULL,
  "name" varchar COLLATE "pg_catalog"."default" NOT NULL,
  CONSTRAINT "players_pkey" PRIMARY KEY ("tenant_id", "id"),
  CONSTRAINT "players_tenant_id_fkey" FOREIGN KEY ("tenant_id") REFERENCES "public"."tenants" ("id") ON DELETE NO ACTION ON UPDATE NO ACTION
)
```

I cannot find any info for this.

a8m:
Hey @frederikhors 👋 
Ent does not support composite primary keys for nodes currently (unlike [edge schemas](https://entgo.io/docs/schema-edges#edge-schema)). 

Technically, the implementation does not require significant changes in Ent because edge schemas already support it. However, I decided not to support it atm (until we'll finalize the design) because it won't play nicely with many of the integration we provide today with Ent. e.g. `entgql` - a `Node` in the graph must have an ID (it's also called ID in your tables). If users want to set up a unique constraint on two or more fields for now, they can use [index options](https://entgo.io/docs/schema-indexes#multiple-fields). 

I think we can close this as it's a duplicate of https://github.com/ent/ent/issues/400.

PR #3114, opened 2022-11-24
dialect/sql: avoid setting the primary key when no primary key is present

crossworth:
I'm playing with [Timescale](https://www.timescale.com/), Timescale is a PostgreSQL extension that adds/improves support for time series data.

Timescale works by converting a normal table into a `hypertable`, it's not possible to use Timescale with Ent right now because the primary key of the ID field cause problems.

More info: 
- https://docs.timescale.com/timescaledb/latest/how-to-guides/schema-management/about-constraints/
- https://stackoverflow.com/a/68066535

One solution is avoiding creating the primary key in the first place, to do this on the user side of things a migration hook can be used:

```go
func WithoutPrimaryKey(tableName string) schema.MigrateOption {
	return schema.WithHooks(func(next schema.Creator) schema.Creator {
		return schema.CreateFunc(func(ctx context.Context, tables ...*schema.Table) error {
			for i := range tables {
				if tables[i].Name == tableName {
					tables[i].PrimaryKey = nil
				}
			}
			return next.Create(ctx, tables...)
		})
	})
}
```

But this generates invalid migration code `PRIMARY KEY ()`, this PR solves this problem and allows the use of the migration hook above without problems.

Maybe in the future we could add `WithoutPrimaryKey` as a official option like `WithDropColumn` is today (but I think is better wait to see if there is a need for it, I dont think there is huge user base with tables without primary keys)

It's worth noting that another solution would be to allow more control of the primary key on Ent schema, but based on [this comment](https://github.com/ent/ent/issues/2594#issuecomment-1145056254), seems like there is other's considerations that should be analyzed first and this PR at least solves the problem in a easy way.

### DIFF
--- a/dialect/sql/schema/atlas.go
+++ b/dialect/sql/schema/atlas.go
@@ -979,7 +979,9 @@ func (a *Atlas) aIndexes(et *Table, at *schema.Table) error {
 		}
 		pk = append(pk, c2)
 	}
-	at.SetPrimaryKey(schema.NewPrimaryKey(pk...))
+	if len(pk) > 0 {
+		at.SetPrimaryKey(schema.NewPrimaryKey(pk...))
+	}
 	// Rest of indexes.
 	for _, idx1 := range et.Indexes {
 		idx2 := schema.NewIndex(idx1.Name).
--- a/dialect/sql/schema/atlas.go
+++ b/dialect/sql/schema/atlas.go
@@ -979,6 +979,7 @@ func (a *Atlas) aIndexes(et *Table, at *schema.Table) error {
 		}
 		pk = append(pk, c2)
 	}
+	// CreateFunc might clear the primary keys.
 	if len(pk) > 0 {
 		at.SetPrimaryKey(schema.NewPrimaryKey(pk...))
 	}
--- a/dialect/sql/schema/migrate_test.go
+++ b/dialect/sql/schema/migrate_test.go
@@ -246,7 +246,7 @@ func TestMigrate_Diff(t *testing.T) {
 	require.NoError(t, err)
 	require.NoError(t, m.Diff(ctx, &Table{Name: "users"}))
 	v := time.Now().UTC().Format("20060102150405")
-	requireFileEqual(t, filepath.Join(p, v+"_changes.up.sql"), "-- create \"users\" table\nCREATE TABLE `users` (, PRIMARY KEY ());\n")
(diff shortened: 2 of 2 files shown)


giautm (approved):
LGTM, 💯

dialect/sql/schema/atlas.go
```diff
@@ -979,7 +979,9 @@ func (a *Atlas) aIndexes(et *Table, at *schema.Table) error {
 		}
 		pk = append(pk, c2)
 	}
-	at.SetPrimaryKey(schema.NewPrimaryKey(pk...))
+	if len(pk) > 0 {
+		at.SetPrimaryKey(schema.NewPrimaryKey(pk...))
+	}
```
a8m, line 984: Using DiffHooks, the primary keys can be cleaned before generating the changes. Can you try this also? 

```go
schema.WithDiffHook(func(next schema.Differ) schema.Differ {
	return schema.DiffFunc(func(current, desired *atlas.Schema) ([]atlas.Change, error) {
		if t, ok := current.Table(name); ok {
			t.PrimaryKey = nil
		}
		if t, ok := desired.Table(name); ok {
			t.PrimaryKey = nil
		}
		next.Diff(current, desired)
	})
})
```

I don't mind adding this change, but I prefer having an explicit comment explaining that it might be cleared by the user using a `CreateHook` (the old API).

a8m:
> Maybe in the future we could add WithoutPrimaryKey as a official option like WithDropColumn is today (but I think is better wait to see if there is a need for it, I dont think there is huge user base with tables without primary keys)

In case we'll add this in the future, I would suggest that it should be defined on the ent/schema and not on runtime. WDYT?

crossworth:
`WithDiffHook` works great! 

I have added a comment as requested, but if there is plans to remove the legacy API in the near future, I think we can close this PR without merging it.

The new API seems more powerful and more easy to work it, now that I learned about it I will use all the time 🤣.

> In case we'll add this in the future, I would suggest that it should be defined on the ent/schema and not on runtime. WDYT?

I agree! The "better" way is allowing customization of the primary key on ent/schema, mostly to support composite keys, I don't see any benefits from not-having a primary key[1], the only reason I needed is because Timescale/Postgres partition requirements.

Anyway, feel free to close or merge this PR as seems fit, I will change my examples to use the `WithDiffHook` API for now.

1 - Refs:
- https://dba.stackexchange.com/a/190678
- https://stackoverflow.com/a/840182
- https://stackoverflow.com/a/39650093
- https://stackoverflow.com/a/24215524

a8m (approved):
Looks good. I’m OK with merging this as there are still users that use the previous hook API.

PR #3115, opened 2022-11-26
Fuzzy query: Problems occurred when I used "_汉字" for fuzzy query

cheny-alf:
Since a Chinese character takes up three bytes, we cannot only return the index of the first byte of each Chinese character


dialect/sql/builder.go
```diff
@@ -1689,11 +1689,11 @@ func escape(w string) (string, bool) {
...
 	var b strings.Builder
 	b.Grow(len(w) + n)
-	for i := range w {
+	for i, character := range w {
 		if c := w[i]; c == '%' || c == '_' || c == '\\' {
 			b.WriteByte('\\')
 		}
-		b.WriteByte(w[i])
+		b.WriteRune(character)
 	}
```
a8m, line 1697:
```suggestion
	for _, c := range w {
		if c == '%' || c == '_' || c == '\\' {
			b.WriteByte('\\')
		}
		b.WriteRune(c)
	}
````
Isn't this simpler?

File: dialect/sql/builder_test.go [hunk 2]
```diff
@@ -2067,6 +2067,12 @@ func TestEscapePatterns(t *testing.T) {
 		Query()
 	require.Equal(t, "UPDATE `users` SET `name` = NULL WHERE `nickname` LIKE ? ESCAPE ? OR `nickname` LIKE ? ESCAPE ? OR `nickname` LIKE ? ESCAPE ? OR LOWER(`nickname`) LIKE ? ESCAPE ?", q)
 	require.Equal(t, []any{"\\%a8m\\%%", "\\", "%\\_alexsn\\_", "\\", "%\\\\pedro\\\\%", "\\", "%\\%abcd\\%efg%", "\\"}, args)
+
+	//Chinese
```
a8m, line 2071:
```suggestion
```

Excellent test, just remove the comment. It's not needed.

a8m (approved):
Thanks for fixing this, @cheny-alf 🚀 

I added 2 minor style comments.

dialect/sql/builder_test.go
```diff
@@ -2067,6 +2067,12 @@ func TestEscapePatterns(t *testing.T) {
...
+	query, args := Select("*").From(Table("dataset")).
+		Where(Contains("title", "_第一")).Query()
+	require.Equal(t, "SELECT * FROM `dataset` WHERE `title` LIKE ?", query)
+	require.Equal(t, []any{"%\\_第一%"}, args)
```
a8m, line 2075:
```suggestion
	q, args = Select("*").From(Table("dataset")).
		Where(Contains("title", "_第一")).Query()
	require.Equal(t, "SELECT * FROM `dataset` WHERE `title` LIKE ?", q)
	require.Equal(t, []any{"%\\_第一%"}, args)
```

Use the same variables from above to keep this function consistent. Thanks 🙏

cheny-alf:
Made changes according to your suggestions 🙏 @a8m

PR #3141, opened 2022-12-03
feat: soft delete

nzlov:
Soft delete support.
Comments also need to be added.
Just preview.
[demo](https://github.com/nzlov/entsoftdelete)


nzlov:
I will try to differentiate the schema...

Issue #2850, opened 2022-08-13
[HELP] Trying to implement soft delete logic using Hooks and Mixins


mirusky:
I'm trying to implement a soft delete operation, using `Hooks` and `Mixins`, as documentation sugests:

> A Mixin allows you to create reusable pieces of ent.Schema code that can be injected into other schemas using composition.
> The Hooks option allows adding custom logic before and after operations that mutate the graph.

Then I've created the following Mixin, with a Hook in Delete operations:

``` go
type DeletedAt struct {
	mixin.Schema
}

// Fields for all schemas that embed TenantMixin.
func (DeletedAt) Fields() []ent.Field {
	return []ent.Field{
		field.Time("deleted_at").Optional(),
	}
}

func (DeletedAt) Hooks() []ent.Hook {
	return []ent.Hook{
		hook.On(
			func(next ent.Mutator) ent.Mutator {
				type DeletedAtSetter interface {
					SetDeletedAt(time.Time)
				}

				return ent.MutateFunc(func(ctx context.Context, m ent.Mutation) (ent.Value, error) {
					if ns, ok := m.(DeletedAtSetter); ok {
						ns.SetDeletedAt(time.Now())
					}

					return next.Mutate(ctx, m)
				})
			}, ent.OpDeleteOne|ent.OpDelete,
		),
	}

```

But for some reason it always delete the entity instead of setting deletedAt column, I use it as follow:

``` go
// schame/user.go
func (User) Mixin() []ent.Mixin {
	return []ent.Mixin{
		CreatedAt{},
		UpdatedAt{},
		DeletedAt{},
	}
}

// main.go
func main() {
   // setup and other stuff ...

    client.User.DeleteOne(user).Exec(ctx)
}
```

What I'm missing ?

crossworth:
Hello, the way hooks works is like middlewares, you can hook into a flow of operation and change it. In your case, you are not changing the flow, you return `next.Mutate(ctx, m)`, so you are setting the field and right after that you execute the mutation (that deletes the record).

What you want to do is change the operation, but thats not that easy because the entity type can be anything, I don't think go has a concept of type lookup by name, so we have limited options, you could define this code on each entity definition (not using mixin) and then you could ensure that you know the type and change as you like, but in a generic way is kinda hard.

One simple solution would be to execute raw SQL commands to update the field, that would work, but you need to know the IDs and table name.

A better solution would be to use templates, I created a example for you:
https://github.com/crossworth/ent-soft-delete-hook

[...]

PR #3145, opened 2022-12-04
feat: schema soft delete

nzlov:
Soft delete support.
[demo](https://github.com/nzlov/ent/tree/softdelete/examples/softdelete)


giautm:
Many thanks for your contribution to this feature.

However, in my own opinion, Soft-Delete should be implemented at the Application Layer. Ent client provides templates, and extension hooks to support the implementation.

Can you change this PR to an example without touching Ent's core? We will support the example if you need it.

nzlov:
Thank you for your reply.
I think soft delete should be implemented in ORM again.
I know ent provides great extensions, but soft delete is a very common feature and should not be left to each developer to implement on their own.

If you can provide an example, I will close the merge request.
Thanks.

a8m:
Agree with Giau. Soft-delete can be implemented using a privacy policy or hook+interceptor (not released yet) on shared Mixin. I'll create an example PR in the upcoming hour, but here's an example:

```go

type SoftDeleteMixin struct {
	mixin.Schema
}

func (SoftDeleteMixin) Fields() []ent.Field {
	return []ent.Field{
		field.Time("deleted_at").
			Optional(),
	}
}

type softDeleteKey struct{}

// SkipSoftDelete returns a new context that skips the soft-delete interceptor/mutators.
func SkipSoftDelete(parent context.Context) context.Context {
	return context.WithValue(parent, softDeleteKey{}, true)
}

func (SoftDeleteMixin) Interceptors() []ent.Interceptor {
	return []ent.Interceptor{
		ent.TraverseFunc(func(ctx context.Context, q ent.Query) error {
			// Skip soft-delete, means include soft-deleted entities.
			if skip, _ := ctx.Value(softDeleteKey{}).(bool); skip {
				return nil
			}
			// Query non-deleted entities.
			if pq, ok := q.(*gen.PetQuery); ok {
				pq.Where(pet.DeletedAtIsNil())
			}
			return nil
		}),
	}
}

func (SoftDeleteMixin) Hooks() []ent.Hook {
	return []ent.Hook{
		hook.On(
				return hook.PetFunc(func(ctx context.Context, m *gen.PetMutation) (gen.Value, error) {
					// Skip soft-delete, means delete the entity permanently.
					if skip, _ := ctx.Value(softDeleteKey{}).(bool); skip {
						return next.Mutate(ctx, m)
					}
					ids, err := m.IDs(ctx)
					if err != nil {
						return nil, err
					}
					if err := m.Client().Pet.Update().Where(pet.IDIn(ids...)).SetDeletedAt(time.Now()).Exec(ctx); err != nil {
						return nil, fmt.Errorf("mark %d pets as deleted", len(ids))
					}
					return len(ids), nil
				})
			},
			ent.OpDeleteOne|ent.OpDelete,
		),
	}
}

```

nzlov:
> Agree with Giau. Soft-delete can be implemented using a privacy policy or hook+interceptor (not released yet) on shared Mixin. I'll create an example PR in the upcoming hour, but here's an example:
> 
> ```go
> type SoftDeleteMixin struct {
> 	mixin.Schema
> }
> 
> func (SoftDeleteMixin) Fields() []ent.Field {
> 	return []ent.Field{
> 		field.Time("deleted_at").
> 			Optional(),
> 	}
> }
> 
> type softDeleteKey struct{}
> 
> // SkipSoftDelete returns a new context that skips the soft-delete interceptor/mutators.
> func SkipSoftDelete(parent context.Context) context.Context {
> 	return context.WithValue(parent, softDeleteKey{}, true)
> }
> 
> func (SoftDeleteMixin) Interceptors() []ent.Interceptor {
> 	return []ent.Interceptor{
> 		ent.TraverseFunc(func(ctx context.Context, q ent.Query) error {
> 			// Skip soft-delete, means include soft-deleted entities.
> 			if skip, _ := ctx.Value(softDeleteKey{}).(bool); skip {
> 				return nil
> 			}
> 			// Query non-deleted entities.
> 			if pq, ok := q.(*gen.PetQuery); ok {                  <-------- repeat more
> 				pq.Where(pet.DeletedAtIsNil())
> 			}
> 			return nil
> 		}),
> 	}
> }
> 
> func (SoftDeleteMixin) Hooks() []ent.Hook {
> 	return []ent.Hook{
>               -------- repeat more -------- 
> 		hook.On(
> 			func(next ent.Mutator) ent.Mutator {
> 				return hook.PetFunc(func(ctx context.Context, m *gen.PetMutation) (gen.Value, error) {
> 					// Skip soft-delete, means delete the entity permanently.
> 					if skip, _ := ctx.Value(softDeleteKey{}).(bool); skip {
> 						return next.Mutate(ctx, m)
> 					}
> 					ids, err := m.IDs(ctx)
> 					if err != nil {
> 						return nil, err
> 					}
> 					if err := m.Client().Pet.Update().Where(pet.IDIn(ids...)).SetDeletedAt(time.Now()).Exec(ctx); err != nil {
> 						return nil, fmt.Errorf("mark %d pets as deleted", len(ids))
> 					}
> 					return len(ids), nil
> 				})
> 			},
> 			ent.OpDeleteOne|ent.OpDelete,
> 		),
> 	}
> }
> ```

Thanks.

The interceptor feature is great.
But this example is not simple.
1. if there is `softDeleteKey ` in the context, then all `edge` queries will skip the filter, and there is no precise control. For example: query soft-delete `user` for unsoft-delete `pet` data.
2. if there are 20 schemas, you need to write 20 query filters and 20 hooks.

a8m:
@nzlov, thanks for the comment. I’ll share a generic example once the code is ready 🙂

nzlov:
> @nzlov, thanks for the comment. I’ll share a generic example once the code is ready 🙂

Okay, thanks.

crossworth:
Hello guys 👋,

I have spent a few hours already trying to implement the soft delete feature, my last attempt was [described here](https://github.com/ent/ent/issues/2850#issuecomment-1214251964).

Now with Interceptors I think is possible to create a generic solution with the help of [EntQL](https://entgo.io/docs/feature-flags/#entql-filtering) and a single template file (and I think this template file can be converted to be a Ent extension to make things easy to work with or maybe even a feature flag).

I have made a few changes to the example @a8m provided:
https://github.com/crossworth/ent-soft-delete-hook/blob/soft-delete-with-interceptors/ent/schema/softdeletemixin.go

And this is the template file:
https://github.com/crossworth/ent-soft-delete-hook/blob/soft-delete-with-interceptors/ent/schema/softdelete.tmpl

Usage:
https://github.com/crossworth/ent-soft-delete-hook/blob/soft-delete-with-interceptors/bug_test.go

The template is only needed because we can not use Go generics in a few cases (at least I could not make it work) and to make it easy to use on the hooks/interceptor.

Of course this is only an example.

I think Interceptors open a new range of features that can be implemented with Ent.

--- 

> if there is softDeleteKey in the context, then all edge queries will skip the filter, and there is no precise control. For example: query soft-delete user for unsoft-delete pet data.

@nzlov I think the [Traverse functions](https://github.com/ent/ent/pull/3157/files#diff-fd90afb0d858d72b002f7ac2589e1e26122158d3b144f018dbad6b79d9e9a79dR399-R402) can be used to have a fine control, but I'm not sure if I understood the use case to be honest, your PR don't seems to have support for this feature as well.
This examples only care about the "first" query, maybe a easy solution would to invert the queries (pet -> user vs user -> pet), or maybe create a more "complete and complex" context, with more knobs.

a8m:
Amazing, @crossworth! I'll update my PR and will update this thread. I hope it will help

nzlov:
> Hello guys wave,
> 
> I have spent a few hours already trying to implement the soft delete feature, my last attempt was [described here](https://github.com/ent/ent/issues/2850#issuecomment-1214251964).
> 
> Now with Interceptors I think is possible to create a generic solution with the help of [EntQL](https://entgo.io/docs/feature-flags/#entql-filtering) and a single template file (and I think this template file can be converted to be a Ent extension to make things easy to work with or maybe even a feature flag).
> 
> I have made a few changes to the example @a8m provided: https://github.com/crossworth/ent-soft-delete-hook/blob/soft-delete-with-interceptors/ent/schema/softdeletemixin.go
> 
> And this is the template file: https://github.com/crossworth/ent-soft-delete-hook/blob/soft-delete-with-interceptors/ent/schema/softdelete.tmpl
> 
> Usage: https://github.com/crossworth/ent-soft-delete-hook/blob/soft-delete-with-interceptors/bug_test.go
> 
> The template is only needed because we can not use Go generics in a few cases (at least I could not make it work) and to make it easy to use on the hooks/interceptor.
> 
> Of course this is only an example.
> 
> I think Interceptors open a new range of features that can be implemented with Ent.
> 
> > if there is softDeleteKey in the context, then all edge queries will skip the filter, and there is no precise control. For example: query soft-delete user for unsoft-delete pet data.
> 
> @nzlov I think the [Traverse functions](https://github.com/ent/ent/pull/3157/files#diff-fd90afb0d858d72b002f7ac2589e1e26122158d3b144f018dbad6b79d9e9a79dR399-R402) can be used to have a fine control, but I'm not sure if I understood the use case to be honest, your PR don't seems to have support for this feature as well. This examples only care about the "first" query, maybe a easy solution would to invert the queries (pet -> user vs user -> pet), or maybe create a more "complete and complex" context, with more knobs.

@crossworth  Thank you
I have added the [edge query example](https://github.com/ent/ent/pull/3145/files#diff-927f574721d86e5cbfb77a3916ff0d2f09f6c621030f2aaf8de0ab4fd52d303aR85).
I forked your demo , added the same example [my repo](https://github.com/nzlov/ent-soft-delete-hook/blob/soft-delete-with-interceptors/softdelete.go), as I did and nothing works? Is it possible that the interceptor only works on the first level of queries?

crossworth:
Hello @nzlov, I think the Interceptors are draft yet, so maybe there is rough edges.
My first example only cared about the "first level" of query (or main query), but we can improve it to support the "edges" as well.

I have updated my example. I  moved the intercept code from the Mixin to the [template](https://github.com/crossworth/ent-soft-delete-hook/blob/26eaf82c7d3eddeb066331362bb5ddec50f6bf1b/ent/schema/softdelete.tmpl#L10-L24) (to be able to use private fields and check for the annotation), this simplify things greatly, we [only call the new function](https://github.com/crossworth/ent-soft-delete-hook/blob/26eaf82c7d3eddeb066331362bb5ddec50f6bf1b/ent/schema/softdeletemixin.go#L53) on the mixin.

We solve the problem by applying the filter and checking for the relations queries and using recursion to add the filter to them as well.
https://github.com/crossworth/ent-soft-delete-hook/blob/soft-delete-with-interceptors/ent/softdelete.go#L17-L38


I think there is possible problems with multi levels of `With<Type>`:
```go
client.Group.Query().WithUsers(func(query *ent.UserQuery) {
	query.WithGroups(func(query *ent.GroupQuery) {
		query.WithUsers()
	})
}).OnlyX(ctx)
```

Here the examples updated:
https://github.com/crossworth/ent-soft-delete-hook/blob/soft-delete-with-interceptors/bug_test.go

Anyway, this is just an example, it's a chance to play with the Interceptors and learn news ways to solve this problem (like I said, I already tried so many things while trying to solve the soft delete problem).

nzlov:
Hi @crossworth , nested queries can solve this query, but they shouldn't.
Ent should be made to support nesting.https://github.com/ent/ent/pull/3157#issuecomment-1343802853

This example still doesn't solve the query control problem. It would still be unfriendly to add multiple switch controls.
https://github.com/ent/ent/pull/3145/files#diff-927f574721d86e5cbfb77a3916ff0d2f09f6c621030f2aaf8de0ab4fd52d303aR183

nzlov:
Interceptors are powerful, but I don't see any simplicity in using interceptors to implement soft deletes.

PR #3152, opened 2022-12-07
ci: add Postgres 15 to integration

crossworth:


a8m (approved):
💯 

Thanks, @crossworth

Issue #833, opened 2020-10-09
Hooks for Read Operations


asonawalla:
Reading the docs and code, it looks like ent exposes the ability to register hooks in mutating operations, but not on reads.  The use case I'm trying to solve for is one in which sensitive data is encrypted before being persisted to a database and then decrypted on subsequent reads, assuming the caller provides the appropriate keys/metadata.  Given the lack of read hooks, the approaches to this problem today seem to be limited to (1) encrypting in a mutation hook and decrypting in the application or (2) moving all encryption / decryption to the application.  (1) is less than ideal because of divergent and arguably non-intuitive implementations and (2) opens up a greater possibility of accidentally circumventing the rules.

Is there another approach to this scenario I'm not seeing?  If not, would the ent authors consider either adding this or accepting a contribution that adds this?

a8m:
Hey @asonawalla and thanks for raising up this issue! I'll answer about it briefly and add more details about it tomorrow (it's a holiday evening atm here).

The only place for hooking into queries is the privacy layer (you won't find in the website, because docs is in progress). However, @alexsn and I just discussed about it yesterday, and agreed that we want to add support for read-hooks/interceptors. We thought about naming it as `Interceptor`s, but we're not sure about the end design. So, you're welcome to start the discussion here and propose  the API (implementation should be simple).  Once we add this, the privacy layer will be migrated to this generic layer as well.

asonawalla:
Perfect, I'll get a bit familiar with the privacy layer and circle back with any thoughts on API.  Enjoy your evening!

dilipkk-foyernet:
This is needed feature for us also, we want to inject some params to both read and write apis based user context. This is useful for security and customisation.

a8m:
Hey, the privacy doc is available now in [entgo.io/docs/privacy](https://entgo.io/docs/privacy/).

asonawalla:
I've been thinking through my use case and the ideal API a bit more.  For symmetric encryption, defining a `ReadHooks()` method in the schema which has access to field getters and setters will be able to convert persisted ciphertext into plaintext once the query has been run.  Unfortunately modifying fields in this way would break the ability to use that field in a query predicate, e.g. it wouldn't be straightforward to look up a user based on an API key that's hashed in the DB but passed plain in the application.

This is very likely out of scope for this issue, but maybe the ideal solution to this problem is to offer even more granular hook capabilities that are evaluated independently of the DB?  Something that might compile to roughly:

```
type Entity struct {
[...]

PR #3157, opened 2022-12-08
ent: export query interceptors

### LINKED ISSUE #3130: Hooks for query
According to the https://entgo.io/docs/hooks, there are hooks for mutate action, but there is no hooks for query, I want to add a opentelemetry tracer to auto-record the span of all actions to database like following 
``` golang
client.Use(func(next ent.Mutator) ent.Mutator {
	        opType := m.Op().String()
	        tablesName := m.Type()
	        spanName := fmt.Sprintf("%s-%s", tablesName, opType)
                // tracer is `trace.Tracer` of "go.opentelemetry.io/otel/trace"
	        newCtx, span := tracer.Start(ctx, spanName)
	        span.SetAttributes(
		        attribute.String("table", tablesName),
		        attribute.String("operation", opType),
	        )
	        defer span.End()
	        return next.Mutate(newCtx, m)
        })
})
```

- [ x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Motivation 🔦
It is hard to record each span manually in the service, so I want wrap the trace span in the hooks.

a8m:
See interceptors.mdx for more info.

Fixed #3130 and https://github.com/ent/ent/issues/833.


nzlov:
I see that you have not added an interceptor for the edge query

PR #3158, opened 2022-12-09
entc/gen: add driver as a reserved field word

tankbusta:
This fixes the code generator to make fields in an entity with the name `driver` to get generated as `_driver` to avoid a collision preventing compilation. 

### Steps to Reproduce

1. `go run -mod=mod entgo.io/ent/cmd/ent init Test`
2. Add a field called `driver`

```
// Fields of the Test.
func (Test) Fields() []ent.Field {
	return []ent.Field{
		field.String("driver"),
	}
}
```
3. `go generate`

Code wont compile because `<T>Mutation` expects to access `m.driver` which is originally contained from the embedded struct `<T>.config`

### DIFF
--- a/entc/gen/type.go
+++ b/entc/gen/type.go
@@ -2003,6 +2003,7 @@ var (
 		"predicates",
 		"typ",
 		"unique",
+		"driver",
 	)
 )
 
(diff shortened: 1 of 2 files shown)


a8m (approved):

a8m:
Thanks for the contribution, @tankbusta 🙏🏻

PR #3161, opened 2022-12-09
entc/gen: implement the ent.Mutator interface in the generated client

a8m:
cc @crossworth, @giautm


crossworth:
**This is great!**
I tried something similar to this in the past (changing the Op on the mutation), but could not make it work.

I tested with my softdelete tests and with this PR is possible to complete eliminate the template (at least for the hooks part).

short and sweet
```go
client.Use(
	hook.On(
				if schema.ShouldSKipDeletedTimeHook(ctx) {
					return next.Mutate(ctx, m)
				}
				sm, ok := m.(interface {
					SetOp(op ent.Op)
					SetDeletedTime(t time.Time)
					Client() *ent.Client
				})
				if !ok {
					return next.Mutate(ctx, m)
				}
				op := ent.OpUpdate
				if m.Op().Is(ent.OpDeleteOne) {
					op = ent.OpUpdateOne
				}
				sm.SetOp(op)
				sm.SetDeletedTime(time.Now())
				return sm.Client().Mutate(ctx, m)
			})
		},
		ent.OpDelete|ent.OpDeleteOne,
	),
)
```

PR #3162, opened 2022-12-09
dialect/sql/schema: fix issue with WriteDriver when using Postgres with Query

### LINKED ISSUE #3137: Generate migrations using Atlas
image

How can I achieve similar thing here with write driver using `Postgres` dialect

Using `Postgres` dialect here cause this error: query is not supported by the WriteDriver

Version: v0.11.4 using golang-migrate/migrate main.go

crossworth:
I think a return is missing, since the case for `INSERT/UPDATE`  is delegated to `Exec`, but I'm not 100% sure.
Please review if this makes sense.

I think this fixes #3137, I have tested and the migration is generated as expected.

https://github.com/ent/ent/blob/d7f2f3a098633ae8c6f844ce707a2eb983b1a340/dialect/sql/sqlgraph/graph.go#L1500-L1507


a8m:

PR #3164, opened 2022-12-10
Add schema.CommentAnnotation

### LINKED ISSUE #3155: Add new schema annotation to easily customize it's GoDoc comment

## Summary 💡

It would be nice to be able to trivially add a custom comment for a Schema in code. I think we could do something very similar to the `edge.Annotation` implementation.

In `schema/schema.go`(or maybe in a separate file in that package?) you could add a new type:
```go
// Better name?
type ModelAnnotation struct {
	Comment string
}

func (m *ModelAnnotation) Name() { return "Model" }
```

You could then simply add the following annotation to any schema:

```go
func (Foo) Annotations() []schema.Annotation {
	return []schema.Annotation{
		schema.ModelAnnotation{Comment: `Foo has a custom GoDoc comment.`},
	}
}
```

Then finally in `entc/gen/template/ent.tmpl` you could add something like:
```go
{{/* A template for setting the model comment. */}}
{{- define "model/comment" }}
	{{- if ne $.Annotations.Model.Comment "" }}
		{{- range $line := split $.Annotations.Model.Comment "\n" }}
			// {{ $line }}
		{{- end }}
	{{- else }}
		// {{ $.Name }} is the model entity for the {{ $.Name }} schema.
	{{- end }}
{{- end }}
```

## Motivation 🔦

It seems way easier than adding an additional template file for any Schema you want to add a comment to. It's also backwards compatible, as you could still leave the mechanism for addition additional comments via template.

Happy to implement if this makes sense!

adayNU:
Adds new mechanism for customizing a model's godoc comment in code (via an Annotation). Resolves #3155.

### DIFF
--- a/entc/gen/template/ent.tmpl
+++ b/entc/gen/template/ent.tmpl
@@ -19,7 +19,7 @@ import (
 	{{- end }}
 )
 
-// {{ $.Name }} is the model entity for the {{ $.Name }} schema.
+{{ template "model/comment" $ }}
 {{- with $tmpls := matchTemplate "model/comment/additional/*" }}
 	{{- range $tmpl := $tmpls }}
 		{{- xtemplate $tmpl $ }}
@@ -184,6 +184,17 @@ func ({{ $receiver }} {{ $slice }}) config(cfg config) {
 		return builder.String()
 	}
 {{ end }}
+
+{{/* A template for generating the Model's comment. */}}
+{{- define "model/comment" }}
+	{{- if $.Annotations.Model.Comment }}
+		{{- range $line := split $.Annotations.Model.Comment "\n" }}
+			// {{ $line }}
+		{{- end }}
+	{{- else }}
+		// {{ $.Name }} is the model entity for the {{ $.Name }} schema.
+	{{- end }}
+{{- end }}
 
 {{/* A template for generating the tag of the Edges struct-field. */}}
 {{- define "model/edgetags" }}
--- a/entc/gen/template/ent.tmpl
+++ b/entc/gen/template/ent.tmpl
@@ -185,10 +185,10 @@ func ({{ $receiver }} {{ $slice }}) config(cfg config) {
 	}
 {{ end }}
 
-{{/* A template for generating the Model's comment. */}}
+{{/* A template for generating the Entity's comment. */}}
 {{- define "model/comment" }}
-	{{- if $.Annotations.Model.Comment }}
-		{{- range $line := split $.Annotations.Model.Comment "\n" }}
+	{{- if and $.Annotations.Comment $.Annotations.Comment.Text }}
+		{{- range $line := split $.Annotations.Comment.Text "\n" }}
 			// {{ $line }}
 		{{- end }}
 	{{- else }}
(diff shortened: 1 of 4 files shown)


entc/gen/template/ent.tmpl
```diff
@@ -185,6 +185,17 @@ func ({{ $receiver }} {{ $slice }}) config(cfg config) {
 	}
 {{ end }}
 
+{{/* A template for generating the Model's comment. */}}
```
a8m, line 188:
```suggestion
{{/* A template for generating the Entity's comment. */}}
```

entc/gen/template/ent.tmpl
```diff
@@ -185,6 +185,17 @@ func ({{ $receiver }} {{ $slice }}) config(cfg config) {
...
+{{- define "model/comment" }}
+	{{- if $.Annotations.Model.Comment }}
+		{{- range $line := split $.Annotations.Model.Comment "\n" }}
+			// {{ $line }}
+		{{- end }}
+	{{- else }}
+		// {{ $.Name }} is the model entity for the {{ $.Name }} schema.
+	{{- end }}
```
a8m, line 196:
```suggestion
	{{- if and $.Annotations.Comment $.Annotations.Comment.Text }}
		{{- range $line := split $.Annotations.Comment.Text "\n" }}
			// {{ $line }}
		{{- end }}
	{{- else }}
		// {{ $.Name }} is the model entity for the {{ $.Name }} schema.
	{{- end }}
```

Name the annotation `Comment`. We can use it in other resources if necessary.

schema/schema.go
```diff
@@ -22,3 +22,23 @@ type Annotation interface {
 type Merger interface {
 	Merge(Annotation) Annotation
 }
+
+// CommentAnnotation is a builtin schema annotation for
+// configuring the schema's Godoc comment.
+type CommentAnnotation struct {
+	// Comment allows overriding the Godoc comment
+	// of the generated entity.
+	Comment string
```
a8m, line 31:
```suggestion
	Text string // comment text.
```

schema/schema.go
```diff
@@ -22,3 +22,23 @@ type Annotation interface {
...
+}
+
+// Name implements the Annotation interface.
+func (c *CommentAnnotation) Name() string {
+	return "Model"
```
a8m, line 36:
```suggestion
	return "Comment"
```

schema/schema.go
```diff
@@ -22,3 +22,23 @@ type Annotation interface {
...
+}
+
+// Comment returns a CommentAnnotation.
+func Comment(text string) *CommentAnnotation {
+	return &CommentAnnotation{Comment: text}
```
a8m, line 41:
```suggestion
	return &CommentAnnotation{Text: text}
```

schema/schema.go
```diff
@@ -22,3 +22,23 @@ type Annotation interface {
...
+	// of the generated entity.
+	Comment string
+}
+
+// Name implements the Annotation interface.
+func (c *CommentAnnotation) Name() string {
+	return "Model"
+}
+
+// Comment returns a CommentAnnotation.
```
a8m, line 39:
```suggestion
// Comment is a builtin schema annotation for
// configuring the schema's Godoc comment.
```

a8m (approved):
Thanks for the contribution, @adayNU. It looks great 😃 

I added a few style suggestions. Please, address them, and I'll merge this PR.

PR #3178, opened 2022-12-19
doc/md: update join community page

yonidavidson:
image

### DIFF
--- a/doc/md/community.md
+++ b/doc/md/community.md
@@ -1,5 +1,5 @@
 ---
-id: slack
+id: community
 title: Join our Community
 ---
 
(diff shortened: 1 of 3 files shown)


doc/md/community.md
```diff
@@ -0,0 +1,13 @@
+---
+id: slack
+title: Join our Community
```
rotemtam, line 3: this breaks the old URL please add an alias

rotemtam (changes requested):

doc/md/community.md
rotemtam: Sorry. I misread, you can give this a new URL and add the old one as an alias so it doesn't break 😊

yonidavidson: changed the name to community and added a redirect from slack to community

rotemtam (approved):

PR #3181, opened 2022-12-19
fix: Remove deprecated call in documentation

DanyHenriquez:
The documentation uses a deprecated function to setup a connection for an insecure gRPC call

Should be fixed with this PR


rotemtam (approved):
Thanks for the contribution @DanyHenriquez !
Also fixed in https://github.com/rotemtam/ent-grpc-example/pull/5

rotemtam (approved):

Issue #3193, opened 2022-12-27
Interceptor doesn't work for count query


taraktikos:
## Current Behavior 😯

After adding soft deletion to entity and fetching graph gql query with total I see the following queries:
```
2022-12-27T23:23:26.912+0200    INFO    dialect/dialect.go:79   driver.Query: query=SELECT COUNT(DISTINCT "users"."id") FROM "users" args=[]
2022-12-27T23:23:26.914+0200    INFO    dialect/dialect.go:79   driver.Query: query=SELECT DISTINCT "users"."id", "users"."created_at", "users"."updated_at", "users"."deleted_at", "users"."email", "users"."password", "users"."first_name", "users"."last_name", "users"."roles", "users"."status", "users"."phone", "users"."language", "users"."telegram_id", "users"."last_login_at" FROM "users" WHERE "users"."id" > $1 AND "users"."deleted_at" IS NULL ORDER BY "users"."id" ASC LIMIT 4 args=....
```
The first query without `"users"."deleted_at" IS NULL`
The second query with `"users"."deleted_at" IS NULL`

## Expected Behavior 🤔
Count query should be with intercepted predicate

a8m:
Closed with #3194.

Thanks for reporting this, @taraktikos.

PR #3194, opened 2022-12-27
entc/gen: copies interceptors on Clone

crossworth:
`entgql` uses [`Clone`](https://github.com/ent/contrib/blob/e2b0072613c094af4b0f47f6e62dd2ce88677051/entgql/template/pagination.tmpl#L530) when handling the pagination, the [Soft Delete example](https://entgo.io/docs/interceptors/#soft-delete) will not work in a few cases (like when using `first` or `last`), because the `Clone` does not copy the interceptors.

This PR fix this problem by copying the interceptors on Clone.

This fixes: #3193


a8m (approved):
💯 

Thanks, @crossworth.

Issue #3208, opened 2023-01-03
Running go generate nondeterministically changes generated code


jdhenke:
Running `go generate` nondeterministically changes the generated code that ent produces.

- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

```
# running generation the first time produces a diff
$ go generate -x ./services/index/internal/ent
go run -mod=mod entgo.io/ent/cmd/ent generate ./schema --feature sql/upsert
$ git diff -- services/index/internal/ent/ent.go
diff --git a/services/index/internal/ent/ent.go b/services/index/internal/ent/ent.go
index 6da39c7..e824fbd 100644
--- a/services/index/internal/ent/ent.go
+++ b/services/index/internal/ent/ent.go
@@ -11,6 +11,7 @@ import (
        "entgo.io/ent"
        "entgo.io/ent/dialect/sql"
        "entgo.io/ent/dialect/sql/sqlgraph"
+
        entfield "gitlab.com/adyton/platform/services/index/internal/ent/field"

# re-running generation produces a different output which this time happens to match what's checked in, so no diff
$ go generate -x ./services/index/internal/ent  
go run -mod=mod entgo.io/ent/cmd/ent generate ./schema --feature sql/upsert
$ git diff -- services/index/internal/ent/ent.go
$ grep entgo.io go.mod
        entgo.io/ent v0.11.5-0.20230102205927-542f36d4abda

```

## Expected Behavior 🤔

Code generation should be deterministic and not result in any changes.

## Steps to Reproduce 🕹

See above.

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | go version go1.19.4 darwin/amd64  |
| Ent         | entgo.io/ent v0.11.5-0.20230102205927-542f36d4abda   |
| Database    | PostgreSQL   |
| Driver      | https://github.com/lib/pq       |

a8m:
Hey @jdhenke 👋 
As discussed on Discord, can you please provide an example that reproduces it using https://github.com/ent/bug? Thanks 🙏

jdhenke:
@a8m reproduced here: https://github.com/jdhenke/ent-3208/pull/1

I'm guessing it has to do with creating an entity called `Field` which forces a named import which hits an edge case in the code generation that makes the output non-deterministic.

```
[...]

PR #3209, opened 2023-01-04
entc/gen: explicitly import node packages in generate ent.go file

a8m:
Fixed https://github.com/ent/ent/issues/3208

Verified this with the following command:

```zsh
repeat 50 { go generate ./... && git --no-pager diff }
```


entc/integration/ent/ent.go
```diff
@@ -28,6 +28,7 @@ import (
 	"entgo.io/ent/entc/integration/ent/node"
 	"entgo.io/ent/entc/integration/ent/pet"
 	"entgo.io/ent/entc/integration/ent/spec"
+
```
giautm, line 31: @a8m There is a new blank line. Is it correct?

PR #3211, opened 2023-01-06
Do not increment idx for interceptors when a policy is present

ryanahall:
**Issue**
I have a mixin that includes both a policy and an interceptor. After running `go generate ./...`, `runtime.go` is invalid and includes a number of errors, namely `invalid argument: index 1 out of bounds [0:1]`. 

As discussed on [discord](https://discord.com/channels/885059418646003782/885059419107360779/1060990994243391528), this change fixes the above issue and ensures that interceptors are indexed starting at 0.


a8m (approved):
Thanks for the contribution, @ryanahall. Looks good 🚀

Issue #1865, opened 2021-08-23
Generated package names conflict


maxifom:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

I have a package called my-proj/internal/message, and I have a message type in schema.
Generated ent package is my-proj/ent/message.
If I want to use string enum from my-proj/internal/message in generated code there is naming conflict, and it's bad to change generated code with hand as we know.

## Expected Behavior 🤔

No naming conflict. Maybe we can detect this cases and alias imports?

## Steps to Reproduce 🕹

Repo: https://github.com/maxifom/ent-naming-conflict

Steps:

1. Create package <PROJECT>/internal/message
2. Create Message type in schema.go
3. go generate
4. There should be naming conflict in message_create.go, message_update message_query message_delete

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.16.6  |
| Ent         | v0.9.1-0.20210814144404-b8532f87a648  |

phamvinhdat:
hi @a8m, I we just need an import tracker something like that  https://github.com/kubernetes/gengo/blob/master/namer/import_tracker.go

PR #3236, opened 2023-01-12
support import alias

phamvinhdat:
a demo to fix issues: https://github.com/ent/ent/issues/1865


phamvinhdat:
Hi @a8m, can you take a look to see if we can go this way to avoid duplicate import?

PR #3243, opened 2023-01-14
entc: Return errors from both packages

tmc:
This returns errors from both packages after Load() -> issues loading that can crop up by having a `go.work` file present are surfaced more clearly with this in place which improves user experience.


a8m (approved):
Nice! Thanks for fixing this.

Issue #3261, opened 2023-01-19
Support Subqueries in Selector.From()


agratta-silo:
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

Given a custom function `calculate_score`, I'd like to run a query of the form
```
SELECT * 
FROM (SELECT *, calculate_score(users.name, 'searchString') as score FROM users)
WHERE score > 1
ORDER BY score
```

I'm imagining we'd write this in ent as follows
```
s.Select("*").From(
	sql.SelectExpr(sql.Raw(fmt.Sprintf("*, calculate_score(users.name, '%s') AS score", searchString))).
	From(sql.Table("users"))).
Where(sql.ExprP("score > 1")).
OrderBy("score desc")
```

The existing Selector.From method has the signature `func (s *Selector) From(t TableView) *Selector`
The `TableView` parameter can be a `Selector`, `TableSelector`, or `View`.
However, if you attempt to pass in a `Selector` here, the query will panic as the builder attempts to cast it to a `TableSelector`.

## Motivation 🔦
Using a custom SQL function demands that we use a subquery or CTE (at least in Postgres)
I can't currently do this with ent, so I need to write the entire query as raw SQL.
Ideally I could abstract this logic into a custom predicate and do something like
`ent.Users.Query().Where(user.NameFuzzilyMatches(searchString)).All(ctx)`

crossworth:
Hi @agratta-silo , if you don't need the score, you can handle it quit easily, like the example below:

```go
ScoreGT := func(searchString string, val int) func(s *sql.Selector) {
	return func(s *sql.Selector) {
		f := sql.P()
		f.WriteString("calculate_score")
		f.Wrap(func(b *sql.Builder) {
			b.WriteString(s.C(user.FieldName))
			b.Comma()
			b.WriteString("'" + searchString + "'")
		})
		s.Where(sql.GT(f.String(), val))
	}
}
users := client.Debug().User.Query().Where(ScoreGT("input", 1)).AllX(ctx)
```

See that I avoid the subquery, but lost the result of the function, if you need the result you can use the [Custom SQL Modifiers feature flag](https://entgo.io/docs/feature-flags/#custom-sql-modifiers).

In that case you can use something like this:

```go
searchString := "input"
val := 1
var result []struct {
	ent.User
	Score int
}
err = client.Debug().User.Query().Modify(func(s *sql.Selector) {
	f := sql.P()
[...]

PR #3274, opened 2023-01-25
dialect/sql: improve support for subqueries

crossworth:
This PR improves the way we handle returning the `Table` for the `Selector`.

---

When handling complex queries, `Table()` will be called with a `*Selector`, causing a panic, this PR tries to solve this problem by checking the type and doing a recursive search for the first `SelectTable` found.

I'm not sure about the part were we use the first element, not sure if this is the correct behavior, but the test pass and things works.

We can improve code from:
```go
err = client.Debug().User.Query().Modify(func(s *sql.Selector) {
	f := sql.P()
	f.WriteString("calculate_score")
	f.Wrap(func(b *sql.Builder) {
		b.WriteString(s.C(user.FieldName)).Comma().WriteString("'" + searchString + "'") // <--- We improve this part
	})
	u := sql.Table(user.Table)
	subquery := sql.Select("*").AppendSelectExprAs(sql.Raw(f.String()), "score").From(u).As("user_with_score")
	s.Select("*").From(subquery)
	s.Where(sql.GT(subquery.C("score"), val))
	s.OrderBy(sql.Asc(subquery.C("score")))
}).Scan(ctx, &result)
```
To:
```go
err = client.Debug().User.Query().Modify(func(s *sql.Selector) {
	f := sql.P(func(b *sql.Builder) {
		b.SetDialect(s.Dialect())
		b.WriteString("calculate_score")
		b.Wrap(func(bb *sql.Builder) {
			bb.WriteString(s.Table().C(user.FieldName)).Comma().Args(searchString)
		})
	})
	u := sql.Table(user.Table)
	subquery := sql.Select("*").AppendSelectExprAs(f, "score").From(u).As("user_with_score")
	s.Select("*").From(subquery)
	s.Where(sql.GT(subquery.C("score"), val))
	s.OrderBy(sql.Asc(subquery.C("score")))
}).Scan(ctx, &result)
```

See https://github.com/ent/ent/issues/3261#issuecomment-1404058116 for more context.

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -2367,7 +2367,25 @@ func (s *Selector) Table() *SelectTable {
 	if len(s.from) == 0 {
 		return nil
 	}
-	return s.from[0].(*SelectTable)
+	return selectTable(s.from[0])
+}
+
+// selectTable returns a *SelectTable from the given TableView.
+func selectTable(tb TableView) *SelectTable {
+	if tb == nil {
+		return nil
+	}
+	switch view := tb.(type) {
+	case *SelectTable:
+		return view
+	case *Selector:
+		if len(view.from) == 0 {
+			return nil
+		}
+		return selectTable(view.from[0])
+	default:
+		panic(fmt.Sprintf("unhandled TableView type %T", tb))
+	}
 }
 
 // TableName returns the name of the selected table or alias of selector.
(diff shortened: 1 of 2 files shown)


a8m (approved):
🚀

a8m:

Issue #3255, opened 2023-01-18
Invalid Code Generation using a JSON Field with a ValueScanner type


agratta-silo:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

The following schema will fail to compile after code generation because it has a JSON field using the `MyJson` type, which implements the ValueScanner interface:
```
type Todo struct {
	ent.Schema
}

func (Todo) Fields() []ent.Field {
	return []ent.Field{
		field.JSON("myJson", MyJson{}).Optional(),
	}
}

type MyJson struct {
	Name string `json:"name"`
}

func (p *MyJson) Scan(v interface{}) error {
	if b, ok := v.([]byte); !ok {
		return fmt.Errorf("could not convert %T to []byte", v)
	} else if err := json.Unmarshal(b, p); err != nil {
		return err
	}
	return nil
}

func (p MyJson) Value() (driver.Value, error) {
	b, err := json.Marshal(&p)
	if err != nil {
		return nil, err
	}
	return driver.Value(b), nil
}
```

The problematic code is in the generated `todo.go`:
```
func (t *Todo) assignValues(columns []string, values []any) error {
	if m, n := len(values), len(columns); m < n {
		return fmt.Errorf("mismatch number of scan values: %d != %d", m, n)
	}
	for i := range columns {
		switch columns[i] {
		case todo.FieldID:
			value, ok := values[i].(*sql.NullInt64)
			if !ok {
				return fmt.Errorf("unexpected type %T for field id", value)
			}
			t.ID = uint(value.Int64)
		case todo.FieldMyJson:
			if value, ok := values[i].(*schema.MyJson); !ok {
				return fmt.Errorf("unexpected type %T for field myJson", values[i])
			} else if value != nil && len(*value) > 0 { // <---------------------------- These two lines are treating value as []byte
				if err := json.Unmarshal(*value, &t.MyJson); err != nil { // <-------- but we've cast the data to *schema.MyJson
					return fmt.Errorf("unmarshal field myJson: %w", err)
				}
			}
		}
	}
	return nil
}
```

If I remove the ValueScanner implementation, the value is cast to `[]byte` instead, and the code compiles.

## Expected Behavior 🤔

Compile whether or not the ValueScanner interface is implemented. In my use case, the ValueScanner is implemented for use in another tool (not for ent), so either ent should ignore the implementation for json fields or support it fully.

[...]

PR #3275, opened 2023-01-25
entc/gen: improve support for JSON types

crossworth:
This PR adds support for using `JSON` fields with types that implement the `ValueScanner` and are not slices.

This fixes: #3255

### DIFF
--- a/entc/gen/template/dialect/sql/decode.tmpl
+++ b/entc/gen/template/dialect/sql/decode.tmpl
@@ -100,16 +100,22 @@ func ({{ $receiver }} *{{ $.Name }}) assignValues(columns []string, values []any
 	{{- $f := $.Scope.Field -}}
 	{{- $ret := $.Scope.Rec -}}
 	{{- $field := $f.StructField }}{{ with $.Scope.StructField }}{{ $field = . }}{{ end -}}
+	{{- $scantype := $f.ScanType -}}
 	{{- if $f.IsJSON -}}
-		if value, ok := values[{{ $i }}].(*{{ $f.ScanType }}); !ok {
+		if value, ok := values[{{ $i }}].(*{{ $scantype }}); !ok {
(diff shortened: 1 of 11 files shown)


a8m (commented):

JSON types are marshaled/unmarshaled by Ent. Thus, the ValueScanner is skipped and the result is `json.RawMessage`/`[]byte`. What's the reason to support ValueScanner in this case?

a8m:
Reading your comment in #3255, this sounds like a good alternative:
> Hi @agratta-silo, I have [create an PR to solve this issue](https://github.com/ent/ent/pull/3275), I would like to also note that it's possible to implement the desired behavior with the `Other` type as well.
> 
> ```go
> type Todo struct {
> 	ent.Schema
> }
> 
> func (Todo) Fields() []ent.Field {
> 	return []ent.Field{
> 		field.Other("my_json", types.MyJson{}).
> 			Optional().
> 			SchemaType(map[string]string{
> 				dialect.SQLite:   "json",
> 				dialect.MySQL:    "json",
> 				dialect.Postgres: "jsonb",
> 			}),
> 	}
> }
> ```

crossworth:
>  What's the reason to support ValueScanner in this case?

While I reviewed the issue I thought that the problem seems like strange, since I remembered using types with `ValueScanner` before, but after checking I always used `Other`, so I don't really have a reason on my side, but I think it would make sense, since it's a relative small change.

Now, I think we have to address the problem, either we allow it or we fail during the generating step, I don't think makes sense to generate invalid code.

> JSON types are marshaled/unmarshaled by Ent. Thus, the ValueScanner is skipped and the result is `json.RawMessage/[]byte`.

 Are you sure? I tested and when the type implement `Scan`, the type is used (and thats generates the invalid code), `Value` don't seems to be used.

So maybe a better solution would be to ignore the fact that the type implement `ValueScanner` when using `JSON` field and always use `json.RawMessage/[]byte`?

Let me know what you think.

PR #3305, opened 2023-02-06
entc/gen: reduce the usage of DISTINCT in queries

a8m:
Most queries are not graph traversals but rather regular table scans, in which case the DISTINCT clause is not needed as duplicates cannot be returned (unless query was modified by the user).


a8m:
cc @crossworth, @giautm

PR #3306, opened 2023-02-07
entc/gen: support type-based matching for edge-schema columns

a8m:
In case of no match by foreign-keys, we search by edge-type. This can happen if the type (edge owner) is named `T`, but the edge-schema `E` names its edge field as `u_id`. We consider it as a match if there is only one usage of `T` in `E`.


NickDubelman:
Hmm just want to make sure but this looks like it was a response to the repro I created and posted in Discord ~2 weeks ago or so?

https://github.com/NickDubelman/ent/tree/edge-schema-repro

I'm going to try and pull master and see if it resolves my issue!

NickDubelman:
Seems to fix it! Thanks a lot for the quick fix, your solution is better than what I expected because it doesn't require any explicit `StorageKey`

PR #3308, opened 2023-02-07
doc: add an instruction to create a migrations directory

noamcattan:
add code that checks if the migration directory exists, and creates it otherwise

### DIFF
--- a/doc/md/versioned-migrations.mdx
+++ b/doc/md/versioned-migrations.mdx
@@ -118,7 +118,7 @@ docker run --name migration --rm -p 5432:5432 -e POSTGRES_PASSWORD=pass -e POSTG
 </TabItem>
 </Tabs>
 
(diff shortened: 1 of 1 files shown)


doc/md/versioned-migrations.mdx
```diff
@@ -152,6 +152,12 @@ import (
 
 func main() {
 	ctx := context.Background()
+	
+	// Check if the migration directory exists and create it otherwise
+	if _, err := os.Stat(migrationDirPath); os.IsNotExist(err) {
+		os.Mkdir(migrationDirPath, 0755)
+	}
```
a8m, line 159: `os.Mkdir` requires another error handling. Better to tell the user to create it manually in the first step of the tutorial rather than having it in the script.

noamcattan: changed

a8m (approved):

PR #3314, opened 2023-02-13
dialect/sql/schema: use the Atlas as state reader

a8m:
Calling `StateReader.ReadState` should set up an Ent driver instead of panicking.


masseelch (commented):
Shouldn't `NewMigrate` do this?

a8m:
> Shouldn't `NewMigrate` do this?

We have both `NewMigrate` and `NewMigrateURL` and two different ways for initialization. I prefer keep the logic like the other functions because the flow is more consistent this way.

masseelch (approved):

PR #3336, opened 2023-02-20
ent: Attach Funcs for `ent new` templates

tmc:
This enables the use of the [`gen.Funcs`](https://github.com/ent/ent/blob/master/entc/gen/func.go#L28) helper FuncMap in `ent new`.


a8m (approved):
Thanks, @tmc ❤️

PR #3359, opened 2023-03-01
schema/field: fallback to native validator if not configured one

a8m:


masseelch:
You need to run go generate

PR #3360, opened 2023-03-01
withHooks: redundant any() conversion

Antonboom:
Inspired by https://github.com/golang/go/issues/58633.

Added `asciicheck` to linters list like SSA-assertion.

### DIFF
--- a/entc/gen/template/base.tmpl
+++ b/entc/gen/template/base.tmpl
@@ -307,7 +307,8 @@ func withHooks[V Value, M any, PM interface {
 		return exec(ctx)
 	}
 	var mut Mutator = MutateFunc(func(ctx context.Context, m Mutation) (Value, error) {
-		mutationT, ok := m.(PM)
+		{{- /* redundant any() conversion is a workaround for golang/go#58633 */}}
+		mutationT, ok := any(m).(PM)
 		if !ok {
 			return nil, fmt.Errorf("unexpected mutation type %T", m)
 		}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Run linters
         uses: golangci/golangci-lint-action@v3.4.0
         with:
-          version: v1.48.0
+          version: v1.50.1 # The next version requires Go 1.20
 
   unit:
     runs-on: ubuntu-latest
@@ -76,7 +76,7 @@ jobs:
         run: |
           status=$(git status --porcelain)
           if [ -n "$status" ]; then
-            echo "you need to run 'go generate ./...' and commit the changes"
+            echo "you need to run 'go generate ./...' in root and 'examples' dirs and commit the changes"
             echo "$status"
             git --no-pager diff
             exit 1
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,6 +18,7 @@ linters-settings:
 linters:
   disable-all: true
   enable:
+    - asciicheck
     - bodyclose
     - depguard
     - dogsled
--- a/entc/integration/cascadelete/ent/ent.go
+++ b/entc/integration/cascadelete/ent/ent.go
@@ -513,7 +513,7 @@ func withHooks[V Value, M any, PM interface {
 		return exec(ctx)
 	}
 	var mut Mutator = MutateFunc(func(ctx context.Context, m Mutation) (Value, error) {
-		mutationT, ok := m.(PM)
+		mutationT, ok := any(m).(PM)
 		if !ok {
 			return nil, fmt.Errorf("unexpected mutation type %T", m)
 		}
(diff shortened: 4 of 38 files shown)


amaneshi:
Also need to mention that this fix require custom linter build

> Ah, you also need to have run "go get golang.org/x/tools@master" in the golangci-lint module and rebuilt the tool, since that picks up a fix that allows the workaround to work. (I had done that step, which was not sufficient by itself, and then forgotten that it enabled the workaround.)

And this is temp workaround, maybe better to wait some mature solution...

Antonboom:
> this fix require custom linter build

Yes, the next step is MR to golangci-lint. 

> maybe better to wait some mature solution

As u wish. For example, do not use so complex generic code :) 

And replace
```go
*mutation = *mutationT
```
with 
```go
mutation.Set(mutationT) // new interface method.
```

But at the moment **ent** is the main blocker to upgrade linter in our projects to 1.20 support.

entc/gen/template/base.tmpl
```diff
@@ -307,7 +307,7 @@ func withHooks[V Value, M any, PM interface {
 		return exec(ctx)
 	}
 	var mut Mutator = MutateFunc(func(ctx context.Context, m Mutation) (Value, error) {
-		mutationT, ok := m.(PM)
+		mutationT, ok := any(m).(PM) // redundant any() conversion is a workaround for golang/go#58633
```
a8m, line 310: Change it to a template comment, please?
```suggestion
		{{- /* redundant any() conversion is a workaround for golang/go#58633 */}}
		mutationT, ok := any(m).(PM)
```

a8m:
Thanks for the contribution, @Antonboom.

entc/gen/template/base.tmpl
Antonboom: Great! Completely forgot about them :)
Also did not like that the result gets into the generated code.
Fixed

a8m:
@Antonboom, can you elaborate on the discussion in https://github.com/golang/go/issues/58633#issuecomment-1458561802? Does it solve the problem atm? Is there any estimation to support this?

Antonboom:
@a8m 

I think these are two different problems union in 1 issue, maybe better to create a separate one.


This panic was fixed by workaround in `ent`:

```
cannot convert *t0 (M) to PM
```


But this panic is known problem and sometimes appears [before](https://github.com/golangci/golangci-lint/issues/3086):

```
interface conversion: interface {} is nil, not *buildssa.SSA
```
I suppose a fixes will be in **x/tools/go** or **golangci-lint** or specific linters.
And there's nothing we can do from `ent` side.

---

And how to fix [CI / generate](https://github.com/ent/ent/actions/runs/4363426781/jobs/7649769357)?

```                                                                                                                                                              
$ go generate ./...
$ git status       
On branch fixes/ssa
nothing to commit, working tree clean
```

a8m:
@Antonboom, run `go generate ./...` from the `examples` as well. Thanks for the contribution 🙏

Antonboom:
@a8m 👍

a8m (approved):
Thanks for the contribution, @Antonboom. Let's just remember to revert this patch when the issue is resolved in x/tools.

sagikazarmark:
Any chance this could be released as a patch? 🙏

a8m:
Done. v0.11.10 was released. 

Note everyone, v1 is going to be released in the upcoming weeks 😎

antoniopironti:
Hi @Antonboom, thank you very much for this PR. 

I upgraded ent to the version just released, ran `ent generate` and `golangci-lint run` is still panicing for me (I tried both `1.51.1` and `1.51.2`). Is there any additional step I am missing to workaround this problem? Thank you!

Antonboom:
Hi @antoniopironti

Yep, it should be fixed with the next `golangci-lint` version. I want to propose to make the new release in one-two weeks.

antoniopironti:
That would be awesome. Thank you very much again to be on top of this!

mgabeler-lee-6rs:
Any tips on how to get `govulncheck` to not panic with this error, even after taking in the ent release with this fix?

Antonboom:
@mgabeler-lee-6rs it seems you need to bump [x/tools in govulncheck](https://github.com/golang/vuln/blob/8a73a7e4f20339b847bea345d670eb6c238e800f/go.mod#L11).

antoniopironti:
Hello folks! A new version of golang-ci has [been released yesterday](https://github.com/golangci/golangci-lint/releases/tag/v1.52.0), which unfortunately still seems to be affected by this bug. My environment:
* go `v1.20.2`
* golangci-lint `v1.52.0`
* ent `v0.11.10`

Antonboom:
@antoniopironti hello!

I surprised too, my local build of `golangci-lint` passed tests on the moment of this fix 🤯 

Workaround is to disable the next [linters](https://github.com/golangci/golangci-lint/issues/3086#issuecomment-1475232706).
And please, move to issue above, stop comment in this pull request. Thanks!

PR #3367, opened 2023-03-06
dialect/sql/schema: strip returning from seed

masseelch:


dialect/sql/schema/writer.go
```diff
@@ -229,6 +231,60 @@ func (w *WriteDriver) formatArg(v any) (string, error) {
...
+				switch query[j] {
+				case '\\':
+					j++
+				case query[index]:
+					return query[index : j+1], j
+				}
+			}
+			// Unexpected EOS.
+			return query, -1
+		}
```
a8m, line 251:
```suggestion
		skipQuoted = func(query []byte, idx int) ([]byte, int) {
			for j := idx + 1; j < len(query); j++ {
				switch query[j] {
				case '\\':
					j++
				case query[idx]:
					return query[idx : j+1], j
				}
			}
			// Unexpected EOS.
			return query, -1
		}
```

dialect/sql/schema/writer.go
```diff
@@ -229,6 +231,60 @@ func (w *WriteDriver) formatArg(v any) (string, error) {
...
+		case '\'', '"', '`': // string or identifier
+			s, skip := skipQuoted(query, i)
+			if skip == -1 {
+				return query
+			}
+			b.Write(s)
+			i = skip
+			continue
+		case ' ', 'r', 'R': // start of RETURNING?
+			if bytes.HasPrefix(bytes.ToLower(bytes.TrimSpace(query[i:])), returning) {
```
a8m, line 265:
```suggestion
		switch r := query[i]; {
		case r == '\'', r == '"', r == '`': // string or identifier
			s, skip := skipQuoted(query, i)
			if skip == -1 {
				return query
			}
			b.Write(s)
			i = skip
			continue
		case bytes.HasPrefix(query[i:], returingL), bytes.HasPrefix(query[i:], returningU): // start of RETURNING?
```
More efficient and cleaner, IMO. Just call TrimRight(query, " ") at the end.

a8m (approved):
Looks good, but added two small comments.

PR #3386, opened 2023-03-13
dialect/sql: Add INNER JOIN to Selector

C-Deck:
This adds the ability to do inner joins.


a8m (commented):
Thanks for the contribution, @C-Deck.

I agree with @masseelch that there's no need for `INNER JOIN` or `CROSS JOIN` because they are equivalents (I actually find it confusing for beginners).

However, I don't mind leaving this PR open and hearing what others think about it.

C-Deck:
I'll go ahead and close it. No worries.

Issue #3064, opened 2022-11-03
quality of life: handle GoTypes() that implement encoding.Unmarshal{Binary,Text} & encoding.Marshal{Binary,Text}


devnode:
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

```go
// continue checking if GoType(V{}) implements this
type ValueScanner interface {
	driver.Valuer
	sql.Scanner
}

// but if not, and if field of type {Bytes,Other,String,Text}, then check if V implements
type TextEncoder interface {
	encoding.TextMarshaler
	encoding.TextUnmarshaler
}

// if it does, then just regurgitate something like this:
type TextValueScanner[V TextEncoder] struct {
	value V
}

func (s *TextValueScanner[V]) Scan(in interface{}) error {
	switch v := in.(type) {
	case string:
		return s.value.UnmarshalText([]byte(v))

	case []byte:
		return s.value.UnmarshalText(v)

	default:
		return fmt.Errorf("invalid type: %T", in)
	}
}

func (s *TextValueScanner[V]) Value() (driver.Value, error) {
	return s.value.MarshalText()
}

// or whatever you want to call it
func (s *TextValueScanner[V]) ToFieldValue() V {
	return s.value
}

// then plumb it up to the generated struct
```

- https://pkg.go.dev/encoding#pkg-types
- https://pkg.go.dev/net/netip
- https://pkg.go.dev/math/big#Float (and others...)
- https://pkg.go.dev/cloud.google.com/go/civil

## Motivation 🔦

- to reduce the amount of code consumers of `ent` have to write
- and to just avoid the gymnastics of having to type wrap fairly common types

a8m:
Hey @devnode, and thanks for proposing this issue. I was thinking about adding another option for setting the Value/Scan functions (handlers) as an additional option. This way, it will be possible to override it for all types. The `Binary`/`Text` helpers can be added to the Ent stdlib to simplify the usage. e.g.:

```go
field.String("t").
    ValueScanner(Tokenizer{})

field.String("url").
    GoType(&url.URL{}).
    ValueScanner(sql.BinaryCodec[*url.URL]{})
```

Thoughts?

PR #3391, opened 2023-03-15
schema/field: add support or external ValueScanner

a8m:
This PR allows attaching custom `ValueScanner` for basic or custom Go types. This enables the use of standard
schema fields while maintaining control over how they are stored in the database without implementing a `ValueScanner`
interface. Additionally, this option enables users to use `GoType` that does not implement the `ValueScanner`, such
as `*url.URL`.

Note: At this stage, this option is only available for text and numeric fields, but it will be extended to other types in the future.

### TextMarshaller

Fields with a custom Go type that implements the `encoding.TextMarshaller` and `encoding.TextUnmarshaller` interfaces can use the `field.TextValueScanner` as a `ValueScanner`. This `ValueScanner` calls `MarshalText` and `UnmarshalText` for writing and reading field values from the database:

```go
field.String("big_int").
    GoType(&big.Int{}).
    ValueScanner(field.TextValueScanner[*big.Int]{})
```

### BinaryMarshaller

Fields with a custom Go type that implements the `encoding.BinaryMarshaller` and `encoding.BinaryUnmarshaller` interfaces can use the `field.BinaryValueScanner` as a `ValueScanner`. This `ValueScanner` calls `MarshalBinary` and `UnmarshalBinary` for writing and reading field values from the database:

```go
field.String("url").
    GoType(&url.URL{}).
    ValueScanner(field.BinaryValueScanner[*url.URL]{})
```

### Functions based"

The `field.ValueScannerFunc` allows setting two functions to be used for writing and reading database values: `V` for `driver.Value` and `S` for `sql.Scanner`:

```go
field.String("encoded").
	ValueScanner(field.ValueScannerFunc[string, *sql.NullString]{
		V: func(s string) (driver.Value, error) {
			return base64.StdEncoding.EncodeToString([]byte(s)), nil
		},
		S: func(ns *sql.NullString) (string, error) {
			if !ns.Valid {
				return "", nil
			}
			b, err := base64.StdEncoding.DecodeString(ns.String)
			if err != nil {
				return "", err
			}
			return string(b), nil
		},
	})
```

### Custom

```go
field.String("prefixed").
	ValueScanner(PrefixedHex{
		prefix: "0x",
	})
```

```go 

// PrefixedHex is a custom type that implements the TypeValueScanner interface.
type PrefixedHex struct {
	prefix string
}

// Value implements the TypeValueScanner.Value method.
func (p PrefixedHex) Value(s string) (driver.Value, error) {
	return p.prefix + ":" + hex.EncodeToString([]byte(s)), nil
}

// ScanValue implements the TypeValueScanner.ScanValue method.
func (PrefixedHex) ScanValue() field.ValueScanner {
	return &sql.NullString{}
}

// FromValue implements the TypeValueScanner.FromValue method.
func (p PrefixedHex) FromValue(v driver.Value) (string, error) {
	s, ok := v.(*sql.NullString)
	if !ok {
		return "", fmt.Errorf("unexpected input for FromValue: %T", v)
	}
	if !s.Valid {
		return "", nil
	}
	d, err := hex.DecodeString(strings.TrimPrefix(s.String, p.prefix+":"))
	if err != nil {
		return "", err
	}
	return string(d), nil
}
```

Fixed https://github.com/ent/ent/issues/3064


a8m:
> Awesome! Though, why the short field names V and S? Value and Scan or Valuer and Scanner or ValuerFunc and ScanerFunc?

For some reason, I found it cleaner. Because both the Value/Scanner are used in the definition. 
```
ValueScanner(field.ValueScannerFunc{...})
```

Thanks for the review, @masseelch!

PR #3401, opened 2023-03-20
entc/gen: ensure edge-fields are selected when their edges are loaded

a8m:


dialect/sql/sqlgraph/graph.go
```diff
@@ -337,6 +337,25 @@ func NewFieldSpec(column string, typ field.Type) *FieldSpec {
 	return &FieldSpec{Column: column, Type: typ}
 }
 
+// AddColumnOnce adds the given column to the spec if it is not already present.
+func (n *NodeSpec) AddColumnOnce(column string) *NodeSpec {
+	for _, c := range n.Columns {
+		if c == column {
+			return n
+		}
+	}
```
masseelch, line 346: Not majorly important but I think the majority of cases you add a column twice right after another, right? Would be faster to loop in reverse order. 

Also, did you consider a map for O(1)?

a8m: > Not majorly important but I think the majority of cases you add a column twice right after another, right?

There's no specific behavior in the generated code. I use this function to ensure edge fields are selected when edges are loaded in case of a custom user selection. The order can be changed. 

> Also, did you consider a map for O(1)?

I considered it, but I preferred to avoid generating garbage for small iterations (we had such a discussion in the issue tracker almost two years ago).

PR #3420, opened 2023-03-29
ci: add spell checker and fix existing typo

hezhizhen:


typos.toml
```diff
@@ -0,0 +1,12 @@
+# See https://github.com/crate-ci/typos/blob/master/docs/reference.md to configure typos
```
a8m, line 1: Move this file to the .github directory.

a8m (commented):
Thanks, @hezhizhen. Please, see my comment

typos.toml
hezhizhen: Why? It is supposed to be placed in the ROOT.

hezhizhen: When you run typos locally, it will search `typos.toml` in the current directory and use it.

a8m: Because I try to keep the root directory thin as readers don't get values from seeing it. The `.all-contributorsrc` and `.golangci-lint` will be moved as well when I get to this.

hezhizhen: Or rename it as `.typos.toml`?

hezhizhen: > Because I try to keep the root directory thin as readers don't get values from seeing it. The `.all-contributorsrc` and `.golangci-lint` will be moved as well when I get to this.

I see. Do you mind it if I add a Makefile target?

a8m: But we don't have or use Makefile in this project. Are there any problems with storing this file inside the `.github` directory?

hezhizhen: It 's not about github actions, but about locally run binaries. I need to know where the config file is and specify them. So it's better to add it to Makefile.

hezhizhen: But I get what you mean. Let me do it.

a8m: > It's not about github actions, but about locally run binaries.

I appreciate the help, but I'm not going to run this locally. 

Thinking of it now - don't we have something like this in golangci?

hezhizhen: > > It's not about github actions, but about locally run binaries.
> 
> I appreciate the help, but I'm not going to run this locally.
> 
> Thinking of it now - don't we have something like this in golangci?

I see.

.github/workflows/typos.toml
```diff
@@ -0,0 +1,12 @@
+# See https://github.com/crate-ci/typos/blob/master/docs/reference.md to configure typos
```
hezhizhen, line 1: @a8m How about this place?

hezhizhen:
@a8m Ready to merge?

a8m (approved):
💯

PR #3435, opened 2023-04-03
doc: replace example links

noamcattan:


a8m (approved):
I'll squash the two commits before merging.

PR #3448, opened 2023-04-07
entc/gen: ensure package alias is used correctly with enum

### LINKED ISSUE #3444: If the entity name is User, when importing other packages with name user. generated code is malformed.
- [ ] The issue is present in the latest release.
- [ ] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Steps to Reproduce 🕹

Steps:

1.
2.
3.
4.

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.19.4  |
| Ent         | 0.11.10   |

crossworth:
This fixes #3444

I created a new package (`packagealias`) to keep the reproducible example short, but I can combine it with other package if needed.

The test will not check for the correctness on the database level (I'm pretty sure other tests already cover it), since the problem generates invalid code, the fact that the test is able to run in the first place is already a test for the correct behavior.


entc/gen/type.go
```diff
@@ -1078,6 +1078,14 @@ func (f Field) BuilderField() string {
 	return builderField(f.Name)
 }
 
+// TypeName returns the type name for the field.
+func (f Field) TypeName() string {
+	if n := f.typ.PackageAlias(); n != "" && f.IsEnum() {
+		return strings.Replace(f.Type.String(), f.typ.PackageDir(), f.typ.PackageAlias(), 1)
+	}
+	return f.Type.String()
+}
```
a8m, line 1087: Please, see: https://github.com/ent/ent/blob/master/entc/gen/type.go#L981-L982
I believe we can handle this in `aliases`. Check out: https://github.com/ent/ent/blob/master/entc/gen/type.go#L1006-L1035

entc/integration/packagealias/ent/generate.go
```diff
@@ -0,0 +1,7 @@
+// Copyright 2019-present Facebook Inc. All rights reserved.
```
a8m, line 1: I prefer to keep this in `integration/ent`. There's already a similar use case, but for GoType (not enums): https://github.com/ent/ent/blob/master/entc/integration/ent/schema/task.go#L26-L27

PR #3451, opened 2023-04-08
actual pgx version in example

sashamelentyev:


a8m (approved):
Thanks, @sashamelentyev

PR #3452, opened 2023-04-08
add go1.20

sashamelentyev:


a8m:
Thanks, but I don’t think we need to update this doc on every Go release.

PR #3454, opened 2023-04-08
add is int 32 method

sashamelentyev:


a8m (commented):
Hey! Thanks for the contribution. 

What’s the use case for this?

sashamelentyev:
@a8m hey! Thanks for ent!
We have `IsInt` and `IsInt64`, I think needing IsInt32. But if no - we can close this PR. Sorry for noise.

a8m:
The codegen uses the IsInt/IsInt64 methods. This one is not needed/used by Ent itself.

Issue #1381, opened 2021-03-24
Add interface for field builder to make partial prebuild field


wenerme:
- [x] I have searched the [issues](https://github.com/facebook/ent/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

```go
func IDField(name string)  StringFieldBuilder  {
	return field.String(name).GoType(models.ID(""))
}
```

Current return `*stringBuilder`, can not make this happen.

## Motivation 🔦
* Customize ID field are spread everywhere in schema, every id field need to repeat this
* Make Customized field build easier, like a helper to inject customized default handling

giautm:
Hi, you can use [Schema Mixin](https://entgo.io/docs/schema-mixin) to share common fields with all entities.

wenerme:
@giautm I fixed the demo code, what I want is make

```go
func (Ent)Fields(name string) []ent.Field {
	return []ent.Field {
		field.String("productID").GoType(models.ID("")),
		field.String("orderID").GoType(models.ID("")),
		field.String("accountID").GoType(models.ID("")),
		field.String("contactID").GoType(models.ID("")),
	}
}
```

become

```go
func (Ent)Fields(name string) []ent.Field {
	return []ent.Field {
		IDField("productID"),
		IDField("orderID"),
		IDField("accountID"),
		IDField("contactID"),
	}
}
```

a8m:
You can return an `ent.Field` from your helper function.

```go
func IDField(name string) ent.Field  {
	return field.String(name).GoType(models.ID(""))
}
```

But if you want to have something like `IDField("foo").Optional()`, it won't work. 

If your use-case requires changing all fields (e.g. adding an annotation to all fields), you can do it using the `Descriptor` method, as [in this example](https://github.com/ent/ent/blob/master/schema/mixin/mixin.go#L105-L112).

wenerme:
@a8m make `IDField("foo").Optional()` work that's why I propose the change.

Maybe I should use the Descriptor, write a build from scratch.

nix-sys:
Hi there,
@a8m Following up on the field template idea: Is there a reason why the fieldBuilders (stringBuilder, etc.) are kept as unexported variables? 

For our own use case it would save us a ton of time if the fieldBuilders were accessible by our own packages:
- we use nanoids with custom prefix
- have company wide policies like *no empty strings* `""` in the db (either `null` or `"some string"`)
- various custom go / sql types for business specific reasons. 

Our field definitions look like this:
```go
[...]

PR #3472, opened 2023-04-12
schema/fields changed fieldBuilders to exported types

nxso:
Allows to create **field templates** as mentioned here https://github.com/ent/ent/issues/1381:
- fields can be defined in a much cleaner, consistent and flexible way
- simplifies to comply with company wide schema policies
- allows to create tests against complex field definitions once

A function like
```go
func PrimaryNanoID(fieldname string, prefix prefixes.Prefix) *field.StringBuilder {
	fieldname = strings.ToLower(prefix)

	return field.String("id").
		StorageKey(fieldname).
		GoType(nanoid.NanoID("")).
		DefaultFunc(prefix.NewNanoID).
		Match(regexp.MustCompile("^" + string(prefix) + "_[A-Za-z0-9]{19,25}$")).
		NotEmpty().
		Immutable().
                Annotations(
	        	entsql.Checks(map[string]string{
		        	fmt.Sprintf(`%v_valid_key`, fieldname):        fmt.Sprintf(`(%v ~ '^%v_[A-Za-z0-9]*$')`, fieldname, string(prefix)),
			       fmt.Sprintf(`%v_non_empty_string`, fieldname): fmt.Sprintf(`(utils.null_or_non_empty_trimmed_string(%v))`, fieldname),
	      }),
	)
}

```

makes it possible to reduce the field definition on each schema to one line (and remains expandable). E.g.
```go
customfield.PrimaryNanoID(prefixes.FileID).Comment("This is the primary key")
```

### DIFF
--- a/schema/field/field.go
+++ b/schema/field/field.go
@@ -20,8 +20,8 @@ import (
 )
 
 // String returns a new Field with type string.
-func String(name string) *stringBuilder {
-	return &stringBuilder{&Descriptor{
+func String(name string) *StringBuilder {
+	return &StringBuilder{&Descriptor{
 		Name: name,
 		Info: &TypeInfo{Type: TypeString},
 	}}
@@ -29,8 +29,8 @@ func String(name string) *stringBuilder {
 
 // Text returns a new string field without limitation on the size.
 // In MySQL, it is the "longtext" type, but in SQLite and Gremlin it has no effect.
-func Text(name string) *stringBuilder {
-	return &stringBuilder{&Descriptor{
+func Text(name string) *StringBuilder {
+	return &StringBuilder{&Descriptor{
 		Name: name,
 		Size: math.MaxInt32,
 		Info: &TypeInfo{Type: TypeString},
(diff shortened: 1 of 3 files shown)


a8m:
Thanks for the contribution, @nxso. At the moment, we don't plan to export these builders because there are upcoming changes to make their implementations more type-safe using generics. I intend to address this in future versions. Since we cannot introduce breaking changes to the ent/schema, I prefer not to export the actual types at this stage.

Having said that, the issue described in https://github.com/ent/ent/issues/1381 can be solved the same as we do in the `schema/mixin` package and also internally. Since all builders expose the `Descriptor` method, you can add to your custom  constructors additional options as follows:

```go
type FieldOption func(ent.Field)

func WithFieldComment(comment string) FieldOption {
	return func(f ent.Field) {
		f.Descriptor().Comment = comment
	}
}

func WithFieldAnnotations(annotations ...schema.Annotation) FieldOption {
	return func(f ent.Field) {
		f.Descriptor().Annotations = append(f.Descriptor().Annotations, annotations...)
	}
}

 // More options.

// Custom constructors.
func PrimaryNanoID(fieldname string, prefix prefixes.Prefix, opts ...FieldOption) ent.Field {
	fd := field.String("id").
		StorageKey(fieldname)
		// ...

	// Set field options.
	for _, opt := range opts {
		opt(fd)
	}
	return fd
}
```

Then, the usage is as follows:

```go
PrimaryNanoID("id", prefixes.User, WithFieldComment("User ID"), WithFieldAnnotations(...))
CustomIntField("name", WithFieldComment("..."), ...)
```

I actually prefer this approach because it can be used for any field type.

PR #3484, opened 2023-04-16
feat: add Len for plural

sashamelentyev:


a8m:
Hey! Thanks for the contribution, but this method is not needed. The standard one is preferred IMO.

You can always use custom templates to generate such methods for your projects. Check out: https://entgo.io/docs/templates

PR #3485, opened 2023-04-17
Feat/dynamic schema

Gitigi:
# Set table schema dynamically

Am working on this project that uses schema to separate different clients. So I need to be able to dynamically set the a global schema name in which all tables will be searched from.

Am using context to set the global schema name that will be used for all queries

I have build on top of `sql/schemaconfig` feature flag so that we can also have the ability to set a constant schema name for a particular table

## Notable changes

Changes outside `sql/schemaconfig` feature flag include

- I have updated `createSpec(ctx context.Context)` to accept context as a parameter. Am note sure why it was missing as all other functions that make db queries seem to have context as a parameter

- I have also updated all calls to `createSpec` to include context


giautm:
Hello @Gitigi, why you don't create an Ent client for each connection per schema to the database?

Gitigi:
Hi @giautm I wanted  a single instance of my backend application to be able to serve multiple tenants (customers)

Gitigi:
Yes that will work. Maybe I am overthink this but creating a new connection per request might be a bit taxing. I can't remember properly if it was pg_pool or pg_bouncer that might help if you take this approach.

For me using the schema approach was just the easiest way to get things started. 
But in the future I will have to move to Row-level security (RLS) as a meaning of tenant data isolation.

lloydchandran:
This is an awesome feature. I was looking for an orm that can support schema-based tenancy without redeploying the service layer. @a8m is it possible to merge this PR to main and release?

lloydchandran (commented):
LGTM

robinhuiser (approved):

robinhuiser:
Just to give some background here... 

In Enterprise software for large financial institutions it is not accepted to share storage with other tenants. Hence... the need to set a context for an incoming tenant request without having to setup a new database connection (performance wise this will not work). RLS (as @Gitigi suggested as a future improvement) is not an option in this situation due to the strict regulations for data segregation. 

Hope this helps moving this PR forward.

lloydchandran:
Adding to @robinhuiser comment 

This also help to design a system where the schema is created at the time of tenant onboarding based on the tenant id . The service layer is agnostic to this schema and does not need to re deploy to add a new connection pool and ent.Client for the tenant . 

This means using ent we can support shared database model and schema per tenant with the same service layer. Both tenant id and tenant scheme name is present in the ctx

Excited !

shinusuresh:
This appears to be a missing feature that needs attention from the maintainers.

a8m:
As @giautm mentioned above, we actually use a schema/database-per-tenant approach, but separate an ent.Client per tenant. See: https://github.com/ent/ent/pull/3485#issuecomment-1840208144

We will try to release an architecture blogpost and might also OSS some of our codebase for this sometime soon.

lloydchandran:
@a8m . If am not wrong the service layer needs to know about the tenants at deployment time using the proposed approach by @giautm . This means we need to redeploy the service layer the time a new tenant is onboarded. Or is there a better way to do this ?

a8m:
No, see @giautm’s explanation; 

> When the request hit to application, it has a HTTP header that let the application know which tenant it need to working on. So, it open the connection in the middleware and attach to request's context to passing down to next level.
> 
> Other logic just take the Ent client from the context to query and update data. This design allow me security the connection to single point, also able to point the Ent client to a separated database instance if needed.

PR #3492, opened 2023-04-19
dialect/sql/sqlgraph: replace id IN with EXISTS for HasNeighbors predicates

a8m:


mgabeler-lee-6rs:
Borne out of slack discussion: https://gophers.slack.com/archives/C01FMSQDT53/p1681910810225949

In my case, I had two tables, tracking messages and their deliveries for a (google) pubsub emulator. I had ~160k rows in each, with one delivery row for each message row. The code in question was looking for messages with no remaining deliveries to clean them up in batches (at a high level: get messages older than a threshold age with no associated deliveries, with a limit).

With the prior implementation, this query was often taking over 15 minutes to run (and never less than 6 minutes according to my metrics) due to the query having to do a seq scan on both tables, and being unable to use any indexes such as the one indexing the delivery->message FK field.

With this implementation, the query takes 500ms or less -- it's somewhat variable depending on other system load and whether there are any rows for it to find -- finding rows results in shorter runtimes because the `limit` clause allows it to short circuit I think. Here is the "after" data:
image

Issue #3554, opened 2023-05-23
ent.Open() overrides any custom driver provided with the default driver


Kranex:
We'd like to wrap the default MySQL database driver with a sentry tracer, but the generated `ent.Open()` function from `ent/client.go` overrides any provided driver with the default. I've already fixed this in my own fork, I'll be opening a PR shortly after this issue.

- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

`Open()` applies the default Driver option after all other options provided to `Open()`.

## Expected Behavior 🤔

`Open()` should apply the default Driver option first, then apply all other options provided to `Open()`.

## Steps to Reproduce 🕹

Provide any alternative or wrapped driver to the while opening the database.

```go
	entClient, err = Open("mysql", dbUri, func(c *config) {
		c.driver = SentryTracer(c.driver)
	})
```

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.19.4 |
| Ent         | 0.11.9   |
| Database    | MySQL   |
| Driver      | https://github.com/go-sql-driver/mysql |

a8m:
Hey, @Kranex 👋 
I'll recommend using a different (more standard) way to wrap the underlying database driver without dealing with the internal `config` object:

```go
// Open the database connection.
db, err := sql.Open(dialect.MySQL, "...")
if err != nil {
	log.Fatal("opening database", err)
}
// Decorates the dialect/sql.Driver with a custom driver (e.g., entcache.Driver).
drv := entcache.NewDriver(db)
// Create an ent.Client by passing the driver as an option.
client := ent.NewClient(ent.Driver(drv))
```

We often use this pattern internally and externally in the integrations we open-sourced.

jerbob92:
@a8m thanks for the workaround! It would work for us but we would rather use the generated Open client which also seems to provide a nice way to override options on the client, so don't you think that should be fixed?

a8m:
Hey, @jerbob92. The proposed solution above is actually the standard and recommended way to decorate drivers. As I mentioned previously, we frequently use it to wrap our drivers multiple times for telemetry, caching, tenancy, logging, etc. 

I presume that if you pass `func(*config)` to `Open`, this function is implemented within the generated package and interacts with the internal types (I don't think it's necessary for saving 1-2 lines),  am I right?

[...]

PR #3555, opened 2023-05-23
Fix dialect client open options order

Kranex:
Fixes https://github.com/ent/ent/issues/3554

Prepends the default driver to the additional options provided to `Open()` rather than appending it.

I also fixed the same issue gremlin driver while I was at it.


a8m:
Hey, @Kranex 👋 

Thanks for the contribution. Please, see my comment at https://github.com/ent/ent/issues/3554#issuecomment-1560002857.

PR #3651, opened 2023-07-18
mark field as cleared when clearing edges

yasser-sobhy:
I am writing an audit log mixin for one of my apps and found what I think is a bug in the generated edge clearing code. 

Let's assume I have a User entity with LockedBy edge. Clearing the LockedBy edge won't mark the field LockedByID as cleared when I call `user_mutation.ClearedFields()`

The following code is supposed to clear both the edge and the field and I should be able to tell that the field was cleared when I call `user_mutation.ClearedFields()`
```
err := database.Client.User.UpdateOneID(d.ID).ClearLockedBy().Exec(c)
```

but it actually doesn't.  It does clear the field in the db but I can't detect that change through `m.ClearedFields()`
 
Generated code before this change
```
// ClearLockedBy clears the "manually_locked_by" edge to the User entity.
func (m *DocMutation) ClearLockedBy() {
	m.clearedmanually_locked_by = true
}

// LockedByCleared reports if the "manually_locked_by" edge to the User entity was cleared.
func (m *DocMutation) LockedByCleared() bool {
	return m.LockedByIDCleared() || m.clearedmanually_locked_by
}
```

Generated code after this change
```
// ClearLockedBy clears the "manually_locked_by" edge to the User entity.
func (m *DocMutation) ClearLockedBy() {
	m.clearedmanually_locked_by = true
	m.clearedFields[doc.FieldLockedByID] = struct{}{}
}

// LockedByCleared reports if the "manually_locked_by" edge to the User entity was cleared.
func (m *DocMutation) LockedByCleared() bool {
	return m.LockedByIDCleared() || m.clearedmanually_locked_by
}
```


yasser-sobhy:
@a8m I pushed one more commit to update the generated code in `entc/integration`.  

Running `go generate ./..` in `entc` didn't work, I ran it in `entc/integration` and updated the contribution guide

a8m:
Thanks, @yasser-sobhy. Make sure you also run `go mod tidy` on these two directories.

a8m (approved):
Thanks for your contribution, @yasser-sobhy 🙏🏻

PR #3668, opened 2023-07-31
entc/gen: remove dollar sign from table name definitions

### LINKED ISSUE #3657: Dollar sign in table names
Is there currently a way to use ent with tables containing a dollar sign?

Let's take table "relations$person" for example:

```
// Annotations of the Person.
func (Person) Annotations() []schema.Annotation {
	return []schema.Annotation{
		schema.Annotation(entsql.Annotation{Table: "relations$persons"}),
	}
}
```

Produces an error after generating this:

```
var (
		// Relations$personsColumns holds the columns for the "relations$persons" table.
		Relations$personsColumns = []*schema.Column{
				{ Name: "id", Type: field.TypeInt, Increment: true,}
}
```

Not sure if annotations is the right way to do this but if there is another way I would be glad to know. If not then we definitely need some sanitization over here since it's named according MySQL specs.

GreatChiefPowhatan:
Fixes #3657


GreatChiefPowhatan:
@a8m Thanks for your review.

I would definitely agree that having the var name generation responsibility as a receiver method on the table struct would be an improvement in terms of code organisation and future scalability.

Tackling the future scalability, let's just make it responsible to generate a valid token no matter the input as you suggested right away.

It sounds like this logic could be reused by other parts of the code as well since it's a fundamental part of code generation. With that though I would suggest to put the actual logic somewhere shareable?

Also let's scope it's domain to the Go standards and either create a PascalCase or camelCase variable name by passing a boolean controlling exported?

```golang
func (t *Table) VarName() string {
    return codegen.StringToIdentifier(t.TableName, true)
}
```

```golang
StringToIdentifier(s string, exported bool) string {
// Filter out all illegal chars

// Filter out wrongly placed legal chars (starting with number, etc..)

// Transform to camelCase if not exportable

// Check if leftover is keyword

// Transform to PascalCase if exported

// Check if leftover is exported (not sure what func PascalCase(_someString) does ...)
}
```

PR #3701, opened 2023-08-18
Feat: added nil checks to sqlgraph constraint errors

lucvankessel:
Adding these checks makes sure that when the error is nil these do not panic because of a nil pointer exception.
It is just a minor thing but something that bugged me when using these functions.

The ```IsConstraintError``` does not need it due to the use of ```errors.As``` which has these same checks.

### DIFF
--- a/dialect/sql/sqlgraph/errors.go
+++ b/dialect/sql/sqlgraph/errors.go
@@ -21,7 +21,6 @@ func IsUniqueConstraintError(err error) bool {
 	if err == nil {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


a8m (approved):
Thanks for the contribution, @lucvankessel. Please, address the comments and I will merge this

lucvankessel:
@a8m 👍 resolved suggested changes

a8m (approved):
Thanks for the contribution, @lucvankessel

PR #3770, opened 2023-09-30
Fix example in docs/crud/#create-many

adamzapasnik:
we need to use the given pointer `c`  to update it instead of creating a new `ent.PetCreate`, right? 🙈


a8m (approved):
Thanks for the contribution, @adamzapasnik 🫶

Issue #3722, opened 2023-08-29
GROUP BY problem with OrderByField in O2O


huoayi:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

- Following the docs of ent, query User and want it to be ordered by its O2O edge
- DB failed with GROUP BY synatx error

### User Schema
```go
// User holds the schema definition for the User entity.
type User struct {
	ent.Schema
}

// Fields of the User.
func (User) Fields() []ent.Field {
	return []ent.Field{
		field.String("name").Default("Unknown"),
		field.Int64("age").Default(0),
		field.String("nickname").Default(""),
	}
}

// Edges of the User.
func (User) Edges() []ent.Edge {
	return []ent.Edge{
		edge.To("card", Card.Type).
			Unique(),
	}
}

```

### Card Schema
```go
// Card holds the schema definition for the Card entity.
type Card struct {
	ent.Schema
}

// Fields of the Card.
func (Card) Fields() []ent.Field {
	return []ent.Field{
		field.String("number").Default(""),
		field.Time("expired").Default(pkg.ZeroTime),
		field.Int("owner_id"),
	}
}

// Edges of the Card.
func (Card) Edges() []ent.Edge {
	return []ent.Edge{
		edge.From("owner", User.Type).
			Ref("card").
			Unique().
			Field("owner_id").
			// We add the "Required" method to the builder
			// to make this edge required on entity creation.
			// i.e. Card cannot be created without its owner.
			Required(),
	}
}
```

### Main.go
```go
import (
	"context"
	"github.com/sirupsen/logrus"
	"testredis/db"
	"testredis/ent/card"
	"testredis/ent/user"
)

func main() {
	ctx := context.Background()
	all, err := db.DB.Debug().User.Query().Order(user.ByCardField(card.FieldNumber)).All(ctx)
	if err != nil {
		logrus.Info(err)
		return
	}
	logrus.Info(all)
}

```

### Console
```shell
2023/08/29 14:51:55 driver.Query: query=SELECT "users"."id", "users"."name", "users"."age", "users"."nickname" FROM "users" LEFT JOIN (SELECT "cards"."owner_id", "cards"."number" FROM "cards" GROUP BY "cards"."owner_id") AS "t1" ON "users"."id" = "t1"."owner_id" ORDER BY "t1"."number" NULLS FIRST args=[]
[...]

PR #3783, opened 2023-10-11
dialect/sql/sqlgraph: allow OrderByField without aggregate function on OrderByNeighborTerms/O2M

crossworth:
This allows for a fix to #3722

```go
client.Card.Query().Order(func(s *sql.Selector) {
	sqlgraph.OrderByNeighborTerms(s,
		sqlgraph.NewStep(
			sqlgraph.From(card.Table, card.FieldID),
			sqlgraph.To(user.Table, user.FieldID),
			sqlgraph.Edge(
				sqlgraph.O2M,
				false,
				card.OwnerTable,
				card.OwnerColumn,
				card.FieldNumber, // <--- This PR adds support for this. 
			),
		),
		sql.OrderByField(card.FieldNumber, sql.OrderDesc()),
	)
}).IDsX(ctx)
```

Previous the query generated was:
```sql
SELECT "cards"."id" FROM "cards" LEFT JOIN (SELECT "cards"."owner_id", "cards"."number" FROM "cards" GROUP BY "cards"."owner_id") AS "t1" ON "cards"."id" = "t1"."owner_id" ORDER BY "t1"."number" DESC NULLS LAST
```

Generating the error:
```
ERROR: column "cards.number" must appear in the GROUP BY clause or be used in an aggregate function (SQLSTATE 42803)
```

With this PR, the query generate is:
```sql
SELECT "cards"."id" FROM "cards" LEFT JOIN (SELECT "cards"."owner_id", "cards"."number" FROM "cards" GROUP BY "cards"."owner_id", "cards"."number") AS "t1" ON "cards"."id" = "t1"."owner_id" ORDER BY "t1"."number" DESC NULLS LAST
```

**This PR will not solve the problem when using `ByCardField` method, it only allows for possible fix using Ent constructors.**

### DIFF
--- a/dialect/sql/sqlgraph/graph.go
+++ b/dialect/sql/sqlgraph/graph.go
@@ -511,9 +511,13 @@ func OrderByNeighborTerms(q *sql.Selector, s *Step, opts ...sql.OrderTerm) {
 			On(q.C(s.From.Column), join.C(pk2))
 	case s.ToEdgeOwner():
 		toT := build.Table(s.Edge.Table).Schema(s.Edge.Schema)
+		var groupByColumns []string
+		for _, c := range s.Edge.Columns {
+			groupByColumns = append(groupByColumns, toT.C(c))
+		}
 		join = build.Select(toT.C(s.Edge.Columns[0])).
 			From(toT).
-			GroupBy(toT.C(s.Edge.Columns[0]))
+			GroupBy(groupByColumns...)
 		selectTerms(join, opts)
 		q.LeftJoin(join).
 			On(q.C(s.From.Column), join.C(s.Edge.Columns[0]))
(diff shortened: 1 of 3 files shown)


a8m:
Thanks for the contribution, @crossworth. However, a `sqlgraph.Edge` describes graph edge using SQL relations, and adding `card.FieldNumber` should not be part of it. The group by info should be extracted from the "order field" option.

Let me know if you need help with this.

crossworth:
hi @a8m, I'm not sure I understood.

The option calls `sqlgraph.OrderByNeighborTerms` to build the query:
https://github.com/ent/ent/blob/8a848a6475861354447d6338d9c10849d0683b0e/entc/integration/ent/user/user.go#L312

Ofcourse would be possible to extract the info from the order option, but would that require this change as well, since the invalid query is generated inside `OrderByNeighborTerms`?

Is there something else that I'm missing?

crossworth:
PR #3943 fix the #3722 issue and is way smaller and simpler solution.

Closing this one.

PR #3792, opened 2023-10-17
doc/website: we stand with israel; we we israel

a8m:
Surprisingly (not really), this is not that obvious these days.


larrasket:
@a8m grow up.

PR #3806, opened 2023-10-30
doc/md: drop v10 support and add v15 for PostgreSQL

oinume:
## WHAT
SSIA

## WHY
PostgreSQL v15 has been tested in CI so we can say that ent officially supports v15.
https://github.com/ent/ent/pull/3152


a8m:
Thanks, @oinume 🙏

PR #3830, opened 2023-11-13
dialect/sql/sqlgraph: fix scanning non-numeric IDs

### LINKED ISSUE #3829: CreateBulk fails when using pgx driver and non-numeric ID field
We use string (varchar) type IDs as primary keys in our Postgres database.  We also use the ubiquitous [jackc/pgx/v5](https://github.com/jackc/pgx) driver.

During CreateBulk operations, we receive the following error:
```
insert nodes to table "<table>": can't scan into dest[0]: cannot scan varchar (OID 1043) in text format into *driver.Value
```

## Current Behavior 😯

The following error is thrown:
```
insert nodes to table "<table>": can't scan into dest[0]: cannot scan varchar (OID 1043) in text format into *driver.Value
```

## Expected Behavior 🤔

The CreateBulk inserts the rows without error.

## Steps to Reproduce 🕹

A small ent schema might be constructed like:
```
type Test struct {
	ent.Schema
}

func (Test) Fields() []ent.Field {
	return []ent.Field{
		field.String("id").Unique().NotEmpty().MaxLen(32).Immutable(),
		field.String("name").MaxLen(128).NotEmpty().Unique(),
		field.Text("description").Optional().Nillable(),
	}
}
``` 

Then running a CreateBulk on the entity:
```
bulkInsert := make([]*ent.TestCreate, 10)
for i := 0; i < 10; i++ {
	
    bulkInsert[i] = tx.Test.
	Create().
	SetID(fmt.Sprintf("id%d", i)).
	SetName(fmt.Sprintf("name%d", i))
}

// Execute bulk insert
return tx.Test.CreateBulk(bulkInsert...).Exec(ctx)
```

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.21.4  |
| Ent         | 0.12.5  |
| Database    | PostgreSQL 14 |
| Driver      | https://github.com/jackc/pgx  |

subfxnet:
Fixes #3829 

Use the same logic for row scanning between `*creator.insertLastID` and `batchCreator.insertLastIDs`.

### DIFF
--- a/dialect/sql/sqlgraph/graph.go
+++ b/dialect/sql/sqlgraph/graph.go
@@ -1924,7 +1924,14 @@ func (c *batchCreator) insertLastIDs(ctx context.Context, tx dialect.ExecQuerier
 		defer rows.Close()
 		for i := 0; rows.Next(); i++ {
 			node := c.Nodes[i]
-			if node.ID.Type.Numeric() {
+			switch _, ok := node.ID.Value.(field.ValueScanner); {
+			case ok:
+				// If the ID implements the sql.Scanner
+				// interface it should be a pointer type.
+				if err := rows.Scan(node.ID.Value); err != nil {
+					return err
+				}
+			case node.ID.Type.Numeric():
 				// Normalize the type to int64 to make it looks
 				// like LastInsertId.
 				var id int64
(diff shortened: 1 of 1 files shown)


a8m (approved):
Thanks for the contribution, @subfxnet 🙏

PR #3838, opened 2023-11-21
entc/gen: fix loss of int64 precision during JSON type decoding

eiixy:


giautm:
Hello @eiixy, can you please add an example or testcase for this change?

eiixy:
> Hello @eiixy, can you please add an example or testcase for this change?

Sure, I have added to example

PR #3841, opened 2023-11-23
fix: swap the order of handling order and pred  in the graph select func

michaelcaulley:
multi order may generate predicates, so running the order funcs first allows reusing aliases. 

If the predicates run first if there is a multiOrder field that generates it's own predicates, the predicates are unable to find the alias of the join and errors not finding the column on the table.

Related to: https://github.com/ent/contrib/pull/559


a8m (approved):

a8m:
Thanks for the contribution, @michaelcaulley 🚀

PR #3844, opened 2023-11-25
Document Small Fix update to `updateOne`

mztnnrt:


a8m:
This change is not correct, but I appreciate the willingness to help

mztnnrt:
Thanks for your reply.

> Update an entity that was returned from the database.
> https://github.com/mztnnrt/ent/blob/24c3b0e1c249b3b1827329d7f8c5fde06d0c43ad/doc/md/crud.mdx#L177-L187

Is my understanding below is wrong.
`Update().Save()` return `(int, error)`, So to return struct that is updated, we need to use `UpdateOne()`, `UpdateOneID()`.

a8m:
The `Update` method is generated for all entities. See: https://github.com/ent/ent/blob/master/examples/start/ent/user.go#L130

Issue #3858, opened 2023-12-13
Occurrence of duplication with variable names when the short form of the schema structure name is "ok"


nandehu0323:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

In the current auto-generation functions of ent, there exists a function named "assignValues" and "Unwrap".

https://github.com/ent/ent/blob/5755614d3f3e6788356569bbc01ec087c65dae31/entc/gen/template/dialect/sql/decode.tmpl#L57-L108

In the these function, there's a section evaluating the feasibility of casts.
I've discovered an issue where the return value ok sometimes conflicts with the abbreviated forms in the schema structures.

```go
// assignValues assigns the values that were returned from sql.Rows (after scanning)
// to the OwnerKey fields.
func (ok *OwnerKey) assignValues(columns []string, values []any) error {
	if m, n := len(values), len(columns); m < n {
		return fmt.Errorf("mismatch number of scan values: %d != %d", m, n)
	}
	for i := range columns {
		switch columns[i] {
		case ownerkey.FieldID:
			if value, ok := values[i].(*sql.NullString); !ok {
				return fmt.Errorf("unexpected type %T for field id", values[i])
			} else if value.Valid {
				ok.ID = value.String
			}
		...
		default:
			ok.selectValues.Set(columns[i], values[i])
		}
	}
	return nil
}

// Unwrap unwraps the OwnerKey entity that was returned from a transaction after it was closed,
// so that all future queries will be executed through the driver which created the transaction.
func (ok *OwnerKey) Unwrap() *OwnerKey {
	_tx, ok := ok.config.driver.(*txDriver)
	if !ok {
		panic("ent: OwnerKey is not a transactional entity")
	}
	ok.config.driver = _tx.drv
	return ok
}
```

## Expected Behavior 🤔

```go
// assignValues assigns the values that were returned from sql.Rows (after scanning)
// to the OwnerKey fields.
func (ok *OwnerKey) assignValues(columns []string, values []any) error {
	if m, n := len(values), len(columns); m < n {
		return fmt.Errorf("mismatch number of scan values: %d != %d", m, n)
	}
	for i := range columns {
		switch columns[i] {
		case ownerkey.FieldID:
			if value, _ok := values[i].(*sql.NullString); !_ok {
				return fmt.Errorf("unexpected type %T for field id", values[i])
			} else if value.Valid {
[...]

PR #3859, opened 2023-12-13
entc/gen: avoid codegen conflict with specific (ok) receivers

nandehu0323:
Fixed: #3858


giautm:
Close as duplicated from https://github.com/ent/ent/pull/4355

PR #3862, opened 2023-12-18
entc/gen: allow skipping of format step

philip-peterson:
In the development cycle, sometimes it is needed to run Ent generate multiple times in a row (such as when rebasing), and this can be quite costly in terms of time (in my case, on the order of minutes for each generation). A large amount of the time is just spent running `goimports` on all the files, so this feature allows the user to skip that step for the sake of time.

For my particular use case, we would just be running `goimports` later in a CI step required before merging, so it just moves it out of the critical path of development iteration and into the "ready to merge" phase.

### DIFF
--- a/entc/gen/feature.go
+++ b/entc/gen/feature.go
@@ -51,6 +51,14 @@ var (
 		Description: "NamedEdges provides an API for eager-loading edges with dynamic names",
 	}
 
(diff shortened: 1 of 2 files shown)


philipsilo:
After some more usage, I am not so sure this will work since the `goimports` step removes some duplicated code. I guess my follow up question though is whether the call to `goimports` can possibly be removed by adding some extra logic to ent?

Issue #3883, opened 2024-01-03
Privacy filter does not apply to IDs inside hooks for *One operations


mwoelk:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

When a privacy rule is applied to a mutation, it correctly filters the mutation itself.
When using hooks on UpdateOne or DeleteOne mutations, the mutations IDs() method does not reflect the expected filtering based on the privacy rule as is the case with Update or Delete mutations.

## Expected Behavior 🤔

When using a hook on UpdateOne or DeleteOne mutations, the result of `mutation.IDs()` should be filtered according to the privacy rule as well. 

## Steps to Reproduce 🕹

https://github.com/mwoelk/bug

mwoelk:
Would be an easy fix by modifying the generation of `ent/mutation.go` by checking during `OpUpdateOne` and `OpDeleteOne` if there are any predicates applied and if yes add another predicate with the respective id and run it against the database to see if the ID is still included. In case there are no predicates defined, we can keep the old behaviour and directly return the id.

Example:
```go
// IDs queries the database and returns the entity ids that match the mutation's predicate.
// That means, if the mutation is applied within a transaction with an isolation level such
// as sql.LevelSerializable, the returned ids match the ids of the rows that will be updated
// or updated by the mutation.
func (m *UserMutation) IDs(ctx context.Context) ([]int, error) {
	switch {
	case m.op.Is(OpUpdateOne | OpDeleteOne):
		id, exists := m.ID()
		if exists && len(m.predicates) > 0 {
			m.predicates = append(m.predicates, user.ID(id))
			return m.Client().User.Query().Where(m.predicates...).IDs(ctx)
		} else if exists {
			return []int{id}, nil
		}
		fallthrough
	case m.op.Is(OpUpdate | OpDelete):
		return m.Client().User.Query().Where(m.predicates...).IDs(ctx)
	default:
		return nil, fmt.Errorf("IDs is not allowed on %s operations", m.op)
	}
}
```

PR #3884, opened 2024-01-03
Fix IDs() to respect mutation predicates for UpdateOne and DeleteOne 

mwoelk:
# Fixes issue #3883

## Summary

Updated the generated `IDs()` function on mutations.

Checks during `OpUpdateOne` and `OpDeleteOne` if there are any predicates set on the mutation and if yes, adds another predicate with the id of the mutation and runs it against the database to see if the ID is still included after all predicates have been applied. 

In case there are no predicates defined, the old behaviour is kept and directly returns the id (to save a query).


mwoelk:
@a8m looking forward to your review 👀

entc/gen/template/builder/mutation.tmpl
```diff
@@ -172,7 +172,10 @@ func (m {{ $mutation }}) Tx() (*Tx, error) {
 		switch {
 		case m.op.Is(OpUpdateOne | OpDeleteOne):
 			id, exists := m.ID()
-			if exists {
+			if exists && len(m.predicates) > 0 {
+				m.predicates = append(m.predicates, {{ $n.Package }}.ID(id))
+				return m.Client().{{ $n.Name }}.Query().Where(m.predicates...).IDs(ctx)
```
a8m, line 177: This returns zero IDs in case the extra predicates are falsey. The extra predicates are respected/evaluated in a different layer of the codegen: https://github.com/ent/ent/blob/master/entc/gen/template/dialect/sql/update.tmpl#L86-L92

mwoelk: That's exactly the intention behind this PR :) Take a look at the issue and example repo in the issue.
When using privacy filters, they are not applied when using the IDs() method inside hooks for UpdateOne and DeleteOne mutations.

This leads to a simple issue. E.g. a privacy filter is defined so that only users that match certain criteria can be mutated. Now there is a hook running OnDelete|OnDeleteOne that should do some cleanup and therefore simply takes the IDs() of the mutation to see which items should be deleted. Although the mutation will not run, the hook will be called with the id of the entity leading to stuff getting cleaned up that souldn't be. 

In our concrete use-case the cleanup is non-reversible due to it calling external APIs and is triggered by another hook on an edge related to the user which requires the user to still exist. So mutating first in the hook and checking for an error is not possible as well as running everything inside a transaction because of the external factor.

In my opinion one should expect the `IDs()` call to work the same way independent of being called from Delete vs. DeleteOne. This way hooks can be better reused and code duplication can be avoided.

mwoelk: Additionally it makes the function behave as it is commented: 
"// IDs queries the database and returns the entity ids **that match the mutation's predicate**."

For a single ID it should be possible to not match the predicate of the mutation resulting in an empty list of IDs.

a8m: Hey @mwoelk, I didn't miss the previous comment, and there's no need to "convince" me that the new logic is correct :). I just need to test some large codebases to ensure this change doesn't break any existing behavior. I will update you soon.

a8m: Can you please add test cases for this? You can add them under the `integration/privacy` package.

mwoelk: Sorry. Have been away a few days. Will add some tomorrow.

mwoelk: Done. Added a new public "Note" entity that can be set to readonly during creation. A filter is then applied during mutation to filter out the ones set to readonly. Added some tests to `privacy_test.go` for UpdateOne, DeleteOne for both filtered (readonly) and unfiltered (not-readonly) notes and test for Update and Delete many. The tests inject a hook that compares present IDs to expected ones.

mwoelk: @a8m let me know if I should add more test cases or if I can help getting this merged in any way

mwoelk:
@a8m any update on this?

a8m:
> @a8m any update on this?

Mutation should not be tested inside hooks. Instead, create a builder and then call `Mutation()` - The assertions should apply on this object. If this requires too much time, just clean the tests and one of the team will take it. 

You can add a simple case to this function: https://github.com/ent/ent/blob/master/entc/integration/integration_test.go#L2122

mwoelk:
Saw the failed CI report. Forgot to run generate for the examples directory as well. Should be fixed now :)

mwoelk:
Hey @a8m, any chance we can get this merged. 
Just synced with master again and ran all generate + mod tidy as well as tests. Hopefully should pass CI now :)

PR #3890, opened 2024-01-05
Allow users to provide their own SchemaSpec to entc

nprintz:
@a8m might remember some of this from his time at Meta/Facebook, but our build system differs from the common OSS model in some ways. In our environment, relying on `entc/load`'s functionality of dynamically loading the user schema package to find the exported structs that satisfy the `entgo.io/ent.Inferace` interface is problematic for two reasons:

* Differences in the build and runtime environments can lead to issues. For example if a different golang version is used, there might be problems doing the `golang.org/x/tools/go/packages` dynamic loading of the schema package caused by syntax or other language differences between the two versions.
* In an environment where the package namespace is not entirely compatible with the OSS model, `entc/load` will be unable to find exported struct names that match the expected due to naming differences, and the schema cannot be loaded. For example if the build environment means the schema package is defined as `some/prefix/github.com/nprintz/ent/schema`, `entc/load` would still be looking for `github.com/nprintz/ent/schema`. 

Allowing the user to provide their own `SchemaSpec` to `entc` via the `Config` struct helps with both of these issues. It helps create deterministic behavior for entc if the build environment is different than the runtime environment (since there is no dependence on running `go` commands as part of Load. It also doesn't matter if the package namespace doesn't align entirely with the built naming, as the values in the provided `SchemaSpec` will be taken at face value. I tried to add a meaningful comment to the new `SchemaSpec` field on the `entc/gen` package's `Config` struct. 

I added an integration test to `entc/integration/schemaspec` that passes. It exemplifies this new option by having:
* `entc/integration/schemaspec/ent/schema/schema.go` defines a function that returns a `entc/load.SchemaSpec` making use of `entc/load`'s Marshal/Unmarshal functions to create the schemas from the `ent.Schema`.
* `entc/integration/schemaspec/ent/entc.go` calls the schema package's `NewLoadSpec` function and provides that to `ent/entc.Generate`.


a8m:
Thanks for the contribution, but why a new API is needed if you can pass loaded schemas to entc/gen: https://github.com/ent/ent/blob/master/entc/gen/graph.go#L108.

We actually use this (and also used at FB) to generate assets from non-Go formats.

nprintz:
To make sure I understand, you mean skip calling `entc.Generate` entirely, and create our own `entc/gen.Graph` by hand (using the `load.Schema` we have already crafted) and call `Graph.Gen()` directly?

a8m:
Exactly. There is nothing interesting or complex in the entc package besides some options to both gen and load packages.

nprintz:
Verified that works, thanks for the tip! I didn't get that far into the internals of the Graph struct, and was trying to get it working via using entc still.

PR #3943, opened 2024-02-09
feat: do not add a group by clause to O2O-edges for order by neighbor…

michaelcaulley:
… terms

this prevents issues with databases that require all selected terms be included in the group by clause. the group by clause is not required if the relationship is O2O, because only one row will be returned if we join on fks.

### DIFF
--- a/dialect/sql/sqlgraph/graph.go
+++ b/dialect/sql/sqlgraph/graph.go
@@ -512,8 +512,10 @@ func OrderByNeighborTerms(q *sql.Selector, s *Step, opts ...sql.OrderTerm) {
 	case s.ToEdgeOwner():
 		toT := build.Table(s.Edge.Table).Schema(s.Edge.Schema)
 		join = build.Select(toT.C(s.Edge.Columns[0])).
-			From(toT).
-			GroupBy(toT.C(s.Edge.Columns[0]))
+			From(toT)
+		if s.Edge.Rel != O2O {
+			join = join.GroupBy(toT.C(s.Edge.Columns[0]))
+		}
 		selectTerms(join, opts)
 		q.LeftJoin(join).
 			On(q.C(s.From.Column), join.C(s.Edge.Columns[0]))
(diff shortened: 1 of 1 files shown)


michaelcaulley:
@giautm can you help review this change?

dialect/sql/sqlgraph/graph.go
```diff
@@ -512,8 +512,10 @@ func OrderByNeighborTerms(q *sql.Selector, s *Step, opts ...sql.OrderTerm) {
 	case s.ToEdgeOwner():
 		toT := build.Table(s.Edge.Table).Schema(s.Edge.Schema)
 		join = build.Select(toT.C(s.Edge.Columns[0])).
-			From(toT).
-			GroupBy(toT.C(s.Edge.Columns[0]))
+			From(toT)
+		if s.Edge.Rel != O2O {
+			join = join.GroupBy(toT.C(s.Edge.Columns[0]))
```
a8m, line 517: That’s true if the edge is unique, not only o2o. 

Also, please add/update tests.

michaelcaulley: @a8m Do you have ideas for tests? It looks like there is already a test that covers ordering by a unique edge field, but all the tests use sqlite, and therefore this specific test passed both before and after this change. This change is only to generate valid sql in other databases too, such as PostgreSQL.

TargetLiu:
This modification is extremely helpful. In PostgreSQL, attempting to use ORDER BY on an edge results in an error: pq: column 'xxx' must appear in the GROUP BY clause or be used in an aggregate function. I believe this is a bug that requires urgent resolution. Thank you very much.

akushsky:
I'm waiting for that fix too, have the same issue when trying to add ORDER BY edge field to my schema.

Edge to:
```
edge.To("stats", StoredPackageVersionStats.Type).
			Annotations(entgql.OrderField("STATS_DOWNLOAD_COUNT")).
			Unique(),
```

Edge from:
```
func (StoredPackageVersionStats) Fields() []ent.Field {
	return []ent.Field{
		field.Int("id").Unique(),
		field.Int("version_id"),
		field.Int("download_count").Annotations(entgql.OrderField("DOWNLOAD_COUNT")),
	}
}

func (StoredPackageVersionStats) Edges() []ent.Edge {
	return []ent.Edge{
		edge.
			From("version", StoredPackageVersion.Type).
			Field("version_id").
			Ref("stats").Required().Unique(),
	}
}
```

Bad SQL in result:
```
SELECT "md_versions"."id", "md_versions"."name", "t1"."stats_download_count" FROM "md_versions" LEFT JOIN (SELECT "md_ver_stats"."version_id", "md_ver_stats"."download_count" AS "stats_download_count" FROM "md_ver_stats" GROUP BY "md_ver_stats"."version_id") AS "t1" ON "md_versions"."id" = "t1"."version_id" ORDER BY "t1"."stats_download_count" NULLS FIRST, "md_versions"."id" LIMIT 51
```

dlukt:
Can this be merged?
Hitting the same wall, will work around client side for now

PR #3945, opened 2024-02-11
dialect/sql/sqlgraph: fix CreateBulk when Fields in BatchCreateSpec is empty

### LINKED ISSUE #3944: M2M CreateBulk fails when entity has only ID field
Hi, thanks for an awesome library!

I want to create entities with M2M relationship by CreateBulk().

The problem is that CreateBulk() throws an error if entity has only ID field.
`failed creating group: add m2m edge for table group_users: Error 1048 (23000): Column 'group_id' cannot be null`

## Current Behavior 😯

We cannot create M2M entities through CreateBulk if entity has only ID field.
(Or, is this the expected behavior?)

## Expected Behavior 🤔

CreateBulk succeeds even if entity has only ID fied.

## Steps to Reproduce 🕹

We first define two simple entities with M2M relationship like below.
```
// ent/schema/group.go
func (Group) Fields() []ent.Field {  // This has only ID field if name field is omitted
	return []ent.Field{field.String("name").Optional()}
}
func (Group) Edges() []ent.Edge {
	return []ent.Edge{
		edge.To("users", User.Type),
	}
}

// ent/schema/user.go
	return []ent.Field{field.String("name").Default("unknown")}
}
	return []ent.Edge{
		edge.From("groups", Group.Type).
			Ref("users"),
	}
}
```

This is my create function with CreateBulk().

```
func Create(ctx context.Context, client *ent.Client) error {
	u, err := client.User.CreateBulk( // this is OK
			client.User.Create().SetName("user1"),
			client.User.Create().SetName("user2"),
			client.User.Create().SetName("user3"),
		).Save(ctx)
	if err != nil {
		return fmt.Errorf("failed creating user: %w", err)
	}
	log.Println("user was created: ", u)

	g, err := client.Group.CreateBulk( // name field is omitted
		client.Group.Create().AddUsers(u[0], u[2]),
		client.Group.Create().AddUsers(u[1]),
		client.Group.Create().AddUsers(u[1], u[0]),
	).Save(ctx) // We got an error here

	if err != nil {
		return fmt.Errorf("failed creating group: %w", err)
	}
	log.Println("group was created: ", g)
	return nil
}
```

Then, we got the following error.
```
driver.Exec: query=INSERT INTO `users` (`name`) VALUES (?), (?), (?) args=[user1 user2 user3]
user was created:  [User(id=1, name=user1) User(id=2, name=user2) User(id=3, name=user3)]
driver.Tx(3ab07907-718f-43b9-92ae-2585c8680838): started
Tx(3ab07907-718f-43b9-92ae-2585c8680838).Exec: query=INSERT INTO `groups` VALUES () args=[]
Tx(3ab07907-718f-43b9-92ae-2585c8680838).Exec: query=INSERT INTO `group_users` (`group_id`, `user_id`) VALUES (?, ?), (?, ?), (NULL, ?), (NULL, ?), (NULL, ?) ON DUPLICATE KEY UPDATE `group_id` = `group_users`.`group_id`, `user_id` = `group_users`.`user_id` args=[1 1 1 3 2 1 2]
Tx(3ab07907-718f-43b9-92ae-2585c8680838): rollbacked
failed creating group: add m2m edge for table group_users: Error 1048 (23000): Column 'group_id' cannot be null
exit status 1
```

It seems to me that ent generates the wrong query to insert group records.
```
query=INSERT INTO `groups` VALUES () args=[]
```
I think the expected query is this (because we want to insert 3 records). :
```
query=INSERT INTO `groups` VALUES (), (), () args=[] // for mysql
or
query=INSERT INTO `groups` (`id`) VALUES (NULL), (NULL), (NULL) args=[] // no dialect, maybe ?
```

The current workaround is to use three Create().Save(ctx) funcs instead of CreateBulk().

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.21.2  |
| Ent         | 0.13.0   |
| Database    | MySQL   |
| Driver      | https://github.com/go-sql-driver/mysql        |

t-coron:
Fixed #3944  Issue in CreateBulk() for m2m edges

CraeteBulk() always generates wrong query when Fields in batchCreateSpec.Nodes is empty 
```
query=INSERT INTO `groups` VALUES () args=[]
````
This query is generated by:
https://github.com/ent/ent/blob/d4560fae6215a2d039a8cab219a667b7a0e972e9/dialect/sql/sqlgraph/graph.go#L1508

If defaults is true and len(sorted) (i.e. len(columns)) is zero, insert builder exec writeDefault(), 
https://github.com/ent/ent/blob/d4560fae6215a2d039a8cab219a667b7a0e972e9/dialect/sql/builder.go#L963-L965
https://github.com/ent/ent/blob/d4560fae6215a2d039a8cab219a667b7a0e972e9/dialect/sql/builder.go#L982-L989

With this PR, we  try to avoid execution of writeDefault.
If we want to insert 3 records through  CreateBulk(), the query is:
````
query=INSERT INTO `groups` (`id`) VALUES (NULL), (NULL), (NULL) args=[]
````

### DIFF
--- a/dialect/sql/sqlgraph/graph.go
+++ b/dialect/sql/sqlgraph/graph.go
@@ -1477,7 +1477,7 @@ func (c *batchCreator) nodes(ctx context.Context, drv dialect.Driver) error {
 			return fmt.Errorf("more than 1 table for batch insert: %q != %q", node.Table, c.Nodes[i-1].Table)
 		}
 		values[i] = make(map[string]driver.Value)
-		if node.ID != nil && node.ID.Value != nil {
+		if node.ID != nil && (node.ID.Value != nil || len(node.Fields) == 0) {
 			columns[node.ID.Column] = struct{}{}
 			values[i][node.ID.Column] = node.ID.Value
 		}
--- a/dialect/sql/sqlgraph/graph_test.go
+++ b/dialect/sql/sqlgraph/graph_test.go
@@ -1595,6 +1595,52 @@ func TestBatchCreate(t *testing.T) {
 			},
 		},
 		{
+			name: "empty fields",
+			spec: &BatchCreateSpec{
+				Nodes: []*CreateSpec{
+					{
+						Table:  "users",
+						ID:     &FieldSpec{Column: "id", Type: field.TypeInt},
+						Fields: []*FieldSpec{},
+						Edges: []*EdgeSpec{
+							{Rel: M2M, Inverse: true, Table: "group_users", Columns: []string{"group_id", "user_id"}, Target: &EdgeTarget{Nodes: []driver.Value{2}, IDSpec: &FieldSpec{Column: "id"}}},
+							{Rel: M2M, Table: "user_products", Columns: []string{"user_id", "product_id"}, Target: &EdgeTarget{Nodes: []driver.Value{2}, IDSpec: &FieldSpec{Column: "id"}}},
+							{Rel: M2M, Table: "user_friends", Bidi: true, Columns: []string{"user_id", "friend_id"}, Target: &EdgeTarget{IDSpec: &FieldSpec{Column: "id", Type: field.TypeInt}, Nodes: []driver.Value{2}}},
+						},
+					},
+					{
+						Table:  "users",
+						ID:     &FieldSpec{Column: "id", Type: field.TypeInt},
+						Fields: []*FieldSpec{},
+						Edges: []*EdgeSpec{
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by t-coron** [comment]
Umm, this might be better.
```
// if node.ID != nil && (node.ID.Value != nil || len(node.Fields) == 0) {
if node.ID != nil {
```

PR #3953, opened 2024-02-23
entc/gen: set unique backref when their inverse is loaded

a8m:
In the context of entgql/gqlgen, it is critical to access the inverse if it has been loaded at a higher level of the query, as field resolvers do not have the full context.


masseelch (approved):
LGTM

PR #3977, opened 2024-03-15
add support for libSQL dialect

golanglemonade:
We are using `atlas` for migrations, which has support for `libsql`, a fork of SQLite, however, today you cannot use versioned migrations because it uses ent, which doesn't support the `libsql` dialect. This PR adds the libSQL dialect, but uses all the same code-paths as SQLite. 

Since libSQL is a fork of SQLite, I didn't find it necessary to create an entirely different `sql/schema` type, but happy to hear other opinions.

Relates to https://github.com/ariga/atlas/issues/2338


PietervdWerk:
I would love this feature ❤️

asilvadesigns:
👀

modeledge:
I would love this feature - any update on this?

kitsoNamane:
> I would love this feature - any update on this?

I would love the this feature as well but it doesn't seem like it's a priority right now.

Issue #3168, opened 2022-12-12
Support for generated columns


carlos0202:
I'm wonder if `entgo` supports generated columns as stated [here](https://www.postgresql.org/docs/current/ddl-generated-columns.html). If so, is there any place on the documentation to configure a column to be generated always?

Thanks in advance for the help!!

a8m:
[Atlas](https://atlasgo.io/) supports it, but Ent does not support it yet. It should be added to Ent, and suggestions for API are welcome.

Ambroos:
I think sticking close to Atlas API-wise would be nice. A `GeneratedAs` on any field type that just takes the "as" SQL string similar to Atlas should suffice. And then not generate any fields on any mutations.

With hooks / diff hooks and some cheating I've been able to use generated columns in one of my ents with no issues so far, although it really is a bit janky:

In my schema, field:

```golang
		field.Enum("status").
			Values("draft", "published", "archived").
			Immutable().
			Optional() // to skip Ent's validation on create
```

Hook:

```golang
		hook.On(
			// Forbid setting status, since it is generated
			func(next ent.Mutator) ent.Mutator {
				return hook.MyEntFunc(func(ctx context.Context, m *gen.MyEntMutation) (ent.Value, error) {
					_, exists := m.Status()
					if exists {
						return nil, fmt.Errorf("status is generated, cannot be updated manually")
					}
					return next.Mutate(ctx, m)
				})
			},
			ent.OpCreate|ent.OpUpdate|ent.OpUpdateOne,
		),
```

And then in my migrate/main.go, which generates an Atlas versioned migration:

```golang
    ...
    schema.WithDiffHook(statusDiffHook),
    ...

func statusDiffHook(next schema.Differ) schema.Differ {
	return schema.DiffFunc(func(current, desired *atlasSchema.Schema) ([]atlasSchema.Change, error) {
		for _, table := range desired.Tables {
			if table.Name == "my_ent" {
				for _, column := range table.Columns {
					if column.Name == "status" {
						column.SetNull(false)
						column.SetGeneratedExpr(&atlasSchema.GeneratedExpr{
							Expr: `CASE WHEN "archived_at" IS NOT NULL THEN 'archived' WHEN "published_at" IS NOT NULL THEN 'published' ELSE 'draft' END`,
						})
					}
				}
			}
		}

		changes, err := next.Diff(current, desired)
		if err != nil {
			return nil, err
		}

		return changes, nil
	})
}
```

[...]

PR #3984, opened 2024-03-21
(Postgres) Add Annotation for Generated Expressions

ivanvanderbyl:
This PR resolves https://github.com/ent/ent/issues/3168 to add support for generated columns as needed when using `tsvector` or similar extensions in Postgres.

**Usage:**

```go
field.String("text_searchable_content").
	SchemaType(map[string]string{dialect.Postgres: "tsvector"}).
	Annotations(entsql.GeneratedAs("to_tsvector(...)")).
	Optional()
```

**Follow up work:**

Obviously for something like `tsvector` where you never read or write the column, this is suboptimal and a follow up PR should add a new field type for this use case, such as `field.Generated("name").As("to_expression()")` with the following properties:
- Does not add Setters or Getters for the field
- Field is immutable
- Field is optional for creates
- Generates a `model.FieldName` var


ivanvanderbyl:
👋🏻  @a8m hey Ariel, I'd love to get your input on this, and any changes I can make to get it merged

PR #3990, opened 2024-03-28
doc/md: fix some broken links caused by trailing slash

luantranminh:
When searching or reloading certain document pages, the website automatically adds a trailing slash, which causes some URL paths to break.

### Steps to Reproduce
1. Access the [Ent docs](https://entgo.io/docs/getting-started).
2. Search for `Custom Table Schema`
3. Click on the `multi-schema doc`.

### Solution: 
Using **file path** instead of **URL path** for link
ref: https://docusaurus.io/docs/markdown-features/links

### DIFF
--- a/doc/md/code-gen.md
+++ b/doc/md/code-gen.md
@@ -26,7 +26,7 @@ is to have an `ent` directory under the root directory of the project.
 
 ## Generate Assets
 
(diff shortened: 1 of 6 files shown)


a8m (approved):
Thanks 🙏🏻

PR #4014, opened 2024-04-14
entsql: added OnUpdate annotation

### LINKED ISSUE #3776: [Feature Request] entsql.OnUpdate Annotation
We have [Foreign Keys Configuration](https://entgo.io/docs/schema-annotations/#foreign-keys-configuration) which makes us able to configure `.OnDelete`, but sometimes we also need to configure `.OnUpdate` which i cannot find it configurable in ent!

## Summary 💡

Just how `.OnDelete` works with annotations, but for OnUpdate! e.g: `.Annotations(entsql.OnUpdate(entsql.Restrict))`

BrandSnob:
adds `OnUpdate` annotation,
So we can figure out the times we need to add `ON UPDATE CASCADE` to prevent foreign key check errors on updating a table's row.
Has been really missing!

closes #3776


gnc-git:
Hey @a8m, this PR has been pending since May. It's a simple but crucial change that adds an OnUpdate annotation to handle cascading updates more effectively. Could you please take a look and push this through when you have a moment? It would really help with maintaining data integrity in projects using this feature. 
Thanks!

a8m:
IDs in Ent are immutable, that’s why this is not supported. 

@BrandSnob, can you please elaborate on the use case?

PR #4017, opened 2024-04-15
Bump version of golang-sync, golang-mod, and golang-tools

philip-peterson:
I am working on a set of patches in other projects that may eventually help Ent avoid some performance bottlenecks. (Reference issue e.g.: https://github.com/golang/go/issues/66802)

If/once those changes are merged, I will hope to bump the version of golang-tools to use this new functionality that can avoid performance issues. So, to be proactive, I have bumped the version here since it seems desirable anyway and for the aforementioned reason.

Thanks,
Phil


giautm:
Close as duplicated with https://github.com/ent/ent/pull/4209

PR #4041, opened 2024-04-30
doc/website: Ent focused announcementBar

andheiberg:
Reverting the announcementBar for https://entgo.io/docs to the generic Ent focused message from prior to https://github.com/ent/ent/pull/3792

I can understand why this messaging change was made on Oct 17, 2023 in the wake of the attacks and share my condemnation and condolences.

It's your project and you should do with it as you please. I harbour no ill intend. But I would submit that this political message can be removed >6 months on.

### DIFF
--- a/doc/website/docusaurus.config.js
+++ b/doc/website/docusaurus.config.js
@@ -253,12 +253,11 @@ const config = {
       `,
     },
     announcementBar: {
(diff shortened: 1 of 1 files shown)


a8m:
Thanks for the suggestion, but I don't have plans to change it.

PR #4056, opened 2024-05-16
Add error handling for nil builders when calling Exec for Upsert

pthm:
A small contribution to add a check to the Exec call as part of the Upsert functionality.

Occasionally I would see cases where developers mistakenly construct a slice of builders usually by instantiating a slice with a set length and then not setting a builder for each index resulting in nils being passed to the CreateBulk method.

This would then result in a panic from the Exec call when it tries to call len on the b.conflict property.

This change checks if b is nil and returns an error to hopefully help people more easily debug this type of issue.


jvgreenaway:
this would be so handy

PR #4120, opened 2024-07-01
entc/gen: fix loss of int64 precision during JSON type decoding

### LINKED ISSUE #4119: loss of int64 precision during JSON type decoding
image

Data type is JSON, the struct contains int64 type, and precision is lost during deserialization.

eiixy:
fix #4119


entc/gen/template/dialect/sql/decode.tmpl
```diff
@@ -129,7 +129,9 @@ func ({{ $receiver }} *{{ $.Name }}) {{ $.ValueName }}(name string) (ent.Value,
 		if value, ok := values[{{ $i }}].(*{{ $f.ScanType }}); !ok {
 			return fmt.Errorf("unexpected type %T for field {{ $f.Name }}", values[{{ $i }}])
 		} else if value != nil && len(*value) > 0 {
-			if err := json.Unmarshal(*value, &{{ $ret }}.{{ $field }}); err != nil {
+			decoder := json.NewDecoder(bytes.NewBuffer(*value))
+			decoder.UseNumber()
```
giautm, line 133: @a8m, wdyt? I also prefer to use Number for this stuff.

entc/gen/template/dialect/sql/decode.tmpl
```diff
@@ -129,7 +129,9 @@ func ({{ $receiver }} *{{ $.Name }}) {{ $.ValueName }}(name string) (ent.Value,
 		if value, ok := values[{{ $i }}].(*{{ $f.ScanType }}); !ok {
 			return fmt.Errorf("unexpected type %T for field {{ $f.Name }}", values[{{ $i }}])
 		} else if value != nil && len(*value) > 0 {
-			if err := json.Unmarshal(*value, &{{ $ret }}.{{ $field }}); err != nil {
+			decoder := json.NewDecoder(bytes.NewBuffer(*value))
```
giautm, line 132:
```suggestion
			decoder := json.NewDecoder(bytes.NewReader(*value))
```

eiixy: ok

entc/gen/template/dialect/sql/decode.tmpl
a8m: By default, we use the Go defaults. However, users can implement the UnmarshalJSON method (on their ent fields) to control the unmarshaling. Another option is a custom ValueScanner. 

If we merge this, it will introduce breaking changes for those who use untyped struct fields which I prefer to avoid.

Issue #4159, opened 2024-07-29
WithQueryBuilder removes custom AppendSelectExprAs in M2M queries


gebes:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

When querying an entity, you can use `AppendSelectExprAs` in the `.Where` function with the selector. This works great.
However, when querying an entity and using `AppendSelectExprAs` in a `With` builder for a M2M edge, the select expressions get removed by the {{entity}}_query.go, load{{SubRow}} function which selects the join columns first and then only the regular selected columns, without the custom expressions.

```go
	query.Where(func(s *sql.Selector) {
		joinT := sql.Table(...)
		s.Join(joinT).On(s.C(...FieldID), joinT.C(...PrimaryKey[1]))
		s.Where(sql.InValues(joinT.C(...PrimaryKey[0]), edgeIDs...))
		columns := s.SelectedColumns()
		s.Select(joinT.C(...PrimaryKey[0])) // THIS CODE CAUSES THE UNWANTED BEHAVIOR
		s.AppendSelect(columns...)
		s.SetDistinct(false)
	})
```

## Expected Behavior 🤔

As with O2M, custom select expressions should be selected for M2M too when queried within a WithBuilder.

PR #4169, opened 2024-08-10
Prepend fix for #4159

gebes:
This PR fixes an issue with the M2M load function where additional select expressions added in the `With{Node}` query functions were being lost. The problem occurred because the selected columns were copied, the join column was selected, and then the selected columns were added back—but the expressions aren't included in `SelectedColumns`

By switching to prepending instead of appending + copying, we keep the expressions, so everything works as expected. This change has been tested in several production projects, and all tests are passing.

I have also added comments to explain the new prepend functions.

Thanks for taking a look!

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -2254,6 +2254,16 @@ func (s *Selector) SelectDistinct(columns ...string) *Selector {
 	return s.Select(columns...).Distinct()
 }
 
+// SelectExpr changes the columns selection of the SELECT statement
+// with custom list of expressions.
+func (s *Selector) SelectExpr(exprs ...Querier) *Selector {
+	s.selection = make([]selection, len(exprs))
+	for i := range exprs {
+		s.selection[i] = selection{x: exprs[i]}
+	}
+	return s
+}
+
 // AppendSelect appends additional columns to the SELECT statement.
 func (s *Selector) AppendSelect(columns ...string) *Selector {
 	for i := range columns {
@@ -2265,16 +2275,6 @@ func (s *Selector) AppendSelect(columns ...string) *Selector {
 // AppendSelectAs appends additional column to the SELECT statement with the given alias.
 func (s *Selector) AppendSelectAs(column, as string) *Selector {
 	s.selection = append(s.selection, selection{c: column, as: as})
-	return s
-}
-
-// SelectExpr changes the columns selection of the SELECT statement
-// with custom list of expressions.
-func (s *Selector) SelectExpr(exprs ...Querier) *Selector {
-	s.selection = make([]selection, len(exprs))
-	for i := range exprs {
-		s.selection[i] = selection{x: exprs[i]}
-	}
 	return s
 }
 
@@ -2298,6 +2298,40 @@ func (s *Selector) AppendSelectExprAs(expr Querier, as string) *Selector {
 		x:  x,
 		as: as,
 	})
+	return s
+}
+
+// PrependSelect prepends additional columns to the SELECT statement.
+func (s *Selector) PrependSelect(columns ...string) *Selector {
+	for i := range columns {
+		s.selection = append([]selection{{c: columns[i]}}, s.selection...)
+	}
+	return s
+}
+
+// PrependSelectAs prepends an additional column to the SELECT statement with the given alias.
+func (s *Selector) PrependSelectAs(column, as string) *Selector {
+	s.selection = append([]selection{{c: column, as: as}}, s.selection...)
+	return s
+}
+
+// PrependSelectExpr prepends additional expressions to the SELECT statement.
+func (s *Selector) PrependSelectExpr(exprs ...Querier) *Selector {
+	for i := range exprs {
+		s.selection = append([]selection{{x: exprs[i]}}, s.selection...)
+	}
+	return s
+}
+
+// PrependSelectExprAs prepends additional expressions to the SELECT statement with the given name.
+func (s *Selector) PrependSelectExprAs(expr Querier, as string) *Selector {
+	x := expr
+	if _, ok := expr.(*raw); !ok {
+		x = ExprFunc(func(b *Builder) {
+			b.S("(").Join(expr).S(")")
+		})
+	}
+	s.selection = append([]selection{{x: x, as: as}}, s.selection...)
 	return s
 }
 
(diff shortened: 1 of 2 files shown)


dialect/sql/builder.go
```diff
@@ -2254,6 +2254,16 @@ func (s *Selector) SelectDistinct(columns ...string) *Selector {
 	return s.Select(columns...).Distinct()
 }
 
+// SelectExpr changes the columns selection of the SELECT statement
+// with custom list of expressions.
+func (s *Selector) SelectExpr(exprs ...Querier) *Selector {
```
a8m, line 2259: No need to change its position.

dialect/sql/builder.go
```diff
@@ -2301,6 +2301,40 @@ func (s *Selector) AppendSelectExprAs(expr Querier, as string) *Selector {
 	return s
 }
 
+// PrependSelect prepends additional columns to the SELECT statement.
+func (s *Selector) PrependSelect(columns ...string) *Selector {
```
a8m, line 2305: Keep only those you use in the codegen. No need to add unused methods.

entc/gen/template/dialect/sql/query.tmpl
```diff
@@ -135,9 +135,7 @@ func ({{ $receiver }} *{{ $builder }}) sqlAll(ctx context.Context, hooks ...quer
 				{{- $fk1idx := 1 }}{{- $fk2idx := 0 }}{{ if $e.IsInverse }}{{ $fk1idx = 0 }}{{ $fk2idx = 1 }}{{ end }}
 				s.Join(joinT).On(s.C({{ $edgeid }}), joinT.C({{ $.Package }}.{{ $e.PKConstant }}[{{ $fk1idx }}]))
 				s.Where(sql.InValues(joinT.C({{ $.Package }}.{{ $e.PKConstant }}[{{ $fk2idx }}]), edgeIDs...))
-				columns := s.SelectedColumns()
-				s.Select(joinT.C({{ $.Package }}.{{ $e.PKConstant }}[{{ $fk2idx }}]))
-				s.AppendSelect(columns...)
+				s.PrependSelect(joinT.C({{ $.Package }}.{{ $e.PKConstant }}[{{ $fk2idx }}]))
 				s.SetDistinct(false)
```
a8m, line 139: If there's a bug in the codegen, use the `entc/integration/integration_test.go` to write a test case that without this fix, the test fails. This will give us more context on the problem (if there's one) it's solved, and ensure we keep BC between versions.

dialect/sql/builder.go
gebes: I undid it

dialect/sql/builder.go
gebes: Done

entc/gen/template/dialect/sql/query.tmpl
gebes: Added a integration test

entc/integration/integration_test.go
```diff
@@ -2870,6 +2871,36 @@ func OrderByEdgeTerms(t *testing.T, client *ent.Client) {
...
+	client.User.Query().OnlyX(ctx)
+
+	inf := client.GroupInfo.Create().SetDesc("desc").SaveX(ctx)
+	client.Group.Create().SetName("Github").SetExpire(time.Now()).AddUsers(usr).SetInfo(inf).SaveX(ctx)
+	client.Group.Query().OnlyX(ctx)
+
+	const key = "computed_value"
+	users := client.User.Query().WithGroups(func(query *ent.GroupQuery) {
+		query.Where(func(selector *sql.Selector) {
+			selector.AppendSelectExprAs(sql.Expr("min(3, 5)"), key)
```
gebes, line 2888: The main issue was that expressions (which contained computed fields, function calls, etc) where discarded and hence not selected @a8m

entc/integration/integration_test.go
```diff
@@ -2870,6 +2871,36 @@ func OrderByEdgeTerms(t *testing.T, client *ent.Client) {
...
+		})
+	}).AllX(ctx)
```
a8m, line 2890: If you need to modify the query selection, use the [query modifiers](https://entgo.io/docs/feature-flags/#select-and-scan-dynamic-values). e.g.,
```suggestion
		query.Modify(func(s *sql.Selector) {
			s.AppendSelectExprAs(sql.Expr("min(3, 5)"), key)
		})
	}).AllX(ctx)
```
This should solve your issue. Please, try update the test and see if this works without changing the codegen.

a8m (commented):
See my suggested change and let me know if this solves your issue.

gebes:
@a8m It resolved the issue, thanks so much!

I would still suggest merging this PR, though, as it makes the behavior consistent. O2O and O2M don't have this issue, but M2M does, so this change would be a nice way to align everything and as a bonus it reduces unnecessary copying

a8m:
> @a8m It resolved the issue, thanks so much!
> 
> I would still suggest merging this PR, though, as it makes the behavior consistent. O2O and O2M don't have this issue, but M2M does, so this change would be a nice way to align everything and as a bonus it reduces unnecessary copying

Thanks for the suggestion, but I prefer to avoid incorrect usage of the API. If we "officially" support it, that means we need to support it for all methods, not just for custom selections. Instead, users can use the right API.

gebes:
> > @a8m It resolved the issue, thanks so much!
> > I would still suggest merging this PR, though, as it makes the behavior consistent. O2O and O2M don't have this issue, but M2M does, so this change would be a nice way to align everything and as a bonus it reduces unnecessary copying
> 
> Thanks for the suggestion, but I prefer to avoid incorrect usage of the API. If we "officially" support it, that means we need to support it for all methods, not just for custom selections. Instead, users can use the right API.

I understand your point about the correct API usage, and I didn't mean officially support this. I am aware that internal APIs can change frequently. My suggestion was just focused on achieving the same result with cleaner code and reduced allocations. So, I would see this PR as a minor code improvement rather than an official bug fix if my mistake was incorrect API usage.

I would be happy if you could merge this PR to appreciate the effort

I'm also curious about the project's guidelines on clean code and optimizing performance by reducing allocations/reducing code. There are some code optimizations that could be made to the generated code, and I'm wondering about the process for making changes. If it's challenging to make such a small adjustment, how do you see the project progressing towards the 1.0 release? Of course, ent shouldn't move fast and break things, but steady quality improvements are still important and better than none imo, as this doesn't break anything.

Issue #4130, opened 2024-07-09
M2M Through Bulk Update reuses defaults for every new entity (even id)


gebes:
When having a resolved M2M Table for a `.Through` Edge, e.g. to add an `added_at` date then all the defaults are being reused when you insert multiple features using `Meal.Update().AddFeatures(features...).Exec(ctx)`.
Thats not an issue for the added_at date, BUT for uuid.New that is really bad.

Current workaround:
Creating all M2M using MealFeature.CreateBulk

Example of the fields used for the Through edge
```go
func (MealFeature) Fields() []ent.Field {
	return []ent.Field{
		field.UUID("id", uuid.UUID{}).Default(uuid.New),

		field.Time("added_at").Default(time.Now).Immutable().Nillable(),

		field.UUID("meal_id", uuid.UUID{}).Default(uuid.New),
		field.Int("feature_id"),
	}
}
```

Version: v0.13.1

- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

PR #4170, opened 2024-08-10
M2M Bulk through add insertion bug for #4130

gebes:
This PR fixes issue #4130, which involved adding multiple M2M rows to a table using `Through` tables. Previously, the same default values were reused for every row, which was fine for fields like `created_at` with a default date. However, this caused issues when using a DefaultID generator, leading to potential duplicate primary key errors.

With this update, the affected `Through` edges are modified to ensure the "default" builder is called for each new row, preventing these errors.

This change has been tested in several production projects, and all tests are passing.

Thanks for taking a look!

### DIFF
--- a/entc/gen/template/dialect/sql/update.tmpl
+++ b/entc/gen/template/dialect/sql/update.tmpl
@@ -150,10 +150,14 @@ func ({{ $receiver }} *{{ $builder }}) sqlSave(ctx context.Context) ({{ $ret }} 
 			}
 		{{- end }}
 		if nodes := {{ $mutation }}.{{ $e.StructField }}IDs(); len(nodes) > 0 {
-			{{- with extend $ "Edge" $e "Nodes" true "Zero" $zero }}
+	    {{- /* Recompute defaults for every through edge */}}
+		  {{- if $e.Through }}for _, node := range nodes { {{end}}
+		  {{- $hasThrough := ne $e.Through nil}}
+			{{- with extend $ "Edge" $e "Nodes" (not $hasThrough) "Zero" $zero "Through" ($hasThrough) }}
 				{{ template "dialect/sql/defedge" . }}
 			{{- end }}
 			_spec.Edges.Add = append(_spec.Edges.Add, edge)
+		  {{- if $e.Through }}}{{end}}
 		}
 	{{- end }}
 	{{- /* Allow mutating the sqlgraph.UpdateSpec by ent extensions or user templates.*/}}
@@ -207,6 +211,9 @@ func ({{ $receiver }} *{{ $builder }}) sqlSave(ctx context.Context) ({{ $ret }} 
 		for _, k := range nodes {
 			edge.Target.Nodes = append(edge.Target.Nodes, k)
 		}
+	{{- end }}
+	{{- with $.Scope.Through }}
+		edge.Target.Nodes = append(edge.Target.Nodes, node)
 	{{- end }}
 	{{- with $e.Through }}
 		{{- if .HasDefault }}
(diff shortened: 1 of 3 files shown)


entc/integration/integration_test.go
```diff
@@ -2870,6 +2871,31 @@ func OrderByEdgeTerms(t *testing.T, client *ent.Client) {
...
+func InsertBulkM2MAdd(t *testing.T, client *ent.Client) {
+	require := require.New(t)
+	ctx := context.Background()
+
+	steve := client.Student.Create().SetName("Steve").SaveX(ctx)
+	bob := client.Student.Create().SetName("Bob").SaveX(ctx)
+	will := client.Student.Create().SetName("Will").SaveX(ctx)
+
+	math := client.Subject.Create().SetName("Math").
+		AddStudents(steve, bob).
```
gebes, line 2883: @a8m the main issue was, that when adding multiple students to a M2M edge which have a through table, and the through table has a default id, then the insertion fails because all of the ids were reused

gebes: Without this fix you will get for this test:

panic: ent: constraint failed: add m2m edge for table subject_students: UNIQUE constraint failed: subject_students.id [recovered]
	panic: ent: constraint failed: add m2m edge for table subject_students: UNIQUE constraint failed: subject_students.id

gebes:
@a8m requesting a review for this example. I created completely new tables, to avoid interfering with the already existing tests and to keep coverage up

entc/gen/template/dialect/sql/update.tmpl
```diff
@@ -150,10 +150,14 @@ func ({{ $receiver }} *{{ $builder }}) sqlSave(ctx context.Context) ({{ $ret }}
 			}
 		{{- end }}
 		if nodes := {{ $mutation }}.{{ $e.StructField }}IDs(); len(nodes) > 0 {
-			{{- with extend $ "Edge" $e "Nodes" true "Zero" $zero }}
+	    {{- /* Recompute defaults for every through edge */}}
+		  {{- if $e.Through }}for _, node := range nodes { {{end}}
+		  {{- $hasThrough := ne $e.Through nil}}
+			{{- with extend $ "Edge" $e "Nodes" (not $hasThrough) "Zero" $zero "Through" ($hasThrough) }}
 				{{ template "dialect/sql/defedge" . }}
```
a8m, line 157: I think we can change `sqlgraph` to accept a `Value` function instead of generating a spec per object - I'll check what's the simplest way to do it.

entc/integration/ent/schema/student.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2019-present Facebook Inc. All rights reserved.
+// This source code is licensed under the Apache 2.0 license found
+// in the LICENSE file in the root directory of this source tree.
+
+package schema
```
a8m, line 5: Edge-schemas are tested in `integration/edgeschema`. See this example test: https://github.com/ent/ent/blob/master/entc/integration/edgeschema/edgeschema_test.go#L184

I prefer to keep edge-schema tests on the same directory, and reuse existing schemas if possible.

entc/gen/template/dialect/sql/update.tmpl
gebes: Any update?

Southclaws:
I've just recently run into this and spend quite a bit of time debugging, didn't notice the issue and this PR. Would love to see this revisited!

rubensayshi:
hitting this issue as well, it's really nasty when you use UUIDs for IDs because it will violate any unique constraints!

PR #4207, opened 2024-09-07
feat(go.mod): use major version instead of minor version

jay-youngn:
Avoid environment and toolchain conflicts.


jay-youngn:
This change will interfere with the `1.23.1` version of the development environment, 
causing the local project to be fixed to `1.23.0`,
and forcing the toolchain to be added to `1.23.1`.

image

PR #4217, opened 2024-09-18
doc: add warning for Unwrap

bitcrshr:
Adds documentation to both generated code and the website noting that calling `someEntity.Unwrap()` on a non-transactional entity will cause a panic.

I wasn't 100% sure if the generated code needed to be included in this PR, but I left it out for now since it's easier to read. 

Please let me know if there's anything I'm missing here. Thank you!

### DIFF
--- a/doc/md/transactions.md
+++ b/doc/md/transactions.md
@@ -62,6 +62,11 @@ You must call `Unwrap()` if you are querying edges off of a created entity after
 
 The full example exists in [GitHub](https://github.com/ent/ent/tree/master/examples/traversal).
 
+:::warning Note
+Calling `Unwrap()` on a non-transactional entity (i.e., after a transaction has been committed or rolled back) will
+cause a panic.
+:::
+
 ## Transactional Client
 
 Sometimes, you have an existing code that already works with `*ent.Client`, and you want to change it (or wrap it)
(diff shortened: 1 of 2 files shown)


doc/md/transactions.md
```diff
@@ -62,6 +62,11 @@ You must call `Unwrap()` if you are querying edges off of a created entity after
 
 The full example exists in [GitHub](https://github.com/ent/ent/tree/master/examples/traversal).
 
+:::warning Note
+Calling `Unwrap()` on a non-transactional entity (i.e., after a transaction has been committed or rolled back) will
+cause a panic.
+:::
```
a8m, line 68: Move this to line 62, please.

entc/gen/template/ent.tmpl
```diff
@@ -111,6 +111,7 @@ func ({{ $receiver }} *{{ $.Name }}) Update() *{{ $.UpdateOneName }} {
 
 // Unwrap unwraps the {{ $.Name }} entity that was returned from a transaction after it was closed,
 // so that all future queries will be executed through the driver which created the transaction.
+// Unwrap will panic if {{ $.Name }} is not a transactional entity.
```
a8m, line 114: No need for this if we already added it to the doc.

a8m (commented):
Thanks for the contribution.

See minor comments.

bitcrshr:
Thanks for the feedback! Resolved those issues.

a8m (approved):
💯 

Thanks @bitcrshr

PR #4233, opened 2024-10-06
HasPrefixFold and HasSuffixFold

mohsinhijazee:
The `HasPrefix` and `HasSuffix` always generate `LIKE` operator which is case sensitive. In some of the operations, we want case insensitive operations instead and this PR adds methods to the sql package:

- `FieldHasPrefixFold`
- `FieldHasSuffixFold`

To support that, the underlying builder gets following methods:

- `HasPrefixFold`
- `HasPrefixFold` on `Predicate` type.
- `HasSuffixFold`
- `HasSuffixFold` on `Predicate` type.

FYI @a8m

### DIFF
--- a/dialect/sql/builder.go
+++ b/dialect/sql/builder.go
@@ -1716,13 +1716,28 @@ func (p *Predicate) escapedLike(col, left, right, word string) *Predicate {
 	})
 }
 
-func (p *Predicate) escapedILike(col, left, right, word string) *Predicate {
+// ContainsFold is a helper predicate that applies the LIKE predicate with case-folding.
+func (p *Predicate) escapedLikeFold(col, left, substr, right string) *Predicate {
 	return p.Append(func(b *Builder) {
-		w, escaped := escape(word)
-		b.Ident(col).WriteOp(OpILike)
-		b.Arg(left + w + right)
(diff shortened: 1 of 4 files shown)


dialect/sql/builder.go
```diff
@@ -3592,6 +3621,7 @@ const (
 	OpIn                // IN
 	OpNotIn             // NOT IN
 	OpLike              // LIKE
+	OpILike             // ILIKE
```
a8m, line 3624: I'd remove this operator as it's not supported by MySQL/MariaDB, and implement it like [ContainsFold](https://github.com/ent/ent/blob/master/dialect/sql/builder.go#L1803-L1824) predicate - the code can be shared between HasPrefixFold/HasSuffixFold/ContainsFold

mohsinhijazee: Okay. let me go through that and adjust accordingly.

mohsinhijazee: Adjusted - now the `ContainsFold` relies on a utility method called `escapedLikeFold` which is reused in `HasPrefixFold`, `HasSuffixFold` and `ContainsFold`

a8m (approved):
💯 

Thanks for the contribution @mohsinhijazee

PR #4244, opened 2024-10-21
dialect/sql: fix error formatting 

ronenlu:
before:
`Error: migration checks are available to logged-in users only. Use 'atlas login' to access this feature: %!w(<nil>)`

after:
`Error: migration checks are available to logged-in users only. Use 'atlas login' to access this feature`

### DIFF
--- a/dialect/sql/schema/atlas.go
+++ b/dialect/sql/schema/atlas.go
@@ -201,7 +201,10 @@ func (a *Atlas) NamedDiff(ctx context.Context, name string, tables ...*Table) er
 
 func (a *Atlas) cleanSchema(ctx context.Context, name string, err0 error) (err error) {
 	defer func() {
-		if err0 != nil {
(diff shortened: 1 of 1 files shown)


dialect/sql/schema/atlas.go
```diff
@@ -201,7 +201,10 @@ func (a *Atlas) NamedDiff(ctx context.Context, name string, tables ...*Table) er
 
 func (a *Atlas) cleanSchema(ctx context.Context, name string, err0 error) (err error) {
 	defer func() {
-		if err0 != nil {
+		switch {
+		case err == nil:
+			err = err0
+		case err != nil && err0 != nil:
 			err = fmt.Errorf("%v: %w", err0, err)
```
a8m, line 208: err = errors.Join(err, err0)

a8m (approved):

PR #4251, opened 2024-11-02
doc: remove `-d` flag in `go get` commands

leakedmemory:
As pointed [here](https://go.dev/doc/go-get-install-deprecation), since Go 1.18, `go get` always act as `go get -d`.


a8m (approved):
Thanks @leakedmemory

Issue #3592, opened 2023-06-08
Add support for NULLS NOT DISTINCT in Postgres 15


ivanvanderbyl:
- [ ] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

Postgres 15 adds the SQL standard `NULLS NOT DISTINCT` option to unique indexes.

[Docs](https://www.postgresql.org/about/featurematrix/detail/392/#:~:text=By%20default%2C%20NULL%20values%20are,or%20in%20other%20words%2C%20equivalently.
)

## Motivation 🔦

This allows for a use case where you might have two edges that are optional, but you want to create an index where one can be null and still unique across the other edge.

PR #4262, opened 2024-11-18
dialect: add `NULLS NOT DISTINCT` in Postgres 15

walkersumida:
`NULLS NOT DISTINCT` option is a feature that has been available since Postgres 15.

close https://github.com/ent/ent/issues/3592

More detailed information can be found in the following documents:
- [UNIQUE NULLS NOT DISTINCT](https://www.postgresql.org/about/featurematrix/detail/392/)
- [Release Notes PostgreSQL 15.0](https://www.postgresql.org/docs/release/15.0/)


adsouza:
I'd like to use Ent but need to use the NULLS NOT DISTINCT feature of Postgres.
Since you already have a PR for this, is it likely you will merge it any time soon?

PR #4290, opened 2025-01-08
dialect/sql/schema: inspect outside transaction in auto migrate

masseelch:
Since SQLite does not allow enabling/disabling foreign key checks within a transaction, Atlas disabled foreign key checks before opening a transaction and re-enables them after commit/rollback. This involves checking for violations every time the auto migrate tool checks for changing. By opening a transaction only in case there are changes, we can avoid this when not needed.

Closes https://github.com/ariga/atlas/issues/3297

### DIFF
--- a/entc/integration/migrate/migrate_test.go
+++ b/entc/integration/migrate/migrate_test.go
@@ -241,6 +241,23 @@ func TestSQLite(t *testing.T) {
 	require.NoError(t, err)
 	defer vdrv.Close()
 	Versioned(t, vdrv, "sqlite3://file?mode=memory&cache=shared&_fk=1", versioned.NewClient(versioned.Driver(vdrv)))
+
+	// In case there are no changes, no transaction should be opened and foreign keys should not be touched.
+	fkdrv, err := sql.Open("sqlite3", "file:fk_ent?mode=memory&cache=shared&_fk=1")
+	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


entc/integration/migrate/migrate_test.go
```diff
@@ -241,6 +242,24 @@ func TestSQLite(t *testing.T) {
...
+		case strings.HasPrefix(s, "driver.Tx"):
+			t.Errorf("unexpected transaction, %q", s)
+		case strings.HasPrefix(s, "driver.Exec") && strings.Contains(s, "PRAGMA foreign_keys"):
+			t.Errorf("unexpected foreign keys pragma: %q", s)
+		}
+	}))
+	require.NoError(t, fkclient.Schema.Create(ctx))
+	time.Sleep(time.Second * 2)
+	// Apply again has no changes -> there should not be any pragma changes and no transaction to be opened.
+	require.NoError(t, fkclient.Debug().Schema.Create(ctx))
```
a8m, line 262: What are these?

masseelch: It is a bit hacky, but this way I am checking if the migration engine is not opening a transaction or fiddling with pragmas. 

Other solution: write a custom driver to check what is executed.

a8m (approved):

PR #4293, opened 2025-01-10
entc: global id feature

masseelch:
Feature to add annotations to ent schema to ensure sql tables have sequential auto-increment id columns. Meant to be a better alternative for entgql / gqlgen globally unique id feature.


entc/gen/template/internal.tmpl
```diff
@@ -17,3 +17,14 @@ package internal
...
+
+{{ define "internal/globalid" }}
+
+{{ with $.Header }}{{ . }}{{ else }}// Code generated by ent, DO NOT EDIT.{{ end }}
+
+// +build tools
+
+package internal
+
+const IncrementStarts = {{ .Annotations.IncrementStarts | json | quote }}
```
giautm, line 29: Maybe add a GoDoc here,

entc/gen/globalid.go
```diff
@@ -0,0 +1,116 @@
...
+	return nil
+}
+
+// Name implements Annotation interface.
+func (IncrementStarts) Name() string {
+	return "IncrementStarts"
+}
+
+func rangesFilePath(dir string) string {
+	return filepath.Join(dir, "internal/globalid.go")
```
giautm, line 115:
```suggestion
	return filepath.Join(dir, "internal", "globalid.go")
```

entc/gen/globalid.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+		a := n.EntSQL()
+		if a == nil {
+			a = &entsql.Annotation{}
+			n.Annotations[a.Name()] = a
+		}
+		switch v, ok := r[n.Table()]; {
+		case a.IncrementStart != nil:
+			// In case the start value is defined by an annotation already, it has precedence.
+			r[n.Table()] = *a.IncrementStart
```
a8m, line 84: I believe we should validate the ranges. Protect the user from doing mistakes here. If global ID is enabled, two node types cannot have overlapped ranges:
image

a8m (commented):
Looks good, just a minor comment

entc/gen/globalid.go
a8m: WDYT, @giautm and @masseelch?

entc/gen/template/internal.tmpl
masseelch: Followed the rest of the template (where there is no GoDoc).

entc/gen/globalid.go
masseelch: How do you want to ensure there is no overlap? The only thing making slightly sense is to check two nodes have not the same increment start counter. But users  might decide a specific table only requires a range of 10 entries. 

Do you want to enforce that there is at least a gap of 1<<32 between nodes?

a8m: > Do you want to enforce that there is at least a gap of 1<<32 between nodes?

I think so. This is how it works now.

a8m (approved):

PR #4296, opened 2025-01-15
dialect/sql/schema: add schema dump command

masseelch:
related to https://github.com/ariga/atlas/issues/3307


dialect/sql/schema/schema.go
```diff
@@ -576,3 +583,63 @@ func indexType(idx *Index, d string) (string, bool) {
...
+		},
+		entdialect.MySQL: {
+			&MySQL{version: v},
+			mysql.DefaultDiff,
+			mysql.DefaultPlan,
+		},
+		entdialect.Postgres: {
+			&Postgres{version: v},
+			mysql.DefaultDiff,
+			mysql.DefaultPlan,
```
giautm, line 608: pg?

dialect/sql/schema/schema.go
```diff
@@ -576,3 +583,63 @@ func indexType(idx *Index, d string) (string, bool) {
...
+		},
+	}
+}
+
+// Dump the schema DDL for the given tables.
+func Dump(ctx context.Context, dialect, version string, tables []*Table, opts ...migrate.PlanOption) (string, error) {
```
giautm, line 614: `Bootstrap` or sth-else, mean create the database from scratch

masseelch: It doesn't create anything. It simply dumps the schema.

dialect/sql/schema/schema.go
```diff
@@ -576,3 +584,63 @@ func indexType(idx *Index, d string) (string, bool) {
...
+	opts = append([]migrate.PlanOption{func(o *migrate.PlanOptions) {
+		o.Mode = migrate.PlanModeDump
+		o.Indent = "  "
+	}}, opts...)
+	d, ok := drivers(version)[dialect]
```
giautm, line 620: The function `Dump` is on the public package. Let allow the user Inject the driver they want.

dialect/sql/schema/schema.go
giautm: But the schema (SQL) generated here can be use to create the database, right?

masseelch: It can be used to create tables inside the schema. But they are not ordered for example (since it uses `PlanModeDump`)

a8m (approved):
I'd prefer having SQL statements formatted with `"  "` by default.

giautm (approved):

PR #4301, opened 2025-01-16
fix: use bigint for uint32 and uint16 in postgres driver

tobikris:
I could not store the max values of `uint32` using the default schema types.
To me it seems that the mapping is incorrect:

https://github.com/ent/ent/blob/b91f8daf0e3267a1c7818f8120e83a36d56054b3/dialect/sql/schema/postgres.go#L117-L118

`integer` type in PG ranges from `-2147483648 to +2147483647` (https://www.postgresql.org/docs/current/datatype-numeric.html) - which is not sufficient.

I then checked the other types and think that `uint16` also needs to be fixed.


a8m (approved):
Thanks for the fix.

PR #4304, opened 2025-01-17
entc/gen: allow writing increments to disk

masseelch:
Required for entfix command.

### DIFF
--- a/entc/gen/globalid.go
+++ b/entc/gen/globalid.go
@@ -121,6 +121,25 @@ func (IncrementStarts) Name() string {
 	return "IncrementStarts"
 }
 
+// WriteToDisk writes the increment starts to the disk.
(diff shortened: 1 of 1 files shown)


cmd/entfix/entfix.go
```diff
@@ -0,0 +1,76 @@
+package main
```
a8m, line 1: License

cmd/entfix/go.mod
```diff
@@ -0,0 +1,30 @@
+module entgo.io/entfix
+
+go 1.23.4
+
+replace entgo.io/ent => ../..
```
a8m, line 5: I think a standalone module cannot use replace (when using go get).

a8m (approved):
See my comment regarding replace

a8m (approved):

PR #4319, opened 2025-02-08
[WIP] ent Computed field support

vtolstov:
First try to be able to say that field can't be created or updated, but says that it contents generated internally.
Tests are not ready now, only to start discussion


vtolstov:
mostly this is for RFC, Computed duplicates some logic of Immutable and also i think that naming not so good. I'm welcome to suggestions

vtolstov:
how to deal with deleted_at for example ? It must not have In Create input and Update input

vtolstov:
this solved my issues	entgql.Skip(entgql.SkipMutationCreateInput|entgql.SkipMutationUpdateInput),

PR #4326, opened 2025-02-12
doc/website: fix go get entfix path

masseelch:


giautm:
why you need to touch yarn.lock?

giautm (approved):

PR #4335, opened 2025-02-21
dialect/sql/schema: add multi schema and view support for schema dump

masseelch:


a8m (approved):
💯 

Beast! 

See my minor comment

PR #4349, opened 2025-03-12
fix: set the dialect for the generated query

michaelcaulley:
different quote characters are used by different dialects


michaelcaulley:
@masseelch, Can you check this change? 

Let me know if there is a good way to add a test. Without using a Postgres instance, I couldn't think of a way beyond stubbing out the interface and checking the generated query text. Since you use SQLite in the existing test, I decided against the change.

masseelch:
I don't think this needs testing as the query builder already has tests in place to ensure the correct quotes are used.

Issue #4351, opened 2025-03-15
`go generate ./ent` creates corrupted object models


altair363:
- [X] The issue is present in the latest release.
- [X] I have searched the [issues](https://github.com/ent/ent/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

If using a schema with the name starting with **O** and **K** (ObjKey in my case) the generated model created with the command `go generate ./ent` will have compiler errors in it because it uses the name as "ok" which is also used to control data retrieval from arrays/maps.

This is a function generated by the command:
`func (ok *ObjKey) assignValues(columns []string, values []any) error { ... }`

ok is used like so:

`value, ok := values[i].(*sql.NullInt64)`
`if !ok {`
`        return fmt.Errorf("unexpected type %T for field id", value)`
`}`
`ok.ID = int(value.Int64)` 

## Expected Behavior 🤔

Checking if the alias choosed for the schema is valid and not reserved.

## Steps to Reproduce 🕹

Steps:

1. Create a new schema ObjKey with the command `go run -mod=mod entgo.io/ent/cmd/ent new ObjKey`
2. (Optional) Populate the schema with fields and edges
3. Execute command `go generate ./ent`

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.24.0  |
| Ent         | 0.14.3   |
| Database    | PostgreSQL |
| Driver      | https://github.com/lib/pq       |

giautm:
Hey, please upgrade to latest version of ent. This issue was fixed by using a static receiver.

PR #4355, opened 2025-03-17
entc/gen: change receivers to static one

giautm:
> I have another suggestion for this change (which I wanted to do for a long time) - what do you think about using constant receiver names for all query builders. For example: q for <T>Query (e.g. UserQuery), c for <T>Create, u for <T>Update, and d for <T>Delete.

Extended [this comment](https://github.com/ent/ent/pull/1345#pullrequestreview-617000783) from Ariel: `_m` for models
- `_m`: for `<T>`
- `_q`: for `<T>Query`
- `_c`: for `<T>Create` or `<T>CreateBulk`
- `_u`: for `<T>Update` or `<T>UpdateOne`
- `_d`: for `<T>Delete` or `<T>DeleteOne`


entc/gen/type.go
```diff
@@ -336,11 +336,7 @@ func (t Type) PackageAlias() string { return t.alias }
 // Receiver returns the receiver name of this node. It makes sure the
 // receiver names doesn't conflict with import names.
 func (t Type) Receiver() string {
-	r := receiver(t.Name)
-	if t.Package() == r {
-		return "_" + r
-	}
-	return r
+	return "m"
```
a8m, line 339: This should be the model prefix imo. But what will happen for all those who use custom templates? The previous receiver should still work, so I believe we should detect this there as well.

giautm: @a8m There problem is, the issue https://github.com/ent/ent/issues/4351 has conflict on this model receiver, not the create/update/delete/query one. So it must be change to static to resolve https://github.com/ent/ent/issues/4351

We can add option to allow the user fallback to old behavior when upgrading. But, honestly, I prefer them to update the custom template, if the methods on Ent's model has variable that conflicting with `m`.

a8m: > @a8m There problem is, the issue https://github.com/ent/ent/issues/4351 has conflict on this model receiver, not the create/update/delete/query one. So it must be change to static to resolve https://github.com/ent/ent/issues/4351

But we can fix the template that is used by #4351. We have such checks in the template like there: https://github.com/ent/ent/blob/master/entc/gen/template/client.tmpl#L287 (e.g., we can replace ok with _ok). 

I just try to avoid breaking changes to user templates.

giautm: @a8m, I actually wanna remove all check like you mention. If we just move to static one, we don't need these check anymore, for all places.

> I just try to avoid breaking changes to user templates.

I prefer to mention this in the release and ask the user update their template if they using custom one.

giautm: Add check for #4351 is an Ad-hoc, and we will need to update for other if there is another conflict. Which still a pan for us.

giautm: Other option, that use `_m` as receiver. It will avoid the conflict with existing custom template.

a8m: You can also change `receiver` to prefix with `_`.

a8m (approved):

a8m (approved):

matdibu:
I'm not sure, but I think this didn't include sql modifiers
it works in v0.14.4 but it breaks in v0.14.5

in the `sqlAll` function

matdibu:
nice! thank you, that fixed it

should we try to get a contrib release so people don't run into this again?

PR #4356, opened 2025-03-18
entc/gen/template: remove receiver conflict handling logic

giautm:
Stacked on: https://github.com/ent/ent/pull/4355


masseelch (approved):
So noisy :smile:

PR #4371, opened 2025-04-08
docs(schema/field): fix UUID field example to use uuid.UUID{} instead of uuid.New()

xolinar:
The UUID field example was using `uuid.New()` as the second parameter, which passes a concrete value instead of a type. This is misleading and may cause errors in usage.

Updated the example to use `uuid.UUID{}` as the zero value type, along with `.Default(uuid.New)` to show how to generate new UUIDs correctly.

This better aligns with official documentation: https://entgo.io/docs/schema-fields/#id-field


vtolstov:
best to use uuid.Nil vs uuid.UUID{}

PR #4388, opened 2025-05-15
docs: adjust globalid migrate link

boostvolt:
We started using the new global unique ID feature flag. Noticed the link to the migration was broken, so I quickly updated it to make things easier to find.


masseelch:
It seems to work just fine currently without the change. Can you point me to where the link is broken?

PR #4389, opened 2025-05-15
Add the ability to enforce adding table columns in user defined order

sjpotter:
We had a problem with migrations that when we add fields to ent tables, that ent wants to generate the schema in a different order than what one would get by altering the table.  This enables the user to request that ent generate the table columns in the order specified in fields array.

### DIFF
--- a/dialect/entsql/annotation.go
+++ b/dialect/entsql/annotation.go
@@ -193,6 +193,10 @@ type Annotation struct {
 	//	}
 	ViewFor map[string]string `json:"view_for,omitempty"`
 
+	// TableFieldOrder tells ent to use the user provided Field order for columns,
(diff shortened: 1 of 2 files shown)


sjpotter:
Some context for why this impacted us.

We have our own migration scripts, and we ensure our migration scripts correspond to the schema ent generates by loading both into postgresql and doing a pg_dump and ensuring they are the same.  As ALTER TABLE puts columns at end, while ent was sticking them before edges, diff was failing.

For now we have worked around it, by writing a small normalize program that sorts the columns that are in the CREATE TABLE block in alphabetical order.  A bit ugly, but solves the immediate problem.

I still think giving users the ability to control the order of their columns is valuable.

PR #4395, opened 2025-05-22
fix(codegen): update groupReceiver and selectReceiver

eiixy:
#4355

### DIFF
--- a/entc/gen/type.go
+++ b/entc/gen/type.go
@@ -899,6 +899,16 @@ func (t Type) MutationName() string {
 	return pascal(t.Name) + "Mutation"
 }
 
+// GroupReceiver returns the receiver name of the delete-builder for this type.
+func (t Type) GroupReceiver() string {
+	return "_g"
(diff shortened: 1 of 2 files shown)


entc/gen/type.go
```diff
@@ -899,6 +899,16 @@ func (t Type) MutationName() string {
 	return pascal(t.Name) + "Mutation"
 }
 
+// GroupReceiver returns the receiver name of the delete-builder for this type.
+func (t Type) GroupReceiver() string {
+	return "_g"
+}
+
+// SelectReceiver returns the receiver name of the delete-builder for this type.
```
a8m, line 907:
```suggestion
// GroupReceiver returns the receiver name of the group-by builder for this type.
func (t Type) GroupReceiver() string {
	return "_g"
}

// SelectReceiver returns the receiver name of the selector builder for this type.
```

giautm (approved):

PR #4396, opened 2025-05-22
Enum Builder takes a string, should take an any

sjpotter:
Without an any, one can pass an enum value to it, and if one tries a string representation of the enum, one ends up with code that will fail to compile as it will generate

var TypeDefault EnumType = "some string"

### DIFF
--- a/schema/field/field.go
+++ b/schema/field/field.go
@@ -248,7 +248,7 @@ func (b *stringBuilder) Validate(fn func(string) error) *stringBuilder {
 }
 
 // Default sets the default value of the field.
-func (b *stringBuilder) Default(s string) *stringBuilder {
+func (b *stringBuilder) Default(s any) *stringBuilder {
 	b.desc.Default = s
 	return b
 }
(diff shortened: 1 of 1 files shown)


sjpotter:
for some background, I am creating a field "scope"

```
func (DependencyResolution) Fields() []ent.Field {
	return []ent.Field{
		field.Bool("circular").Default(false),
		// "Direct" string used as a place holder, fixed with sed in generate_ent.sh
		field.Enum("scope").GoType(scope.Scope(0)).Default("Direct"),
	}
}
```

this generates the code

```
const DefaultScope scope.Scope = "Direct"
```

if I don't provide a default value, it works (but then anything that doesn't set it, fails).  even though scope.Scope implements all the required interfaces for a custom GoType, it can't be a string, I have a hack, that works by running a sed script as part of the same go:generate code to do something like

```
sed -i "" 's/const DefaultScope.*/const DefaultScope scope.Scope = scope.Direct/' ./ent/table/table.go
```

but this is such a hack, I'm not keen on using it.

PR #4398, opened 2025-05-24
schema/field: validate rune length with `MinRuneLen` / `MaxRuneLen`

liangminhua:
- **Issue**: The original `MaxLen` validator used `len(v)` (byte count), while databases like MySQL/PostgreSQL/SQLite define `varchar(size)` as **character count** (not bytes). This caused inconsistent validation for multi-byte characters (e.g., Chinese, Emoji).
- **Fix**: Switched to `utf8.RuneCountInString(v)` to count Unicode characters, ensuring the validation aligns with database constraints.
- **Impact**: All supported databases (MySQL/PostgreSQL/SQLite) now have consistent character-length checks at both application and database layers.
- **Future Compatibility**:
  - For Oracle/SQL Server (where `varchar` uses byte counts by default), explicit syntax like `varchar(N CHAR)` or `nvarchar` can enforce character limits. This change does not affect existing logic but should be noted for future dialect support.

### DIFF
--- a/schema/field/field.go
+++ b/schema/field/field.go
@@ -15,6 +15,7 @@ import (
 	"regexp"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"entgo.io/ent/schema"
 )
@@ -222,6 +223,18 @@ func (b *stringBuilder) MinLen(i int) *stringBuilder {
 	return b
 }
 
+// MinRuneLen adds a rune length validator for this field.
+// Operation fails if the rune count of the string is less than the given value.
+func (b *stringBuilder) MinRuneLen(i int) *stringBuilder {
+	b.desc.Validators = append(b.desc.Validators, func(v string) error {
+		if utf8.RuneCountInString(v) < i {
+			return errors.New("value is less than the required rune length")
+		}
+		return nil
+	})
+	return b
+}
+
 // NotEmpty adds a length validator for this field.
 // Operation fails if the length of the string is zero.
 func (b *stringBuilder) NotEmpty() *stringBuilder {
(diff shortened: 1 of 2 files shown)


a8m:
Hey! Thanks for the contribution, but we can’t accept this change, as it will introduce a breaking-change. 

Please follow this custom validator example in our doc: https://entgo.io/docs/schema-fields#validators

```go
// MaxRuneCount validates the rune length of a string by using the unicode/utf8 package.
func MaxRuneCount(maxLen int) func(s string) error {
    return func(s string) error {
        if utf8.RuneCountInString(s) > maxLen {
            return errors.New("value is more than the max length")
        }
        return nil
    }
}

field.String("name").
    // If using a SQL-database: change the underlying data type to varchar(10).
    Annotations(entsql.Annotation{
        Size: 10,
    }).
    Validate(MaxRuneCount(10))
field.String("nickname").
    //  If using a SQL-database: change the underlying data type to varchar(20).
    Annotations(entsql.Annotation{
        Size: 20,
    }).
    Validate(MaxRuneCount(20))
```

liangminhua:
Thanks for the reply! Adding a new method with no disruptive changes.

a8m:
> Thanks for the reply! Adding a new method with no disruptive changes.

Thanks for the contribution, but I don't see a reason to add it to the standard ent validators.

@giautm, @masseelch - your thoughts?

a8m:
LGTM then.

@liangminhua, just add a small unit-test for it, and one of us will merge this.

liangminhua:
Thanks for the review! ✅

masseelch (approved):
LGTM 

cc @giautm

giautm (approved):

PR #4404, opened 2025-06-20
dialect/sql/sqlgraph: report check constraint violation with `IsCheckConstraintError`

Henry-Le-CS:
- Issue: Currently, entc does not support verifying Check constraint violations
- Implementation: Add the new function `IsCheckConstraintError` to check the returned error message.


giautm:
Please add unit-test in `TestIsConstraintError`

a8m (approved):
These helper functions predate `errors.Is`/`As`/`Wrap`, and with today's API, it's much more convenient to get the underlying database driver `error` type.

Still, I think it can be useful to merge this helper to keep it aligned with the current API provided by `sqlgraph`.

Henry-Le-CS:
Thanks for the reviews :raised_hands:

PR #4405, opened 2025-06-21
dialect: support index comment in mysql and postgresql

liasica:


a8m (approved):
Thanks for the contribution, @liasica.

giautm (approved):
LGTM,

PR #4429, opened 2025-08-13
Set schema when clearing foreign key edges

isazpiazu-roblox:
## Change Description
I am fixing a one-liner bug in `ent/dialect/sql/sqlgraph/graph.go` in which the schema was not being set when clearing foreign key relationships in the function `clearFKEdges` that results in always clearing the reverse edge in the public schema, even when the ent client was configured to use another schema.

### Context
I recently encountered an issue in which clearing a O2O bi-directional relationship cleared the field in for objects in two different schemas.
In our database, we generate copies of our public and data for manipulating data without affecting production. When clearing one side of a O2O, bidirectional relationship, I saw that the SQL commands were clearing the foreign keys for the relationship in the development schema for one object and the public schema for the other.

As an example, I have a table called `intfaces` that has a O2O, bidirectional relationship to itself represented by a foreign key from its column `id` to another column `peer_id`. When clearing the peer_id in one object in my development schema, it cleared the edge in the development schema but the reverse edge in the public schema, resulting in the following SQL commands being generated:
```
UPDATE "development"."intfaces" SET "peer_id" = NULL WHERE "id" = $1 args=[687169]
UPDATE "intfaces" SET "peer_id" = NULL WHERE "peer_id" = $1 args=[687169]
```

### Solution
Adding the `Schema` function to the `*sql.UpdateBuilder` in `clearFKEdges` fixes this issue. Here is the above operation's resulting SQL commands after the fix is implemented:
```
UPDATE "development"."intfaces" SET "peer_id" = NULL WHERE "id" = $1 args=[687173]
UPDATE "development"."intfaces" SET "peer_id" = NULL WHERE "peer_id" = $1 args=[687173]
```

I have created a local patch of the `ent/ent` repo in our project repository with the oneliner, bugfix that we will use until this PR is merged and the bugfix is released.
 
For additonal context, the `addFKEdges` already sets the Schema in its `*sql.UpdateBuilder` ([here](https://github.com/ent/ent/blob/175d96f7ecc21803c6bda4c1b8d8c8c1b7f7a50c/dialect/sql/sqlgraph/graph.go#L1832)). This fix brings `clearFKEdges` inline with `addFKEdges` in its support for multiple schemas.

### DIFF
--- a/dialect/sql/sqlgraph/graph.go
+++ b/dialect/sql/sqlgraph/graph.go
@@ -1801,6 +1801,7 @@ func (g *graph) clearFKEdges(ctx context.Context, ids []driver.Value, edges []*E
 			pred = matchIDs(edge.Target.IDSpec.Column, edge.Target.Nodes, edge.Columns[0], ids)
 		}
 		query, args := g.builder.Update(edge.Table).
+			Schema(edge.Schema).
 			SetNull(edge.Columns[0]).
 			Where(pred).
 			Query()


iazpiazu:
@a8m Can we merge my PR now that it's approved? Please let me know if you have any further feedback. Thank you!

PR #4436, opened 2025-09-01
libsql support

coffe-cafe-zed:
I have been using ent with https://turso.tech in development for a couple of month now and everything is working well except for creating records using ent.

Whenever I call `.Save()` function, the record will be created but and error will be returned. I dug into the issue and found that libsql does not support the "RETURNING" clause, just like MySQL.

This change is working well for me, but I am not sure if I need to do a full libsql driver.

Example connection:

```
	drv, err := sql.Open("libsql", viper.GetString("turso_database_url")+"?authToken="+viper.GetString("turso_auth_token"))
	if err != nil {
		config.Logger.Error("failed to open db", zap.Error(err))
		os.Exit(1)
	}

	Client := gen.NewClient(gen.Driver(drv))
```

### DIFF
--- a/dialect/dialect.go
+++ b/dialect/dialect.go
@@ -20,6 +20,7 @@ const (
 	SQLite   = "sqlite3"
 	Postgres = "postgres"
 	Gremlin  = "gremlin"
+	LibSQL   = "libsql"
 )
 
 // ExecQuerier wraps the 2 database operations.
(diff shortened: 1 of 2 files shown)


a8m:
I think you can do this instead: https://github.com/ent/ent/discussions/1667#discussioncomment-1132296

We don’t have the capacity to support Turso as non of us is using this service atm.

PR #4452, opened 2025-11-06
Clean up old & irrelevant references to Facebook

zpao:
Just cleaning up some things that should have been done when this got moved to LF years ago (https://entgo.io/blog/2021/09/01/ent-joins-the-linux-foundation)


a8m (approved):
💯 

Thanks @zpao

PR #4485, opened 2026-02-08
entc/gen: bake Schema into generated code and allow same table names across schemas

### LINKED ISSUE #2330: PostgreSQL schemas need smarter support
We need a better support for the postgres schema system.

## Current Behavior 😯

It's already possible to define entities schemas through the SchemaConfig feature, which works but it would be easier if it's possible with an annotation on the entity instead the client config.

But now I've conflicting code generation because of "duplicate" table names.

After building with entc I received invalid generated code:

```
validation failed: packages.Load: ...\ent\migrate\schema.go:2400:2: TemplatesColumns redeclared in this block
 ...\ent\migrate\schema.go:706:2:    other declaration of TemplatesColumns
 ...\ent\migrate\schema.go:2407:2: TemplatesTable redeclared in this block
 ...\ent\migrate\schema.go:720:2:    other declaration of TemplatesTable
```

It's caused because in the database are two `templates` tables. One in the `global` schema and one in the `lead` schema.
The code generation ignores the client schema config and doesn't use the schema as a prefix.

## Expected Behavior 🤔

Schemas are a basic feature of postgres and shouldn't cause conflicts with the code generation.

## Steps to Reproduce 🕹

Create a postgres database with the same table name in two different schemas.

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| Go          | 1.17.2  |
| Ent         | 0.9.2   |
| Database    | Postgres |
| Driver      | https://github.com/jackc/pgx        |

Thanks for your awesome work. ❤️

dannyota:
Fixes #2330

When two PostgreSQL schemas have tables with the same name (e.g., `global.templates` and `lead.templates`), code generation fails because `Tables()` rejects same table names across schemas and the template generates duplicate Go identifiers.

### Changes

**`entc/gen/graph.go`**
- Use schema-qualified keys (`"schema.table"`) in the tables map to allow same table names in different schemas
- Use `nodeT` map (`*Type → *schema.Table`) for FK lookups instead of bare table name lookups
- Add `computeTableIdents()` to generate unique Go identifiers: `pascal(name)` by default, `pascal(schema) + "_" + pascal(name)` on collision
- Add `TableIdent(*schema.Table) string` method on `Graph`, used by the schema template

**`entc/gen/template/migrate/schema.tmpl`**
- Replace all 6 `pascal $t.Name` / `pascal $fk.RefTable.Name` usages with `$.TableIdent`
- Emit the `Schema` field in generated table structs when set

**`dialect/sql/schema/schema.go`**
- Fix `CopyTables` to preserve `Schema`, `Comment`, `View`, `Pos` fields (previously silently dropped)
- Use schema-qualified keys in `byName` map and FK ref lookup

### Backward Compatibility

| Scenario | Behavior |
|----------|----------|
| No schema annotation | No `Schema` field emitted, identifiers unchanged. Generated code identical to before. |
| Single table per name | Go identifiers unchanged (e.g., `UsersTable`). |
| Same name, different schemas | New case — previously errored. Gets schema prefix (e.g., `Global_TemplatesTable`). |
| Manual `Table.Schema` overrides | Still work. If Schema is baked in, the manual set is a no-op. |
| `CopyTables` | Now correctly preserves `Schema` field (was silently dropped). |

### Testing

- Unit tests: `TestMultiSchemaSameTableName`, `TestSingleSchemaTableIdent`, `TestDuplicateTableNameSameSchema`, `TestCopyTablesMultiSchema`
- Integration: regenerated `multischema/ent/migrate/schema.go` with Schema baked in, verified against MySQL 8
- Regression: all integration tests pass (SQLite, MySQL 5.6/5.7/8, MariaDB, Postgres 10-17)

### DIFF
--- a/dialect/sql/schema/atlas.go
+++ b/dialect/sql/schema/atlas.go
@@ -693,15 +693,28 @@ func (a *Atlas) planInspect(ctx context.Context, conn dialect.ExecQuerier, name 
 	if err != nil {
 		return nil, err
 	}
-	var desired *schema.Schema
-	switch {
-	case realm != nil && len(realm.Schemas) > 0:
-		desired = realm.Schemas[0]
-	default:
-		desired = &schema.Schema{}
-	}
+	desired := a.desiredSchema(current, realm)
 	desired.Name, desired.Attrs = current.Name, current.Attrs
 	return a.diff(ctx, name, current, desired, a.types[len(types):], noQualifierOpt)
+}
+
+func (a *Atlas) desiredSchema(current *schema.Schema, realm *schema.Realm) *schema.Schema {
+	desired := schema.New(current.Name)
+	if realm == nil || len(realm.Schemas) == 0 {
+		return desired
+	}
+	if s, ok := realm.Schema(current.Name); ok {
+		return s
+	}
+	if a.schema != "" {
+		if s, ok := realm.Schema(a.schema); ok {
+			return s
+		}
+	}
+	if len(realm.Schemas) == 1 {
+		return realm.Schemas[0]
+	}
+	return desired
 }
 
 func (a *Atlas) planReplay(ctx context.Context, name string, tables []*Table) (*migrate.Plan, error) {
--- a/dialect/sql/schema/migrate_test.go
+++ b/dialect/sql/schema/migrate_test.go
@@ -69,6 +69,32 @@ func TestMigrate_SchemaName(t *testing.T) {
 	require.NoError(t, m.Create(context.Background()))
 }
 
+func TestAtlas_DesiredSchema(t *testing.T) {
+	global := schema.New("global").AddTables(schema.NewTable("templates"))
+	lead := schema.New("lead").AddTables(schema.NewTable("templates"))
+	realm := &schema.Realm{Schemas: []*schema.Schema{global, lead}}
+
+	t.Run("current schema", func(t *testing.T) {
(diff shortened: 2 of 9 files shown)


Miraitowa-zcx:
The original logic defaults to 'the first schema in the desired reality is the schema to be migrated'. Should we modify ent/dialect/sql/schema
/Lines 696-702 of atlas.go, this is my modification.
`desired := schema.New(current.Name)
	if realm != nil && len(realm.Schemas) > 0 {
		if s, ok := realm.Schema(current.Name); ok {
			desired = s
		} else if a.schema != "" {
			if s, ok := realm.Schema(a.schema); ok {
				desired = s
			} else if len(realm.Schemas) == 1 {
				desired = realm.Schemas[0]
			}
		} else if len(realm.Schemas) == 1 {
			desired = realm.Schemas[0]
		}
	}`

dannyota:
Thanks, good catch. I updated the inspect-mode migration logic so it no longer defaults to the first schema in the desired realm when multiple schemas are present.

The selection now prefers:
1. the desired schema matching the inspected current schema name,
2. then the configured schema name from WithSchemaName,
3. then the single desired schema only when there is exactly one.

I also added TestAtlas_DesiredSchema to cover current-schema matching, configured-schema fallback, single-schema compatibility, and the multi-schema no-match case.

Issue #4498, opened 2026-04-25
How to configure the generated migration code so that the ID field is NOT auto-incremented?


llh4github:
I configured the ID to be generated by application code, but the generated migration code still has the ID field as auto-increment. How can I configure it so that the ID field in the generated migration code is NOT auto-incremented?

I'm using a `BaseMixin` that defines the `id` field as `field.Uint64("id")` and generates the ID value via a hook (using [sonyflake](https://github.com/sony/sonyflake)):

**BaseMixin (mixin.go):**
```go
type BaseMixin struct {
	mixin.Schema
}

func (BaseMixin) Fields() []ent.Field {
	return []ent.Field{
		field.Uint64("id"),
	}
}

func (BaseMixin) Hooks() []ent.Hook {
	return []ent.Hook{
		hook.On(IDHook(), ent.OpCreate),
	}
}

func IDHook() ent.Hook {
	sf := sonyflake.NewSonyflake(sonyflake.Settings{})
	type IDSetter interface {
		SetID(uint64)
	}
	return func(next ent.Mutator) ent.Mutator {
		return ent.MutateFunc(func(ctx context.Context, m ent.Mutation) (ent.Value, error) {
			is, ok := m.(IDSetter)
			if !ok {
				return nil, fmt.Errorf("unexpected mutation %T", m)
			}
			id, err := sf.NextID()
			if err != nil {
				return nil, err
			}
			is.SetID(id)
			return next.Mutate(ctx, m)
		})
	}
}
```

**User schema (user.go):**
```go
type User struct {
	ent.Schema
}

func (User) Fields() []ent.Field {
	return []ent.Field{
		field.Int("age").
			Positive(),
		field.String("name").
			Default("unknown"),
	}
}

func (User) Mixin() []ent.Mixin {
	return []ent.Mixin{
		BaseMixin{},
	}
}
```

**Generated migration code (migrate/schema.go):**
```go
UsersColumns = []*schema.Column{
	{Name: "id", Type: field.TypeUint64, Increment: true},
	{Name: "age", Type: field.TypeInt},
	{Name: "name", Type: field.TypeString, Default: "unknown"},
}
```

Note that `Increment: true` is set on the `id` column even though the ID is generated by the application.

**Generated SQL:**
```sql
id bigint generated by default as identity primary key,
```

The table column is also auto-incremented, which is not what I want.

[...]

PR #4499, opened 2026-04-28
entc/gen: remove auto-increment for user-defined primary keys regardl…

NaftuliSinger:
This is a fix for issue #4498

I removed the check `&& !f.Type.Numeric()`.

> **Note:** If ent gives the user an option to manually set an `int` field as `incremental` (I am not sure there is a way to do that anyway) and the user has a custom `id` of type `int` and set that as `incremental`, this change will block that functionality.

**Tests:**
Some tests failed but they failed even before my change. I hope the maintainer can run github actions to check if it fails there too.

### DIFF
--- a/entc/gen/type.go
+++ b/entc/gen/type.go
@@ -1633,8 +1633,8 @@ func (f Field) PK() *schema.Column {
 		Comment:   f.sqlComment(),
 		Increment: f.incremental(f.Type.Type.Integer()),
 	}
-	// If the PK was defined by the user, and it is UUID or string.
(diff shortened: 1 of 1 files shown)


a8m:
We cannot accept the PR as this is a breaking change. Use the [`Incremental`](https://github.com/ent/ent/blob/master/dialect/entsql/annotation.go#L127C2-L127C13) annotation instead.

a8m:
Thanks for the contribution 🙏

NaftuliSinger:
That makes sense, no worries!

PR #4507, opened 2026-05-30
ci: add ActionScope GitHub Actions security exposure scan

r12habh:
## What

Adds a lightweight ActionScope workflow to scan GitHub Actions, Terraform, and IAM/policy JSON changes for CI/CD security exposure.

The workflow is intentionally conservative:

- runs only when workflow/action/IaC/policy files change, plus manual dispatch
- uses only `contents: read`
- does not call AWS APIs or require cloud credentials
- pins `actions/checkout` to a full commit SHA
- installs `actionscope>=0.3.5,<1.0` from PyPI
- fails only on `critical` findings, so the current non-critical findings do not block CI

## Why

I ran ActionScope locally against this repository and it found workflow-level CI/CD security signal around AWS credentials, token permissions, environments, or action supply-chain posture.

```text
Overall Risk: 🟠 HIGH
Workflows scanned: 4
AWS credential sources: 1
Critical: 0 | High: 3 | Medium: 1 | Low: 3
```

Because this uses `--fail-on critical`, this PR should add visibility without changing the current pass/fail posture.

### DIFF
--- a/.github/workflows/actionscope.yml
+++ b/.github/workflows/actionscope.yml
@@ -1,0 +1,39 @@
+name: ActionScope
+on:
+  pull_request:
+    paths:
+      - ".github/actions/**"
+      - ".github/workflows/**"
+      - "**/*.tf"
+      - "**/*.tf.json"
+      - "**/*iam*.json"
+      - "**/*policy*.json"
+  push:
+    branches:
+      - master
+    paths:
+      - ".github/actions/**"
+      - ".github/workflows/**"
+      - "**/*.tf"
+      - "**/*.tf.json"
+      - "**/*iam*.json"
+      - "**/*policy*.json"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  scan:
+    name: GitHub Actions security exposure
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Install ActionScope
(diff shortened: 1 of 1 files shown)


a8m:
Thanks for the contribution, but it’s not needed here.

r12habh:
Hey @a8m, you're right, it's not needed here. Actual scan since the original PR didn't include findings:

```
Workflows: 2 | Credential Sources: 0 | Overall Risk: HIGH

  0 AWS credential sources (ent's CI doesn't use AWS)
  0 known-compromised actions
  0 script-injection patterns
  0 artifact-poisoning patterns

  19 unpinned external action references
  3 elevated GITHUB_TOKEN permissions
```

One honest disclosure on the HIGH risk label: 2 of the 3 elevated GITHUB_TOKEN findings are `id-token: write` in `atlas-ci-public.yaml`, which is intentional because `ariga/atlas-action` uses OIDC to authenticate to Atlas Cloud. ActionScope currently only suppresses the HIGH severity for `id-token: write` when AWS's `configure-aws-credentials` is also in the workflow. It doesn't recognize Atlas Cloud, PyPI trusted publishing, npm provenance, Sigstore, GCP, Azure, or other legitimate OIDC consumers. That's a calibration bug in our tool, not a real finding on yours. Filed it as r12habh/ActionScope#85. Even setting that aside, the only real signal left is the 19 unpinned action refs, which actionlint and zizmor already cover. Your call was correct.

Disclosure I owe up front: I'm the author of ActionScope. That should have been line one of the original PR. I posted a longer apology covering the process failures on the powertools-lambda-python thread: aws-powertools/powertools-lambda-python#8253

Thanks for taking the time.

PR #4512, opened 2026-06-27
schema/field: export stringBuilder as StringBuilder

benza-1156c:
stringBuilder is unexported, making it impossible for external packages to reference it in generic type constraints or function signatures.

By exporting it as StringBuilder, downstream packages can now write helpers like:

    func WithLabel(b *field.StringBuilder, label string) *field.StringBuilder {
        return b.Annotations(MetaConfig{Label: label, Type: "TEXT"})
    }

This is a rename-only change with zero behavioral impact. All methods, return types, and generated code remain fully compatible.

### DIFF
--- a/schema/field/field.go
+++ b/schema/field/field.go
@@ -21,8 +21,8 @@ import (
 )
 
 // String returns a new Field with type string.
-func String(name string) *stringBuilder {
-	return &stringBuilder{&Descriptor{
+func String(name string) *StringBuilder {
+	return &StringBuilder{&Descriptor{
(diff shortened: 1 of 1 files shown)


a8m:
Thanks for the contribution, but I prefer to not export it - they are private for a reason.

If you want to attach annotations to string/any other field, you can use the [mixin.AnnotateFields](https://github.com/ent/ent/blob/master/schema/mixin/mixin.go#L92) function or copy its logic (each field exports a Descriptor object you can modify for your needs).

